PR #2, opened 2022-01-09
Basic package structure to get things started

bwengals:
Adds a few init files, setup.py, requirements.txt so it can be pip-installed from Github to get things started.


setup.py
```diff
@@ -0,0 +1,85 @@
...
+from os.path import dirname, join, realpath
+
+from setuptools import find_packages, setup
+
+DISTNAME = "pymc_experimental"
+DESCRIPTION = "A home for new additions to PyMC, which may include unusual probability distribitions, advanced model fitting algorithms, or any code that may be inappropriate to include in the pymc repository, but may want to be made available to users."
+AUTHOR = "PyMC Developers"
+AUTHOR_EMAIL = "pymc.devs@gmail.com"
+URL = "http://github.com/pymc-devs/pymc-experimental"
+LICENSE = "Apache License, Version 2.0"
```
ColCarroll, line 27: perhaps add this as well?

ColCarroll (approved):

setup.py
bwengals: good call, added

bwengals: Looks like I can't merge, if you don't mind and it looks good @ColCarroll would you do it?

PR #3, opened 2022-01-11
Adds 3 basic implementations of approximations for gp.Latent

bwengals:
- KarhunenLoeveExpansion (from @aseyboldt)
- Projected process / DTC
- Hilbert space GP (HSGP)

There's a notebook with a basic 1D demo of each.   

The Karhunen-Loeve expansion GP and the DTC approximation both can take arbitrary covariance functions, while HSGP only works for stationary kernels with a defined spectral density.  What is the best way to interface this with the way the kernel library is designed?


bwengals:
At this point the ProjectedProcess / DTC and KarhunenLoeve implementations are feature complete I think (so not counting tests, docstrings).  

The HSGP one is trickier since it requires the spectral density of the kernel.  I'm not sure if the API should take string arguments (ie, "matern32") for kernels with spectral densities, or whether to expect and check for specific covariance inputs (ie `eta**2 * pm.gp.cov.ExpQuad(input_dim=1, ls=2)`).  The HSGP also needs to be expanded to handle multidimensional inputs (`input_dim > 1`). 

Also am not sure if there should be a factory function similar to `pm.gp.MarginalApprox`, so `pm.gp.LatentApprox`, that should return instances of the approx GPs, or whether the user should call them directly.

pymc_experimental/gp/latent_approx.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class LatentApprox(pm.gp.Latent):
+    ## TODO: use strings to select approximation, like pm.gp.MarginalApprox?
+    pass
+
+
+class ProjectedProcess(pm.gp.Latent):
+    ## AKA: DTC
+    def __init__(self, n_inducing, *, mean_func=pm.gp.mean.Zero(), cov_func=pm.gp.cov.Constant(0.0)):
```
fonnesbeck, line 24: I would add type hints to public-facing API and at least simple docstrings. Can be done in a separate PR if you prefer.

michaelosthege:
@bwengals can you link to papers for the three approximations?

Does one of them correspond to the method from https://arxiv.org/pdf/2002.09309.pdf ?

I'm asking because I want to refer to these methods & implementations in my thesis...

PR #4, opened 2022-02-02
Adding Generalized Extreme Value Distribution

ccaprani:
Follows from https://github.com/pymc-devs/pymc/pull/5085

@ricardoV94 This is an initial not-working commit to see if you agree with the structure. There are a couple of things worth discussing:

1. I don't think this repo needs to replicate `pymc` itself here - any existing distributions/functions/utilities there should just be imported here.
2. Due to the nature of this repo, I think each experimental addition should have it's own file - hence what I've gone with in `__init__`
3. I've cherry picked the imports from pymc, aesera, etc as required for `GenExtreme`; however, I wonder if it would be better to have a generic set of imports that all/most experimental distributions could just use (even if the rather blunt star import).
4. I presume tests should be part of this and have added such. The structure though will be different to `pymc` and perhaps just a lower level pytest format is acceptable?
5. Is there anything else I've missed? In particular, I think the imports should "just work" once there is a `pymc` installed, right?

Can you review what I've done so far, and if it's on the right track I'll get it working.

Once we agree the plumbing, I'll update the [Case Study example PR](https://github.com/pymc-devs/pymc-examples/pull/241) accordingly.

Cheers!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,8 @@ jobs:
       matrix:
         os: [ubuntu-18.04]
         floatx: [float32, float64]
+        test-subset:
+          - pymc_experimental/tests
       fail-fast: false
     runs-on: ${{ matrix.os }}
     env:
@@ -80,6 +82,8 @@ jobs:
       matrix:
         os: [windows-latest]
         floatx: [float32, float64]
+        test-subset:
+          - pymc_experimental/tests
       fail-fast: false
     runs-on: ${{ matrix.os }}
     env:
--- a/conda-envs/environment-test-py38.yml
+++ b/conda-envs/environment-test-py38.yml
@@ -11,3 +11,4 @@ dependencies:
 - xhistogram
 - pip:
   - "git+https://github.com/pymc-devs/pymc.git@main"
+  - blackjax
--- a/conda-envs/environment-test-py39.yml
+++ b/conda-envs/environment-test-py39.yml
@@ -11,3 +11,4 @@ dependencies:
 - xhistogram
 - pip:
   - "git+https://github.com/pymc-devs/pymc.git@main"
+  - blackjax
--- a/pymc_experimental/__init__.py
+++ b/pymc_experimental/__init__.py
@@ -12,3 +12,4 @@ if not logging.root.handlers:
 
 
 from pymc_experimental import distributions, gp, utils
+from pymc_experimental.inference.fit import fit
--- a/pymc_experimental/inference/__init__.py
+++ b/pymc_experimental/inference/__init__.py
@@ -1,0 +1,1 @@
+from pymc_experimental.inference.fit import fit
--- a/pymc_experimental/inference/fit.py
+++ b/pymc_experimental/inference/fit.py
@@ -1,0 +1,37 @@
+#   Copyright 2022 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+
+def fit(method, **kwargs):
+    """
+    Fit a model with an inference algorithm
+
+    Parameters
+    ----------
+    method : str
+        Which inference method to run.
+        Supported: pathfinder
+
+    kwargs are passed on.
+
+    Returns
+    -------
+    arviz.InferenceData
+    """
+    if method == "pathfinder":
+        try:
+            from pymc_experimental.inference.pathfinder import fit_pathfinder
+        except ImportError as exc:
+            raise RuntimeError("Need BlackJAX to use `pathfinder`") from exc
+        return fit_pathfinder(**kwargs)
--- a/pymc_experimental/inference/pathfinder.py
+++ b/pymc_experimental/inference/pathfinder.py
@@ -1,0 +1,149 @@
+#   Copyright 2022 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+
+import collections
+import sys
+from typing import Optional
+
+import arviz as az
+import blackjax
+import jax
+import jax.numpy as jnp
+import jax.random as random
+import numpy as np
+import pymc as pm
+from pymc import modelcontext
+from pymc.sampling import RandomSeed, _get_seeds_per_chain
+from pymc.sampling_jax import get_jaxified_graph
+from pymc.util import get_default_varnames
+
+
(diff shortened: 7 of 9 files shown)


pymc_experimental/tests/test_distributions_random.py
```diff
@@ -0,0 +1,40 @@
...
+import numpy as np
+import numpy.random as nr
+import numpy.testing as npt
+import pytest
+import scipy.stats as st
+
+from numpy.testing import assert_almost_equal, assert_array_almost_equal
+
+
+class TestGenExtreme(BaseTestDistribution):
```
ricardoV94, line 29: The BaseTestDistribution is not imported is it?

pymc_experimental/tests/test_distributions.py
```diff
@@ -0,0 +1,61 @@
...
+import sys
+
+import aesara
+import aesara.tensor as at
+import numpy as np
+import numpy.random as nr
+
+import pytest
+import scipy.stats
+import scipy.stats.distributions as sp
```
ricardoV94, line 25: `sp` is an odd alias (if you even need it). Usually used for `scipy.special`

ricardoV94:
> 1. I don't think this repo needs to replicate `pymc` itself here - any existing distributions/functions/utilities there should just be imported here.

Definitely!

> 2. Due to the nature of this repo, I think each experimental addition should have it's own file - hence what I've gone with in `__init__`

I don't have a strong opinion here. Feel free to set the standard

> 3. I've cherry picked the imports from pymc, aesera, etc as required for `GenExtreme`; however, I wonder if it would be better to have a generic set of imports that all/most experimental distributions could just use (even if the rather blunt star import).

Seems fine to me to use explicit imports in every file

> 4. I presume tests should be part of this and have added such. The structure though will be different to `pymc` and perhaps just a lower level pytest format is acceptable?

As you wish. The advantage of keeping a similar strategy is that it would make upgrading experimental features to PyMC easier

> 5. Is there anything else I've missed? In particular, I think the imports should "just work" once there is a `pymc` installed, right?

Have you ran the tests locally?

> Can you review what I've done so far, and if it's on the right track I'll get it working.

I think you are exactly on the right track

> Once we agree the plumbing, I'll update the [Case Study example PR](https://github.com/pymc-devs/pymc-examples/pull/241) accordingly.


Awesome 👍

ccaprani:
@ricardoV94 I've updated per discussion and `pmx` now works as a package, and at least the basic structure is now in place for further contributions. I've also updated the related example https://github.com/pymc-devs/pymc-examples/pull/241 which you can try locally perhaps, but it works fine for me.

One thing: the tests won't run locally because of https://github.com/pydata/xarray/pull/6039 - not sure if you can replicate?

Please review - I suppose there are two reviews in a way...1) the more general `pmx` structure; 2) the specifics of this distribution, `GenExtreme`.

pymc_experimental/tests/test_distributions.py
ccaprani: This is in the main `pymc` test code (https://github.com/pymc-devs/pymc/blob/1005d20b3c12d9b9a424c069f6a0f9962d73c41d/pymc/tests/test_distributions.py#L47) as such and used by AR1: https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_distributions.py#L502

ricardoV94: Indeed... seems like we should change it over there as well. Feel free to ignore my comment in the meantime :D

pymc_experimental/distributions/genextreme.py
```diff
@@ -0,0 +1,217 @@
...
+from scipy import stats
+
+from aesara.tensor.random.op import RandomVariable
+from pymc.distributions.distribution import Continuous
+from aesara.tensor.var import TensorVariable
+from pymc.aesaraf import floatX
+from pymc.distributions.dist_math import bound
+
+
+class GenExtremeRV(RandomVariable):
```
ricardoV94, line 34: This seems duplicated in continuous.py

pymc_experimental/tests/test_distributions.py
ccaprani: Changed to ssd!

ccaprani:
Something went wrong with my last commit - this new one has those comments addressed. Thanks!

pymc_experimental/tests/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+from numpy.testing import Tester
+
+test = Tester().test
```
ricardoV94, line 17: What is the deal with this?

ccaprani: Not sure; copying from those who know better: https://github.com/pymc-devs/pymc/blob/main/pymc/tests/__init__.py

pymc_experimental/tests/test_distributions_random.py
ccaprani: I think so? Line 19.

ricardoV94: Line 19 is `import aesara`?

pymc_experimental/tests/__init__.py
ricardoV94: I had never heard of that

pymc_experimental/tests/test_distributions_random.py
ccaprani: Gosh, this is odd; I'm looking at the file [test_distributions_random.py](https://github.com/pymc-devs/pymc-experimental/pull/4/files#diff-af58444799088696f461542ffd0363bc1d21cddcc2feba2af75a8eae1d4805ac) and lines 18 to 21 are:

```python
from pymc.tests.test_distributions_random import (
    BaseTestDistribution,
    seeded_scipy_distribution_builder,
)
```
We must be looking at different things?

ricardoV94: https://github.com/pymc-devs/pymc-experimental/blob/01e8f4845a65116510995980461ef156b3e6c1cd/pymc_experimental/tests/test_distributions_random.py

ricardoV94: Now I see it, I have no idea where I was...

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,218 @@
...
+        return check_parameters(logc_expression, 1 + xi * scaled > 0, sigma > 0)
+
+    def get_moment(value_var, size, mu, sigma, xi):
+        r"""
+        Using the mode, as the mean can be infinite when :math:`\xi > 1`
+        """
+        mode = at.switch(
+            at.isclose(xi, 0), mu, mu + sigma * (at.pow(1 + xi, -xi) - 1) / xi
+        )
+        return at.full(size, mode, dtype=aesara.config.floatX)
```
ricardoV94, line 218: I think this is the last thing, can you add tests for the moment? You can borrow the helper from `test_distribution_moments` in the main repo

pymc_experimental/tests/test_distributions_random.py
ccaprani: Ah - that was updated in a later commit https://github.com/pymc-devs/pymc-experimental/pull/4/commits/7356e3f3c4f9aa475e3be5a726fdba748f74ffb4: https://github.com/pymc-devs/pymc-experimental/blob/1232d0bcbf62a72154e3989ddef5d42e90804619/pymc_experimental/tests/test_distributions_random.py

pymc_experimental/tests/test_distributions.py
```diff
@@ -0,0 +1,54 @@
...
+    TestMatchesScipy,
+)
+
+# the distributions to be tested
+from pymc_experimental.distributions import (
+    GenExtreme,
+)
+
+
+class TestMatchesScipyX(TestMatchesScipy):
```
ricardoV94, line 32: Will this rerun all the tests in the baseclass? In a discussion with @canyon289 it seemed like that was the case

ccaprani:
Continued as #84 due to rebase.

PR #14, opened 2022-03-08
Move bart from PyMC

aloctavodia:
This is already usable, for example 

```python
with pm.Model() as model:
    σ = pm.HalfNormal("σ", Y.std())
    μ = pmx.BART("μ", X, Y)
    y = pm.Normal("y", μ, σ, observed=Y)
    idata = pm.sample()
```

will work, but still one features is missing compared to the PyMC version from the main repository.

* The functions in `utils` need to be changed or we need to change how we store the `samples_stats`: `variable_importance` and `bart_trees`. We used to do this preprocessing. 
    * https://github.com/pymc-devs/pymc/compare/main...aloctavodia:byebyebart?expand=1#diff-7b3470589153c7c90a2c05fc86402f9acf65555200d90f01b51a47ccef592628L637-L654

I will fix that in a separated PR.

### DIFF
--- a/pymc_experimental/__init__.py
+++ b/pymc_experimental/__init__.py
@@ -9,3 +9,6 @@ if not logging.root.handlers:
     if len(_log.handlers) == 0:
         handler = logging.StreamHandler()
         _log.addHandler(handler)
+
+
+from pymc_experimental.bart import *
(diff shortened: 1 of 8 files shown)


aloctavodia:
@ricardoV94 Do you have a suggestion to fix the autoassign issue?

ricardoV94:
> @ricardoV94 Do you have a suggestion to fix the autoassign issue?

I am surprised BART variables have a grad xD. I think we need a level above `Ideal`

PR #22, opened 2022-03-16
add CI similar to pymc

### LINKED ISSUE #17: Enable tests on Github

aloctavodia:
Closes #17


conda-envs/environment-test-py39.yml
```diff
@@ -0,0 +1,10 @@
+name: pymc-test-py39
```
twiecki, line 1: maybe let's just test on py39 and not the others.

conda-envs/environment-test-py39.yml
```diff
@@ -0,0 +1,10 @@
...
+channels:
+- conda-forge
+- defaults
+dependencies:
+- pymc>=4.0.0b1
```
twiecki, line 6:
```suggestion
- pymc>=4.0.0b3
```

conda-envs/environment-test-py39.yml
aloctavodia: I dropped 3.7 (we should also do it in PyMC, following numpy schedule). We should try to keep the test here and in pymc (approximately) in sync. As that will help to reduce the friction of moving code from here to pymc. And possible reduce issues with users using both pymc and pymc-experimental.

twiecki: Agreed on all points.

PR #45, opened 2022-06-11
Replace uses of remove `pm.gp.util.infer_size` function

### LINKED ISSUE #44: GP fails CI builds
There is no test for the GP modules in pmx + the is a constant error in CI builds

```
ImportError: cannot import name 'infer_size' from 'pymc.gp.util' (/usr/share/miniconda3/envs/pymc-test-py38/lib/python3.8/site-packages/pymc/gp/util.py
```

CC @michaelosthege @bwengals

michaelosthege:
The `infer_size` function was removed, because it was an overcomplicated way of doing `np.shape(.)[0]`.

`np.shape(tensor)` simply returns `tensor.shape` and since `size` can be symbolic, we don't care if it's numeric or not.

Closes #44

### DIFF
--- a/pymc_experimental/gp/latent_approx.py
+++ b/pymc_experimental/gp/latent_approx.py
@@ -6,7 +6,6 @@ from pymc.gp.util import (
     JITTER_DEFAULT,
     cholesky,
     conditioned_vars,
-    infer_size,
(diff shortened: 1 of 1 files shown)


ferrine:
thanks!

PR #46, opened 2022-06-17
BART: add shape argument

aloctavodia:
This allows fully non-parametrical models of the form

```python
with pm.Model() as model:
    w = pmx.BART("w", X, Y, shape=2)
    μ = pm.Deterministic("μ", w[0])
    σ = pm.Deterministic("σ", np.abs(w[1]))
    y = pm.Normal("y", μ, σ, observed=Y)
```

or models where the Y is not 1D.

### DIFF
--- a/pymc_experimental/bart/bart.py
+++ b/pymc_experimental/bart/bart.py
@@ -40,12 +40,15 @@ class BARTRV(RandomVariable):
         )
 
     def _infer_shape(cls, size, dist_params, param_shapes=None):
-        dist_shape = (cls.X.shape[0],)
+        if cls.shape > 1:
+            dist_shape = (cls.X.shape[0], cls.shape)
+        else:
+            dist_shape = (cls.X.shape[0],)
         return dist_shape
 
     @classmethod
     def rng_fn(cls, rng=np.random.default_rng(), *args, **kwargs):
-        return np.full_like(cls.Y, cls.Y.mean())
+        return np.full((cls.Y.shape[0], cls.shape), cls.Y.mean()).squeeze()
 
 
 bart = BARTRV()
@@ -72,6 +75,8 @@ class BART(NoDistribution):
         Each element of split_prior should be in the [0, 1] interval and the elements should sum to
         1. Otherwise they will be normalized.
         Defaults to None, i.e. all covariates have the same prior probability to be selected.
+    shape : int
+        If 1 (default) the BART distribution will have shape Y.shape[0]. If larger it will have  shape (Y.shape[0], shape).
     """
 
     def __new__(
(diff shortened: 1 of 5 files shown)


ricardoV94:
This seems to deviate from RandomVariables, where shapes grow to the left and there's a default `shape=None`, which is different from `shape=(1,)`. Is that intentional/ am I misreading it?

aloctavodia:
You are totally right, this is something I have to fix.

pymc_experimental/bart/bart.py
```diff
@@ -40,12 +40,18 @@ def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
         )
 
     def _infer_shape(cls, size, dist_params, param_shapes=None):
-        dist_shape = (cls.X.shape[0],)
+        if cls.shape is not None:
```
ricardoV94, line 43: Another question, why not make use of the aesara size parameter, instead of having a self.shape?

aloctavodia: Not sure what I have to change. I tried using size instead of shape and I got 

ValueError: Tensor of type TensorType(int64, (0,)) could not be cast to have 0 dimensions

ricardoV94: I don't mean here in specific, I mean in the whole BartRV. It looked like size is never used, and `self.shape` was basically doing what size does in vanilla RVs.

ricardoV94: For instance if you support `size` then you don't even need this method. You can just override `_supp_shape_from_params`,

Which just has to return `cls.X.shape[0]`. Aesara will then figure out how size expands from this core dimension automatically whenever.

aloctavodia: Yeah, I got that. But my attempt to use size instead of shape was not successful :-( I could try again tomorrow.

pymc_experimental/bart/bart.py
```diff
@@ -29,23 +29,20 @@ class BARTRV(RandomVariable):
...
-        return dist_shape
+    def _supp_shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        return (self.X.shape[0],)
 
     @classmethod
-    def rng_fn(cls, rng=np.random.default_rng(), *args, **kwargs):
-        return np.full_like(cls.Y, cls.Y.mean())
+    def rng_fn(cls, rng, X, Y, m, alpha, split_prior, size):
+        if size is not None:
+            return np.full((size[0], cls.Y.shape[0]), cls.Y.mean())
```
ricardoV94, line 43: Is size limited to one dimension on purpose? If so you could raise when BART variables are raised with size that ate longer than one, like `size=(2, 4)`

aloctavodia: Yes, that's on purpose. Good point!

ricardoV94: Usually size checks would be made in `make_node` or `dist`

PR #56, opened 2022-06-29
Initialize a prior from a fitted posterior

ferrine:
If you want to do knowledge transfer in a smart way, this is how you do this

```python
from pymc.distributions import transforms

with model1:
    trace = pm.sample()
    # trace.posterior.keys() ~ ["a", "b", "c", "d", "f", "g"]
    # a - vector
    # b - matrix
    # c - positive
    # d, f, g - some other variable we do not care about


with pm.Model(coords=dict(test=range(3))) as model:
    priors = pmx.utils.prior.prior_from_idata(
        trace, 
        var_names=["a"],
        b=("test", "test"),
        c=transforms.log, 
        d="e", 
        f=dict(dims="test"),
        g=dict(name="h", dims="test", transform=transforms.log)
    )
    # 0. do nothing special to 'a' and other items in "var_names"
    # 1. 'b' has coords ("test", "test")
    # 2. transform 'c' to logspace
    # 3. rename 'd' to 'e'
    # 4. say 'f' has coords 'test'
    # 5. do everything mentioned with 'g'
    # priors will be a dictionary with all the priors, variables are available by final name keys
```

### DIFF
--- a/pymc_experimental/utils/prior.py
+++ b/pymc_experimental/utils/prior.py
@@ -85,3 +85,10 @@ def _flatten(idata: arviz.InferenceData, **kwargs: ParamCfg) -> FlatInfo:
         info.append(dict(shape=shape, slice=slice(begin, end)))
         begin = end
     return dict(data=np.concatenate(vars, axis=-1), info=info)
+
+
+def _mean_chol(flat_array: np.ndarray):
+    mean = flat_array.mean(0)
+    cov = np.cov(flat_array, rowvar=False)
+    chol = np.linalg.cholesky(cov)
+    return mean, chol
--- a/pymc_experimental/utils/prior.py
+++ b/pymc_experimental/utils/prior.py
@@ -115,3 +115,9 @@ def _mvn_prior_from_flat_info(name, flat_info: FlatInfo):
         var = pm.Deterministic(vinfo["name"], var, dims=vinfo["dims"])
         result[vinfo["name"]] = var
     return result
+
+
+def prior_from_idata(idata, name="trace_prior_", *, var_names: Sequence[str] = (), **kwargs):
+    param_cfg = _parse_args(var_names=var_names, **kwargs)
+    flat_info = _flatten(idata, **param_cfg)
+    return _mvn_prior_from_flat_info(name, flat_info)
--- a/pymc_experimental/utils/prior.py
+++ b/pymc_experimental/utils/prior.py
@@ -9,7 +9,7 @@ class ParamCfg(TypedDict):
 
 
 def arg_to_param_cfg(
-    key, value: Optional[Union[ParamCfg, aeppl.transforms.RVTransform, str, Tuple]]
+    key, value: Optional[Union[ParamCfg, aeppl.transforms.RVTransform, str, Tuple]] = None
 ):
     if value is None:
         cfg = ParamCfg(name=key, transform=None, dims=None)
--- a/pymc_experimental/utils/prior.py
+++ b/pymc_experimental/utils/prior.py
@@ -1,4 +1,4 @@
-from typing import TypedDict, Optional, Union, Tuple
+from typing import TypedDict, Optional, Union, Tuple, Sequence, Dict
 import aeppl.transforms
 
 
@@ -8,7 +8,7 @@ class ParamCfg(TypedDict):
     dims: Optional[Union[str, Tuple[str]]]
 
 
-def arg_to_param_cfg(
+def _arg_to_param_cfg(
     key, value: Optional[Union[ParamCfg, aeppl.transforms.RVTransform, str, Tuple]] = None
 ):
     if value is None:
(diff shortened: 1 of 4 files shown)


ricardoV94:
I don't see the prior being set. Is this still a draft? I assume you will use interpolation / kde?

ferrine:
Yes, this is a draft. In the snippet, I've added the API to discuss. The API is permissive for types and flexible since it is easy to figure out the intension. I plan to approximate the posterior using the MvNormal in the untransformed space.

pymc_experimental/utils/prior.py
```diff
@@ -0,0 +1,76 @@
...
+
+def _flatten(idata: arviz.InferenceData, **kwargs: ParamCfg) -> FlatInfo:
+    posterior = idata.posterior
+    vars = list()
+    info = list()
+    begin = 0
+    for key, cfg in kwargs.items():
+        data = posterior[key].values
+        # omitting chain, draw
+        shape = data.shape[2:]
```
OriolAbril, line 67: There is no guarantee the `chain` and `draw` dimensions will always be in the beginning, there are perfectly valid xarray operations that modify the dimension order. In xarray only the dimension name is relevant.

A quick change to the code to take this into account would be:

```
sample_dims = ["chain", "draw"]
for ...
    batch_dims = [dim for dim in posterior[key].dims if dim not in sample_dims]
    data = posterior[key].stack(__sample__=sample_dims, __batch__=batch_dims)
    end = begin + len(data["__batch__"])
```

I suspect it might even be possible to simplify this further using https://docs.xarray.dev/en/latest/generated/xarray.Dataset.to_stacked_array.html#xarray.Dataset.to_stacked_array plus a `where` to check the start and end positions of each variable. I can take a look towards the end of July if it were to still be helpful by then.

ferrine:
ready for review

twiecki:
About API, the most common case will be to just transfer a single parameter from posterior to prior, so I wonder what that API looks like to make sure that we're not doing the general case well but not the common case well.

ferrine:
> About API, the most common case will be to just transfer a single parameter from posterior to prior, so I wonder what that API looks like to make sure that we're not doing the general case well but not the common case well.

* A scalar variable without transform
```python
with pm.Model(coords=dict(test=range(3))) as model:
    a = pmx.utils.prior.prior_from_idata(trace, var_names=["a"])["a"]
```

* A vector variable without transform, adding dims
```python
with pm.Model(coords=dict(test=range(3))) as model:
    a = pmx.utils.prior.prior_from_idata(trace, a=("test", ))["a"]
```
or 
```python
with pm.Model(coords=dict(test=range(3))) as model:
    a = pmx.utils.prior.prior_from_idata(trace, a=dict(dims=("test", )))["a"]
```
* A vector variable with transform ("what's left from Dirichlet")
```python
with pm.Model(coords=dict(test=range(3))) as model:
    a = pmx.utils.prior.prior_from_idata(trace, a=dict(dims=("test", ), transform=transforms.simplex))["a"]
```
If we do not need coords
```python
with pm.Model(coords=dict(test=range(3))) as model:
    a = pmx.utils.prior.prior_from_idata(trace, a=transforms.simplex)["a"]
```

pymc_experimental/utils/prior.py
```diff
@@ -0,0 +1,172 @@
...
+
+    >>> with pm.Model(coords=dict(test=range(4), options=range(3))) as model2:
+    ...     priors = prior_from_idata(
+    ...         trace,                  # the old trace (posterior)
+    ...         var_names=["a", "d"],   # take variables as is
+    ...
+    ...         e="new_e",              # assign new name "new_e" for a variable
+    ...                                 # similar to dict(name="new_e")
+    ...
+    ...         b=("test", ),           # set a coord to "test"
```
OriolAbril, line 159:
```suggestion
    ...         b=("test", ),           # set a dim to "test"
```

the `coords` argument is called coords because it maps dimensions (also called dimension names or coordinate names sometimes) to coordinates or coordinate values. Here the only info is on the dimension so I think using coord will be confusing.

pymc_experimental/utils/prior.py
```diff
@@ -0,0 +1,172 @@
...
+    ...                                 # similar to dict(dims=("test", ))
+    ...
+    ...         c=transforms.log,       # apply log transform to a positive variable
+    ...                                 # similar to dict(transform=transforms.log)
+    ...
+    ...                                 # set a name, assign a coord and apply simplex transform
```
OriolAbril, line 165:
```suggestion
    ...                                 # set a name, assign a dim and apply simplex transform
```

same here, it now also matches the kwarg which is `dims`.

pymc_experimental/utils/prior.py
```diff
@@ -0,0 +1,172 @@
...
+    ...         f=dict(name="new_f", dims="options", transform=transforms.simplex)
+    ...     )
+    ...     trace1 = pm.sample_prior_predictive(100)
```
OriolAbril, line 168: Might be worth adding a note or even the code to use plot_pair to compare the obtained posterior to the generated prior. Even with a mvnormal and transforms, there might be cases where the posterior is not retrieved correctly, and it will generally fail if the wrong transform is used, not sure how aware of default transforms are users.

Note: regarding auto-use of default transforms. I think that https://github.com/arviz-devs/arviz/issues/2056 plus a key code to map the strings in the attributes to common transforms will generally fix this issue.

twiecki:
@ferrine I like the API.

ferrine:
Time to merge then. Thanks for reviews

PR #59, opened 2022-07-28
added generalized gamma distribution

kylejcaron:
What is this PR about?
This PR implements the generalized gamma distribution, using the parameterization GG(alpha, p, lambda).

This corresponds with the following [PR in aesara](https://github.com/aesara-devs/aesara/pull/1059). 

It also was originally [this PR](https://github.com/pymc-devs/pymc/pull/5985) in pymc, but @ricardoV94 recommended to start off with pymc-experimental which makes sense to me especially since I'm having trouble using it with the `pm.Censored` API

### DIFF
--- a/pymc_experimental/distributions/continuous.py
+++ b/pymc_experimental/distributions/continuous.py
@@ -134,10 +134,10 @@ class GeneralizedGamma(PositiveContinuous):
 
     def logcdf(
         value,
-        alpha: Union[float, np.ndarray, TensorVariable],
-        p: Union[float, np.ndarray, TensorVariable],
-        lambd: Union[float, np.ndarray, TensorVariable],
-    ) -> RandomVariable:
+        alpha: TensorVariable,
+        p: TensorVariable,
+        lambd: TensorVariable,
+    ) -> TensorVariable:
         """
         Compute the log of the cumulative distribution function for GeneralizedGamma
         distribution at the specified value.
--- a/pymc_experimental/distributions/continuous.py
+++ b/pymc_experimental/distributions/continuous.py
@@ -16,13 +16,18 @@ from pymc.distributions.dist_math import check_parameters
 class GeneralizedGamma(PositiveContinuous):
     r"""
     Generalized Gamma log-likelihood.
+    
     The pdf of this distribution is
+    
     .. math::
+       
        f(x \mid \alpha, p, \lambda) =
         \frac{ p\lambda^{-1} (x/\lambda)^{\alpha - 1} e^{-(x/\lambda)^p}}
         {\Gamma(\alpha/p)}
+    
     .. plot::
         :context: close-figs
+        
         import matplotlib.pyplot as plt
         import numpy as np
         import scipy.stats as st
@@ -39,11 +44,13 @@ class GeneralizedGamma(PositiveContinuous):
         plt.ylabel('f(x)', fontsize=12)
         plt.legend(loc=1)
         plt.show()
+
     ========  ==========================================
     Support   :math:`x \in [0, \infty)`
     Mean      :math:`\lambda \frac{\Gamma((\alpha+1)/p)}{\Gamma(\alpha/p)}`
     Variance  :math:`\lambda^2 \left( \frac{\Gamma((\alpha+2)/p)}{\Gamma(\alpha/p)} - \left(\frac{\Gamma((\alpha+1)/p)}{\Gamma(\alpha/p)}\right)^2 \right)`
     ========  ==========================================
+    
     Parameters
     ----------
     alpha : tensor_like of float, optional
@@ -55,8 +62,10 @@ class GeneralizedGamma(PositiveContinuous):
     lambd : tensor_like of float, optional
         Scale parameter :math:`\lambda` (lambd > 0).
         Defaults to 1.
+    
     Examples
     --------
+    
     .. code-block:: python
         with pm.Model():
             x = pm.GeneralizedGamma('x', alpha=1, p=2, lambd=5)
--- a/pymc_experimental/distributions/continuous.py
+++ b/pymc_experimental/distributions/continuous.py
@@ -1,0 +1,166 @@
+from typing import List, Optional, Tuple, Union
+
+import numpy as np
+import aesara
+import aesara.tensor as at
+from pymc.aesaraf import floatX
+from aesara.tensor.var import TensorConstant, TensorVariable
+from aesara.tensor.random.op import RandomVariable
+
+from aesara.tensor.random.basic import gengamma
+from pymc.distributions.continuous import PositiveContinuous
+from pymc.distributions.shape_utils import rv_size_is_none
+from pymc.distributions.dist_math import check_parameters
+
+
+class GeneralizedGamma(PositiveContinuous):
+    r"""
+    Generalized Gamma log-likelihood.
+    The pdf of this distribution is
+    .. math::
+       f(x \mid \alpha, p, \lambda) =
+        \frac{ p\lambda^{-1} (x/\lambda)^{\alpha - 1} e^{-(x/\lambda)^p}}
+        {\Gamma(\alpha/p)}
+    .. plot::
+        :context: close-figs
+        import matplotlib.pyplot as plt
+        import numpy as np
+        import scipy.stats as st
+        import arviz as az
+        plt.style.use('arviz-darkgrid')
+        x = np.linspace(1, 50, 1000)
+        alphas = [1,1,2,2]
+        ps = [1, 2, 4, 4]
+        lambds = [10., 10., 10., 20.]
+        for alpha, p, lambd in zip(alphas, ps, lambds):
+            pdf = st.gengamma.pdf(x, alpha/p, p, scale=lambd)
+            plt.plot(x, pdf, label=r'$\alpha$ = {}, $p$ = {}, $\lambda$ = {}'.format(alpha, p, lambd))
+        plt.xlabel('x', fontsize=12)
+        plt.ylabel('f(x)', fontsize=12)
+        plt.legend(loc=1)
+        plt.show()
+    ========  ==========================================
+    Support   :math:`x \in [0, \infty)`
+    Mean      :math:`\lambda \frac{\Gamma((\alpha+1)/p)}{\Gamma(\alpha/p)}`
+    Variance  :math:`\lambda^2 \left( \frac{\Gamma((\alpha+2)/p)}{\Gamma(\alpha/p)} - \left(\frac{\Gamma((\alpha+1)/p)}{\Gamma(\alpha/p)}\right)^2 \right)`
+    ========  ==========================================
+    Parameters
+    ----------
+    alpha : tensor_like of float, optional
+        Shape parameter :math:`\alpha` (``alpha`` > 0).
+        Defaults to 1.
+    p : tensor_like of float, optional
+        Additional shape parameter `p` (p > 0).
+        Defaults to 1.
+    lambd : tensor_like of float, optional
+        Scale parameter :math:`\lambda` (lambd > 0).
+        Defaults to 1.
+    Examples
+    --------
+    .. code-block:: python
+        with pm.Model():
+            x = pm.GeneralizedGamma('x', alpha=1, p=2, lambd=5)
+    """
+    rv_op = gengamma
+
+    @classmethod
+    def dist(cls, alpha, p, lambd, **kwargs):
+        alpha = at.as_tensor_variable(floatX(alpha))
+        p = at.as_tensor_variable(floatX(p))
+        lambd = at.as_tensor_variable(floatX(lambd))
+
+        return super().dist([alpha, p, lambd], **kwargs)
+
+    def moment(rv, size, alpha, p, lambd):
+        alpha, p, lambd = at.broadcast_arrays(alpha, p, lambd)
+        moment = lambd * at.gamma((alpha + 1) / p) / at.gamma(alpha / p)
+        if not rv_size_is_none(size):
+            moment = at.full(size, moment)
+        return moment
+
+    def logp(
+        value,
+        alpha: Union[float, np.ndarray, TensorVariable],
+        p: Union[float, np.ndarray, TensorVariable],
+        lambd: Union[float, np.ndarray, TensorVariable],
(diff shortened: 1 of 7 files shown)


pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,166 @@
...
+            Shape parameter (alpha > 0).
+        p : tensor_like of float
+            Shape parameter (p > 0).
+        lambd : tensor_like of float
+            Scale parameter (lambd > 0).
+        Returns
+        -------
+        TensorVariable
+        """
+        logcdf_expression = at.log(at.gammainc(alpha / p, (value / lambd) ** p))
```
kylejcaron, line 150: I'm not very familiar with the math, but I am concerned the logcdf function _may_ have some numerical instability. I attempted to fit a model to simulated data with the `pm.Censored` API and it was a bit slow, and had a few divergences

 (I tested this with the [latest aeppl fix](https://github.com/aesara-devs/aeppl/pull/156) for the `pm.Censored` API).

ricardoV94: I don't see some obvious reparametrization. Unless there is some loggammaincc function we could use?

If you are concerned I would suggest you do a grid evaluation of the logcdf and evaluate against an arbitrary precision software like Mathematica.

That will confirm or dispell your concerns of stability, and then we can try to look for an alternative. It could also be the instability is not in the function itself but its gradients.

Do you have a minimum example where you found the sampling issues that you can share?

pymc_experimental/tests/test_distributions.py
```diff
@@ -0,0 +1,21 @@
...
+            GeneralizedGamma,
+            Rplus,
+            {"alpha": Rplusbig, "p": Rplusbig, "lambd": Rplusbig},
+            lambda value, alpha, p, lambd: sp.gengamma.logpdf(value, a=alpha / p, c=p, scale=lambd),
+        )
+        check_logcdf(
+            GeneralizedGamma,
+            Rplus,
+            {"alpha": Rplusbig, "p": Rplusbig, "lambd": Rplusbig},
+            lambda value, alpha, p, lambd: sp.gengamma.logcdf(value, a=alpha / p, c=p, scale=lambd),
```
ricardoV94, line 20: How does scipy compute the logcdf?

ricardoV94 (changes requested):
This looks great!

There are some type-hints issues but otherwise everything looks correct. 

I left some suggestions to investigate the logcdf concerns but I am afraid I was not of much help.

ricardoV94:
@OriolAbril Can you give some guidance on how to include the distribution in the doc pages?

OriolAbril:
It needs to be added to https://github.com/pymc-devs/pymc-experimental/blob/main/docs/api_reference.rst. Which also needs some extra changes (along with maybe some changes to the imports and module hierarchy) I think. For this PR only:

```
:mod:`pymc_experimental.distributions`
=============================

.. automodule:: pymc_experimental.distributions
   :members: histogram_approximation, GeneralizedGamma
```

In general, is the plan for users to use `pymc_experimental.utils.prior.prior_from_idata` or to use `pymc_experimental.utils.prior_from_idata`? Depending on what is the plan, the imports and structure when generating the API docs should be updated. The docs currently tell users to import `prior_from_idata` from `pmx.utils.prior`

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,166 @@
...
+from pymc.distributions.continuous import PositiveContinuous
+from pymc.distributions.shape_utils import rv_size_is_none
+from pymc.distributions.dist_math import check_parameters
+
+
+class GeneralizedGamma(PositiveContinuous):
+    r"""
+    Generalized Gamma log-likelihood.
+    The pdf of this distribution is
+    .. math::
```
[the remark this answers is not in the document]
kylejcaron: added!

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,166 @@
...
+    def moment(rv, size, alpha, p, lambd):
+        alpha, p, lambd = at.broadcast_arrays(alpha, p, lambd)
+        moment = lambd * at.gamma((alpha + 1) / p) / at.gamma(alpha / p)
+        if not rv_size_is_none(size):
+            moment = at.full(size, moment)
+        return moment
+
+    def logp(
+        value,
+        alpha: Union[float, np.ndarray, TensorVariable],
```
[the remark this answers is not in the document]
kylejcaron: updated! did the same for logcdf as well

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,166 @@
...
+        p: Union[float, np.ndarray, TensorVariable],
+        lambd: Union[float, np.ndarray, TensorVariable],
+    ) -> RandomVariable:
```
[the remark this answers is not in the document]
kylejcaron: updated! did the same for logcdf as well

pymc_experimental/tests/test_distributions.py
kylejcaron: I looked closer, the difference in their calculation of the cdf is just related to how they chose to parameterize the distribution. 

Their parameterization seems like it might be less ideal, where they allow one of their parameters to be both negative and positive, but not zero, as opposed to the parameterization here which has the equivalent parameter as constrained positive. So not much to report here, I think not mimicking scipy makes sense here, but I also didnt test it out

pymc_experimental/distributions/continuous.py
kylejcaron: Is Mathematica a paid software? I had trouble finding anything to test against that would guarantee numerical precision

here's a minimum (kind of) example where there are sampling issues (very slow sampling. The distribution is quick when not using the `pm.Censored` api)

```
import pandas as pd
import scipy.stats as stats
import pymc as pm
import arviz as az
import numpy as np
import matplotlib.pyplot as plt
import pymc_experimental as pmx


SEED = 99


# ########################
# code
# ########################

def sim_data(true_paramsA, true_paramsB, SEED):
	
	np.random.seed(SEED)

	# 2 groups that generate data
	a = stats.gengamma(true_paramsA[0]/true_paramsA[1], true_paramsA[1], scale=true_paramsA[2])
	b = stats.gengamma(true_paramsB[0]/true_paramsB[1], true_paramsB[1], scale=true_paramsB[2])
	N = 100 # obs per group

	# create a dataset of N obs from each group
	y_true = np.r_[
	        a.rvs(N),
	        b.rvs(N)
	    ]

	# randomly censor the dataset for survival analysis
	cens_time = np.random.uniform(0, y_true.max(), size=N*2)

	data = (
		pd.DataFrame({
		"group":[0]*N + [1]*N,
		"time": y_true})
		# adjust the dataset to censor observations
		## indicates if an event hasnt occurred yet (cens=1)
		.assign(cens = lambda d: np.where(d.time <= cens_time, 0, 1) )
		## indicates the latest time observed for each record
		.assign(time = lambda d: np.where(d.cens==1, cens_time, d.time) )
	)

	return data

def fit_model(data):

	cens_ = np.where(data.cens==1, data.time, np.inf)
	COORDS = {"group": ["A","B"]}
	group_ = data.group.values
	y_ = data.time.values

	with pm.Model(coords=COORDS) as model:

		# weakly informative priors for each parameter
		log_alpha = pm.Normal("log_alpha", 2, 0.75, dims="group")
		log_p = pm.Normal("log_p", 0.45, 0.3, dims="group")
		log_lambd = pm.Normal("log_lambd", 4.1, 0.4, dims="group")

		# helper vars
		alpha = pm.Deterministic("alpha", pm.math.exp(log_alpha), dims="group")
		p = pm.Deterministic("p", pm.math.exp(log_p), dims="group")
		lambd = pm.Deterministic("lambd", pm.math.exp(log_lambd), dims="group")

		# latent variable needed for censored fitting
		y_latent = pmx.GeneralizedGamma.dist( alpha[group_], p[group_], lambd[group_] )

		# likelihood of the observed data
		obs = pm.Censored("obs", y_latent, lower=None, upper=cens_, observed=y_)

		# fit
		idata = pm.sample(init="adapt_diag", random_seed=SEED)
		idata.extend(pm.sample_prior_predictive())
		idata.extend(pm.sample_posterior_predictive(idata))

	return model, idata


def main():
	true_paramsA = [2, 2, 60]
	true_paramsB = [20, 2, 60]
	data = sim_data(true_paramsA, true_paramsB, SEED)
	model, idata = fit_model(data)


if __name__ == '__main__':
	
	main()
```

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,166 @@
...
+        Returns
+        -------
+        TensorVariable
+        """
+        logp_expression = (
+            at.log(p)
+            - at.log(lambd)
+            + (alpha - 1) * at.log(value / lambd)
+            - (value / lambd) ** p
+            - at.math.gammaln(alpha / p)
```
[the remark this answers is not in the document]
kylejcaron: added!

pymc_experimental/distributions/continuous.py
ricardoV94: Are you using the main version of Aeppl? We haven't done a new release yet, but there was a recent patch that makes the gradients of censored logps more stable: https://github.com/aesara-devs/aeppl/pull/156

ricardoV94: > Is Mathematica a paid software?

I might have jumbled the names. In the past I've used https://www.wolframcloud.com/ with a free account.

kylejcaron: Thank you I'll take a look at that!

I'm currently using the main version of aeppl installed manually (with `pip install -e .` inside of a locally cloned aeppl repo), so it should have the censored logp update. The sampling is >10 minutes to fit a dataset of 200 observations, but it takes less than a minute if I dont use the `Censored` api

Some of the warnings I'm getting are:

```
/Users/kylejcaron/.pyenv/versions/3.9.7/lib/python3.9/site-packages/aesara/scalar/math.py:760: RuntimeWarning: invalid value encountered in double_scalars
  scipy.special.gammaincc(k, x) * (log_x - digamma_k)
```

and 

```
WARNING (aesara.tensor.basic_opt): Optimization Warning: The Op gammainc_der does not provide a C implementation. As well as being potentially slow, this also disables loop fusion.
```

kylejcaron: The other thing I just noticed is that theres correlation between the `lambda` parameter and the `p` parameter. Are there any suggested ways to try and reduce that correlation? Maybe I simply just have to try other parameterizations?

I'm not seeing anything too crazy in the trace plots however

download-1
download

ricardoV94: The gradient using a Python loop could definitely explain slowness :P

But perhaps the censored model is just difficult to sample? Does it improve if you reduce the amount of censoring?

Is a censored normal much faster?

kylejcaron: One thing to note is that everything is fine and samples quickly without censored data and without the `censored` api involved (i.e. just normal fitting on non-censored data)

Also, sampling might've actually gotten slower with less censoring (when using the `censored` api)? I didnt sample to completion, I just looked at the initial guess after 1 minute of sampling
 * Original, 46% (96/200 obs censored) - 40 minutes of sampling estimated
 * 25% censoring (50/200 censored) - 50 minutes
 * 2.5% (5/200 censored) - 1 hour 

 * Censored normal data sampled quickly for me (<5 seconds)
 * Censored Gamma was really slow (estimated to be > 1 hour. Also got slower with less censoring). Uncensored fitting on same data was fine.

The fact that the fitting gets slower with less censoring seems really off to me, and also surprised to see the Gamma distribution suffers from similar issues

kylejcaron: update: just looked at the [log-cdf of the gamma distribution](https://github.com/pymc-devs/pymc/blob/e64942be0bc643b2ffb709213546f3d0243bb5f7/pymc/distributions/continuous.py#L2320) and its implemented very similar to how I implemented the generalized gamma, so makes sense its slow - I guess this points to the gammainc_der op being an issue more than the parameterization

ricardoV94: > update: just looked at the [log-cdf of the gamma distribution](https://github.com/pymc-devs/pymc/blob/e64942be0bc643b2ffb709213546f3d0243bb5f7/pymc/distributions/continuous.py#L2320) and its implemented very similar to how I implemented the generalized gamma, so makes sense its slow - I guess this points to the gammainc_der op being an issue more than the parameterization
> 

Yes that's quite possible

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,175 @@
+from typing import List, Optional, Tuple, Union
+
+import numpy as np
+import aesara
+import aesara.tensor as at
+from pymc.aesaraf import floatX
+from aesara.tensor.var import TensorConstant, TensorVariable
+from aesara.tensor.random.op import RandomVariable
+
+from aesara.tensor.random.basic import gengamma
```
ricardoV94, line 10: I guess the gengamma is not available yet in the Aesara version pinned in PyMC?

kylejcaron: Its been 4 days but I just tried setting up a new clean virtual environment and it seems like its there!

pymc_experimental/distributions/continuous.py
kylejcaron: Given that fitting with the distribution works well without the Censored API and the regular Gamma distribution has the same problem in PyMC anyway, is it fair to call this ok and proceed with the PR (once tests are passing) since its experimental anyway?

It'd atleast be a good starting point to be improved upon, and the current parameterization I chose is fairly interpretable, and I still have some use cases for it atleast

ricardoV94: Definitely! I'll try to review it again soon.

pymc_experimental/distributions/continuous.py
kylejcaron: all set with the latest commit!

kylejcaron:
> @kylejcaron are you still interested in this, or should I close it?

Given so much has changed with the package and backends since, it's probably worth closing - thanks for checking!

PR #60, opened 2022-08-01
Added MVN_cdf.py

HarrySpearing:
Wrapper of scipy logcdf. To calculate the log cdf of  multivariate normal distribution @ricardoV94.

See document for calculation of MVN CDF gradient.
[mvn_cdf.pdf](https://github.com/pymc-devs/pymc-experimental/files/9234690/mvn_cdf.pdf)

### DIFF
--- a/pymc_experimental/distributions/mvncdf.py
+++ b/pymc_experimental/distributions/mvncdf.py
@@ -1,0 +1,65 @@
+import aesara
+from aesara.graph.basic import Apply
+from aesara.graph.op import Op
+from aesara.tensor.type import TensorType
+
+# Op calculates mvncdf
+
+
+def conditional_covariance(Sigma, mu, conditioned, conditioned_values):
+
+    """
+    Sigma: d x d covariance matrix
+    mu: mean, array length d
+    conditioned: array of dimensions to condition on
+    conditioned_values: array of conditional values equal in length to `conditioned`
+    """
+
+    target = (1 - conditioned).astype(bool)
+
+    Sigma_22 = Sigma[conditioned][:, conditioned]
+    Sigma_21 = Sigma[conditioned][:, target]
+    Sigma_12 = Sigma[target][:, conditioned]
+    Sigma_11 = Sigma[target][:, target]
+
+    Sigma_cond = Sigma_11 - np.matmul(np.matmul(Sigma_12, np.linalg.inv(Sigma_22)), Sigma_21)
+    mean_cond = np.delete(mu, conditioned) + np.matmul(Sigma_12, np.linalg.inv(Sigma_22)).dot(
+        conditioned_values - mu[conditioned]
+    )
+
+    return Sigma_cond, mean_cond
+
+
+class Mvncdf(Op):
+    __props__ = ()
+
+    def make_node(self, upper, mu, cov):
+        upper = at.as_tensor_variable(upper)
+        mu = at.as_tensor_variable(mu)
+        cov = at.as_tensor_variable(cov)
+
+        return Apply(self, [x], [x.type()])
+
+    def perform(self, node, inputs, output_storage):
+        # here do the integration
+        return scipy.stats.multivariate_normal.logcdf(upper, mu, cov)
+
+    def infer_shape(self, fgraph, node, i0_shapes):
+        return
+
+    def grad(self, inputs, output_grads):
+
+        grad_ = []
+        for i in range(len(mu)):
+            grad_.append(conditional_covariance(cov, mu, i, upper[i]))
+
+        return np.array(grad_)
+
+    def R_op(self, inputs, eval_points):
+        # R_op can receive None as eval_points.
+        # That mean there is no diferientiable path through that input
+        # If this imply that you cannot compute some outputs,
+        # return None for those.
+        if eval_points[0] is None:
+            return eval_points
+        return self.grad(inputs, eval_points)


pymc_experimental/distributions/mvncdf.py
```diff
@@ -0,0 +1,65 @@
...
+        cov = at.as_tensor_variable(cov)
+
+        return Apply(self, [x], [x.type()])
+
+    def perform(self, node, inputs, output_storage):
+        # here do the integration
+        return scipy.stats.multivariate_normal.logcdf(upper, mu, cov)
+
+    def infer_shape(self, fgraph, node, i0_shapes):
+        return
```
ricardoV94, line 48: This should just be the shape of `upper`, so `i0_shapes[0]`?

pymc_experimental/distributions/mvncdf.py
```diff
@@ -0,0 +1,65 @@
...
+
+class Mvncdf(Op):
+    __props__ = ()
+
+    def make_node(self, upper, mu, cov):
+        upper = at.as_tensor_variable(upper)
+        mu = at.as_tensor_variable(mu)
+        cov = at.as_tensor_variable(cov)
+
+        return Apply(self, [x], [x.type()])
```
ricardoV94, line 41: I guess you mean `upper`, not `x` which isn't defined anywhere.?

pymc_experimental/distributions/mvncdf.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    target = (1 - conditioned).astype(bool)
+
+    Sigma_22 = Sigma[conditioned][:, conditioned]
+    Sigma_21 = Sigma[conditioned][:, target]
+    Sigma_12 = Sigma[target][:, conditioned]
+    Sigma_11 = Sigma[target][:, target]
+
+    Sigma_cond = Sigma_11 - np.matmul(np.matmul(Sigma_12, np.linalg.inv(Sigma_22)), Sigma_21)
```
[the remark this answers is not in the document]
ricardoV94: Similar below

ricardoV94 (changes requested):
I left comments for some issues I spotted at a first glace. 

You can use aesara test grad utility (don't remember the name right now) to check the gradient implementation is correct, and there is also some utility to check the `infer_shape` is correct.

After making sure the `Op` is implemented (and tested), you can create a dispatch function for the MvNormal logcdf, so that calling `pm.logcdf(MvNormal.dist(...), upper)` will return the right thing. Right now the link is not done  yet, and it would raise a `NotImplementedError`. It will be something like:

```python
from aesara.tensor.random.basic import MvNormalRV
from aeppl.logprob import _logcdf

@_logcdf.register(MvNormalRV)
def mvnormal_logcdf(value, mu, cov):
  return ...
```

pymc_experimental/distributions/mvncdf.py
ricardoV94:
```suggestion
        return Apply(self, [upper, mu, cov], [upper.type()])

```

pymc_experimental/distributions/mvncdf.py
```diff
@@ -0,0 +1,65 @@
...
+    def make_node(self, upper, mu, cov):
+        upper = at.as_tensor_variable(upper)
+        mu = at.as_tensor_variable(mu)
+        cov = at.as_tensor_variable(cov)
+
+        return Apply(self, [x], [x.type()])
+
+    def perform(self, node, inputs, output_storage):
+        # here do the integration
+        return scipy.stats.multivariate_normal.logcdf(upper, mu, cov)
```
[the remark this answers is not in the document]
ricardoV94: Also need to extract the inputs here:

```python
upper, mu, cov = inputs
```

pymc_experimental/distributions/mvncdf.py
```diff
@@ -22,38 +23,40 @@ def conditional_covariance(Sigma, mu, conditioned, conditioned_values):
...
 
-    def make_node(self, upper, mu, cov):
-        upper = at.as_tensor_variable(upper)
+    def make_node(self, value, mu, cov):
+        value = at.as_tensor_variable(value)
         mu = at.as_tensor_variable(mu)
         cov = at.as_tensor_variable(cov)
 
-        return Apply(self, [x], [x.type()])
+        return Apply(self, [value], [value.type()])
```
ricardoV94, line 41: The Apply node specifies all inputs and outputs, so it should be 
```suggestion
        return Apply(self, [value, mu, cov], [value.type()])

```

pymc_experimental/distributions/mvncdf.py
```diff
@@ -22,38 +23,40 @@ def conditional_covariance(Sigma, mu, conditioned, conditioned_values):
...
 
     def perform(self, node, inputs, output_storage):
         # here do the integration
-        return scipy.stats.multivariate_normal.logcdf(upper, mu, cov)
+        value, mu, cov = inputs
+        output_storage[0] = scipy.stats.multivariate_normal.logcdf(value, mu, cov)
+
+        return output_storage
```
ricardoV94, line 48: You should not return anything

pymc_experimental/distributions/mvncdf.py
```diff
@@ -22,38 +23,40 @@ def conditional_covariance(Sigma, mu, conditioned, conditioned_values):
...
     def grad(self, inputs, output_grads):
-
+        value, mu, cov = inputs
         grad_ = []
         for i in range(len(mu)):
-            grad_.append(conditional_covariance(cov, mu, i, upper[i]))
+            grad_.append(conditional_covariance(cov, mu, i, value[i]))
 
-        return np.array(grad_)
+        return np.array(grad_)*output_grads[0]
```
ricardoV94, line 59: The returned gradients need to be symbolic Aesara expressions, not the result of Numpy computations. You might need a separate `Op` for the gradient. See an example here: https://www.pymc.io/projects/examples/en/latest/case_studies/blackbox_external_likelihood_numpy.html#blackbox_external_likelihood_numpy and https://www.pymc.io/projects/examples/en/latest/case_studies/wrapping_jax_function.html

pymc_experimental/distributions/mvncdf.py
```diff
@@ -63,3 +66,10 @@ def R_op(self, inputs, eval_points):
         if eval_points[0] is None:
             return eval_points
         return self.grad(inputs, eval_points)
+
+    
+
+    @_logcdf.register(MvNormalRV)
+    def mvnormal_logcdf(value, mu, cov):
+
+        return pm.logcdf(MvNormal.dist(mu, cov), value)
```
ricardoV94, line 75: Should be something like


```suggestion
mvncdf = Mvncdf()

@_logcdf.register(MvNormalRV)
def mvnormal_logcdf(value, mu, cov):
    return mvncdf(value, mu, cov)
```

ricardoV94 (changes requested):
Getting closer, but somethings are still off. I suggest you check the Aesara documentation on how to write a new Op and some of the guides I mentioned in comments below.

https://aesara.readthedocs.io/en/latest/extending/creating_an_op.html

pymc_experimental/distributions/mvncdf.py
ricardoV94: Which you can then test is working by calling `pm.logcdf(MvNormal.dist(mu, cov), value)`

HarrySpearing: Is there any documentation which shows me how to do this? I tried searching for dispatch function but I have no idea how it works, and what it does. So it's quite tricky for me to know how to fix this.

pymc_experimental/distributions/mvncdf.py
HarrySpearing: I think for now it might be best not to implement the gradient method. The scipy autograd library doesn't exist for multivariate normal cdf. For now I can just try and sample using MH.

pymc_experimental/distributions/mvncdf.py
ricardoV94: No documentation AFAIK, but you can see how we do it for some custom distributions: https://github.com/pymc-devs/pymc/blob/9b9826c586f04367b5027a0e472122ebcc636139/pymc/distributions/mixture.py#L359-L383

The idea is that the function should return an Aesara expression, which in your case is just the new `Op` you created.

PR #61, opened 2022-08-01
Finish HSGP

bwengals:
The goal of this PR is to finish the implementation of the [HSGP](https://arxiv.org/pdf/2004.11408.pdf) and make it generally usable, so tests and docs and all that.  This is a function-space GP approximation that is interesting because it uses a fixed set of basis vectors -- but allows the kernel hyperparameters to be learned.  The restriction is that the kernel must be stationary and you need it's power spectral density.

TODO list:
- [x] `prior` method
- [x] `conditional` method
- [x] 1D inputs
- [x] Multidimensional inputs (existing implementation here was 1D only)
- [ ] Multi-dimensional power spectral densities for common kernels
  - [x] ExpQuad
  - [x] Matern52
  - [x] Matern32
- [ ] Support additivity, like `gp.Latent` and `gp.Marginal` do
- [ ] Tests


junpenglao:
In case it is useful: https://colab.research.google.com/drive/10FcpmOm6awyVcR-1DK-xzHTqneD3xfS6

bwengals:
yes thank you @junpenglao! Birthdays would make a great example

pymc_experimental/gp/hsgp.py
```diff
@@ -0,0 +1,154 @@
...
+        psd = self.cov_func.psd(omega)
+        return at.dot(phi * psd, at.transpose(phi))
+
+    def prior(self, name, X, **kwargs):
+        X, _ = self.cov_func._slice(X)
+        self._set_boundary(X)
+        omega, phi, m_star = self._eigendecomposition(X, self.L, self.M, self.D)
+        psd = self.cov_func.psd(omega)
+        self.beta = pm.Normal(f"{name}_coeffs_", size=m_star)
+        self.f = pm.Deterministic(
```
ferrine, line 141: Is it possible to pass dims for the variable?

pymc_experimental/gp/hsgp.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def _build_conditional(self, name, Xnew):
+        Xnew, _ = self.cov_func._slice(Xnew)
+        omega, phi, _ = self._eigendecomposition(Xnew, self.L, self.M, self.D)
+        psd = self.cov_func.psd(omega)
+        return self.mean_func(Xnew) + at.squeeze(at.dot(phi, self.beta * psd))
+
+    def conditional(self, name, Xnew):
+        fnew = self._build_conditional(name, Xnew)
+        return pm.Deterministic(name, fnew)
```
ferrine, line 154: Is it possible to pass dims to the variable?

pymc_experimental/gp/hsgp.py
bwengals: yes! thanks for catching this

bwengals:
I think I'll move HSGP over into to https://github.com/pymc-devs/pymc/pull/6036.  HSGP needs that spectral covariance stuff to work.  Having a prereq PR in pymc for something in pymc-experimental is just weird anyway.  This split I think was a bad idea in hindsight.

I don't know how to make the reverse option work, moving the spectral covariance support into experimental, since you need to mess with the Covariance classes quite a bit.  And I _think_ that the HSGP method is on firm footing with the main GP stuff.  It has a couple papers, and having used it some it really does work super well.

pymc_experimental/gp/hsgp.py
```diff
@@ -0,0 +1,195 @@
...
+        *,
+        mean_func=pm.gp.mean.Zero(),
+        cov_func=pm.gp.cov.Constant(0.0),
+    ):
+        arg_err_msg = (
+            "`m` and L, if provided, must be lists or tuples, with one element per active "
+            "dimension."
+        )
+        try:
+            if len(m) != cov_func.D:
```
fonnesbeck, line 93: Do covariance functions have a `D` attribute?

fonnesbeck: Is this supposed to be `len(active_dims)`?

pymc_experimental/gp/hsgp.py
```diff
@@ -0,0 +1,195 @@
...
+            c = 1.0 / np.sqrt(L[d])
+            phi *= c * at.sin(at.sqrt(eigvals[:, d]) * (at.tile(X[:, d][:, None], m_star) + L[d]))
+        omega = at.sqrt(eigvals)
+        return omega, phi, m_star
+
+    def approx_K(self, X, L, m):
+        """A helper function which gives the approximate kernel or covariance matrix K. This can be
+        helpful when trying to see how well an approximation may work.
+        """
+        X, _ = self.cov_func._slice(X)
```
fonnesbeck, line 145: `_slice` takes two positional arguments.

fonnesbeck:
@bwengals I think you can ignore my comments -- looks like this depends on your PyMC PR being merged as well!

PR #62, opened 2022-08-02
Implementing linear conjugate gradients 

kunalghosh:
This is a draft implementation of the linear conjugate gradients algorithm.
It is designed to be output the same as `gpytorch.util.linear_cg` and can be tested to give the same results using the code snippet below.

Let's first generate the data:
```python
import torch 
import gpytorch
import numpy as np

# Test PSD Matrix

N = 10
rank = 5
np.random.seed(1234) # nans with seed 1234
K = np.random.randn(N, N)
K = K @ K.T + N * np.eye(N)
K_torch = torch.from_numpy(K)
print(np.diag(K))

y = np.random.randn(N, 1)
y_torch = torch.from_numpy(y)
```

Now we get the output from GPyTorch
```python
result, tridiag = gpytorch.utils.linear_cg(K_torch, y_torch, n_tridiag=1)
```

And also from our implementation:
```python
import pymc_experimental as pymx
result_pmx, tridiag_pmx = pymx.utils.linear_cg(K, y, n_tridiag=1)
```

Check the outputs
```python
assert np.allclose(result, result_pmx) is True, "BUG: result doesn't match gpytorch values"
assert np.allclose(tridiag tridiag_pmx) is True, "BUG: tridiagonal values don't match gpytorch values"
```


kunalghosh:
@bwengals Ping, can you take a look at the PR please :)

bwengals:
will do.  It might take me a couple days, it's a lot to go through, but it looks like you got it working, congrats!

pymc_experimental/utils/linear_cg.py
```diff
@@ -0,0 +1,279 @@
...
+    # lets normalize rhs
+    rhs = rhs / rhs_norm
+
+    # residuals
+    residual = rhs - mat @ initial_guess
+    batch_shape = residual.shape[:-2]
+
+    result = np.copy(initial_guess)
+
+    if np.allclose(residual, residual):
```
kunalghosh, line 113: Just a heads up, I noticed this check fail yesterday. Was passing when I ran it on Colab, looking into it.

kunalghosh: Fixed it should have been `if not np.allclose(...)`

pymc_experimental/utils/linear_cg.py
```diff
@@ -0,0 +1,279 @@
+from collections import namedtuple
+import numpy as np
+
+Setting = namedtuple("setting", "on")
+
+
+class settings:
```
bwengals, line 7: Since there are only two parameters here, could you pass `eval_cg_tolerance` and `cg_tolerance` as args to `linear_cg`?  I think it would simplify things a little bit, since you wouldn't need this class.

pymc_experimental/utils/linear_cg.py
```diff
@@ -0,0 +1,279 @@
...
+            (
+                result,
+                alpha,
+                residual_inner_prod,
+                eps,
+                beta,
+                residual,
+                precond_residual,
+                curr_conjugate_vec,
+            ) = linear_cg_retvals
```
bwengals, line 189: This is just a small nitpick, but you could combine this, so you don't need to create the `linear_cg_retvals`.

pymc_experimental/utils/linear_cg.py
```diff
@@ -0,0 +1,279 @@
...
+
```
bwengals, line 224: This block looks repeated, I think they both could be moved out side the if/else?

pymc_experimental/utils/linear_cg.py
```diff
@@ -0,0 +1,279 @@
...
+            " a gpytorch.settings.max_cg_iterations(value) context.".format(
+                k + 1, residual_norm.mean(), tolerance
+            )
+        )
+
+    if n_tridiag:
+        t_mat = t_mat[: last_tridiag_iter + 1, : last_tridiag_iter + 1]
+        return result, t_mat.transpose(-1, *range(2, 2 + len(batch_shape)), 0, 1)
+    else:
+        return result
```
bwengals, line 279: This function returns two different types, depending on inputs.  This can make downstream code more complicated, since it will have to know how to handle both cases.  Is it possible to refactor this into two functions, one for the tridaigonal case and one not?

bwengals:
It would be nice if you included your example code to test this out in the actual PR somewhere!  You could make a brief notebook that demonstrates how to use it.

I left some minor comments, but I think shouldn't hold up the merge process.  Depending on the strategy, this implementation will be translated to aesara/numba/jax, so the goal of this code should be clarity.  Feel free to make changes and merge it when you're ready.

bwengals:
BTW is this still PR still in draft mode, or do you think you're pretty much done?

pymc_experimental/utils/linear_cg.py
kunalghosh: This class is just a place holder, I would ideally put these two values in some sort of default settings file or class used by PyMC. I moved the constants to be global variables and kept the logic same as in GPyTorch.

pymc_experimental/utils/linear_cg.py
kunalghosh: Fixed in the latest commit

pymc_experimental/utils/linear_cg.py
kunalghosh: Would it make sense to return `None` as the second argument when `n_tridiag` is False ?

pymc_experimental/utils/linear_cg.py
bwengals: There's no need to follow GPyTorch's particular way of doing things -- in PyMC there aren't any global variables or settings files, that I'm aware of.  It's important to be consistent with these sort of things across the codebase.

pymc_experimental/utils/linear_cg.py
bwengals: Checking the type vs checking the number of outputs isn't effectively very different.  If it doesn't make sense to split this, I think it makes sense to leave this as is and be mindful of this later, maybe return to it once we see how it fits in.

kunalghosh: I now return identity matrices if n_tridiag = 0. This ensures downstream computations based on tridiagonal matrices are not affected (Eigen values of identity matrices are all one and log of that would be zero, so the log determinant terms would be zero) see. equation S7 (https://arxiv.org/pdf/1809.11165.pdf)

The identity matrices assume n_tridiag = 1 to ensure dimensions are consistent for downstream computations.

kunalghosh:
> BTW is this still PR still in draft mode, or do you think you're pretty much done?

Nope, this is ready to check-in now :)

bwengals (approved):
Sorry for the delay, everything looks copacetic

PR #64, opened 2022-08-09
Added model_builder as directed in PR #6023 on pymc

5hv5hvnk:


pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,129 @@
...
+            self.idata.to_netcdf(file)
+        if save_model:
+            filepath = Path(str(filepath) + str(file_prefix) + ".pickle")
+            Model = cloudpickle.dumps(self)
+            file = open(filepath, "wb")
+            file.write(Model)
+        self.saved = True
+
+    @classmethod
+    def load(cls, file_prefix, filepath, load_model=True, laod_idata=True):
```
tomicapretto, line 58: Typo in `laod_idata`.

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,129 @@
...
+        file = Path(str(filepath) + str(file_prefix) + ".pickle")
+        file = open(file, "rb")
+        self = cloudpickle.loads(file.read())
```
tomicapretto, line 61: A couple of comments here.

1. Wouldn't it be better to use a context manager to handle the connection? That makes sure the connection is closed, which is not happening here. 
2. You're loading a file and assigning it to `self` without knowing if the file you're reading contains an instance of `ModelBuilder`. This class may be helpful

```python
class Loader:
    def __init__(self):
        pass

    def save_to_pickle(self, filename):
        pickle_file = open(filename, "wb")
        pickle.dump(self, file=pickle_file)
        pickle_file.close()
        return None

    @classmethod
    def load_from_pickle(self, filename):
        with open(filename, "rb") as pickle_file:
            model = pickle.load(pickle_file)
        if isinstance(model, self):
            return model
        else:
            raise ValueError(
                f"The route '{filename}' does not contain an object of the class '{self.__name__}'"
            )
```

Then you can do

```python
class Model(Loader):
    def __init__(self):
        self.abc = 1
 
my_model = Model.load_from_pickle("file.pkl")
```

If you call `load_from_pickle` using an pkl that contains an object of a different class, it will fail. 
I'm not suggesting you to use the same names, but I do think this logic can be helpful

twiecki:
Also, I wonder if we should have the data be passed to `fit()` rather than `__init__()` to mimick sklearn's API.

cluhmann:
If that's an option, then you might consider passing `sampler_config` to `fit()` instead of `__init__()` as well.

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,129 @@
...
+
+        with self.model:  # sample with new input data
+            post_pred = pm.sample_posterior_predictive(self.idata.posterior)
+
+        # reshape output
+        post_pred = self._extract_samples(post_pred)
+
+        if point_estimate:  # average, if point-like estimate desired
+            for key in post_pred:
+                post_pred[key] = post_pred[key].mean(axis=0)
```
michaelosthege, line 104: I would advise against switching the kind of return value via a kwarg.  Instead extract this part into a function which can be called on the return value.

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,129 @@
...
+        self = cloudpickle.loads(file.read())
+        filepath = Path(str(filepath) + str(file_prefix) + ".nc")
+        data = az.from_netcdf(filepath)
+        self.idata = data
+        return self
+
+    # fit and predict methods
+    def fit(self, data: pd.DataFrame = None):
+        if data is not None:
+            self.data = data
```
michaelosthege, line 70: What if the model is built for data that is not a single DataFrame?
At least for my models this is almost always the case.

How about making `data` a dictionary that can take DataFrames or NumPy arrays?  (`Dict[Str, Union[np.ndarray, pd.DataFrame, pd.Series]]`

michaelosthege (commented):
There are  several good ideas in here!  With more generaizability I could see some of my models/pipelines be compatible.

I particularly like the ID part.
I'm sceptical about pickling though.

Adding tests that show some of the typical use cases would help tremendously. Maybe take some examples from the documentation and implement them using the class?

michaelosthege:
> Also, I wonder if we should have the data be passed to `fit()` rather than `__init__()` to mimick sklearn's API.

Maybe this distinction could be made for immutable VS.  mutable data?  All immutable things go to the constructor,  while mutable data are passed to the `fit` and `predict` methods..

cluhmann:
That would certainly make sense, but I think it comes down to whether we want to adhere to some existing standard (e.g. sklearn) or do what would make sense in a vacuum.  My recollection is that sklearn model constructors don't really do much except build you an empty shell of a model (so model-building arguments are passed at that point).  Expecting different types of data (which I think are already confusing to some users) to be passed at different points of the pipeline seems like it could be non-intuitive to many users.  Obviously pymc models are different from sklearn models in many ways, but just something to keep in mind.

twiecki:
Sklearn passes global parameters to init, I think with the configuration that's implemented here it's the same thing (and then we don't need a dict but can just turn that into kwargs).

Also, it feels a bit more consistent to pass data to fit() and predict(), rather than to init and predict().

pymc_experimental/model_builder.py
michaelosthege: Be aware of the limitations of `pickle`/`cloudpickle`. Quoting from [the `cloudpickle` README](https://github.com/cloudpipe/cloudpickle#cloudpickle):

> Cloudpickle can only be used to send objects between the **exact same version
of Python**.
>
> Using `cloudpickle` for **long-term object storage is not supported and
strongly discouraged.**
>
> **Security notice**: one should **only load pickle data from trusted sources** as
otherwise `pickle.load` can lead to arbitrary code execution resulting in a critical
security vulnerability.

tomicapretto:
> There are several good ideas in here! With more generaizability I could see some of my models/pipelines be compatible.
> 
> I particularly like the ID part. I'm sceptical about pickling though.
> 
> Adding tests that show some of the typical use cases would help tremendously. Maybe take some examples from the documentation and implement them using the class?

I'm curious about pickling. I'm aware pickling is not the best (or even a good?) alternative to permanently store PyMC models. But I don't know of a suitable alternative. Do you have something in mind?

twiecki:
Pickling is more of a secondary use-case. The primary use-case is: user has access to model class and just loads the netcdf file. In that case, no pickling is required. However, if, for some reason, someone also wants to ship the model class, pickling is supported for that. No access to the model class is required for that. We can turn it off by default.

twiecki:
@tomicapretto I'm curious about Bambi which shares similarities (intentionally). Do you think this would be useful for it?

michaelosthege:
> I'm curious about pickling. I'm aware pickling is not the best (or even a good?) alternative to permanently store PyMC models. But I don't know of a suitable alternative. Do you have something in mind?

I liked the way how Keras stored models in HDF5 and I wonder if a similar thing could be done for an Aesara graph. IIRC they encoded not the code,  but the import route of custom layers,  thereby even supporting custom neural net layers as long as they can be re-imported.
The Aesara equivalent would be the Ops from which the apply nodes and variables are created.

This is of course not easy at all,  and a deep understanding of Aesara is needed.

PyMC-only attributes of the model would have to be included in addition to the Aesara graph. But that's almost trivial compared to serializing arbitrary Aesara graphs to HDF5 or a similar format.

michaelosthege:
@5hv5hvnk plesae make sure to run the `pre-commit` locally (`pre-commit run --all`) and use numpydoc style for the docstrings

pymc_experimental/model_builder.py
tomicapretto: > @tomicapretto I'm curious about Bambi which shares similarities (intentionally). Do you think this would be useful for it?

What we don't have yet is a `.save()` method to save the model and/or the InferenceData object. But apart from that, most of the functionality is currently supported if I understand the aim of this. You can rebuild the Bambi model (which rebuilds the PyMC model) and load the NetCDF file with the InferenceData the model generated. This allows you to keep doing inferences later in time. The problem is, in Bambi, what if you load the wrong combination of model/NetCDF?  Sometimes it will fail, others it may silently do the wrong thing.

What would be better, at least on Bambi's side, is something that stores model configuration, the data, and the NetCDF.

twiecki: That's what the idea of the `id` hash is, even though we currently don't check it yet.

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,277 @@
...
+
+        data: Dictionary of string and either of numpy array, pandas dataframe or pandas Series
+            It is the data we need to set as idata for the model
+        x_only: bool
+            if data only contains values of x and y is not present in the data
+
+        Example: 
+
+        def _data_setter(self, data : pd.DataFrame):
+        with self.model:
```
twiecki, line 91: missing indent

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,277 @@
...
+        self.idata = None  # parameters for
+
+    def _build(self):
+        """
+        Needs to be implemented by the user in the inherited class.
+        Builds user model. Requires suitable self.data and self.model_config. 
+
+        Example:
+        def _build(self):
+        # data
```
twiecki, line 53: missing indent. I also think there is a way to tag that this is python code.

pymc_experimental/tests/test_model_builder.py
```diff
@@ -0,0 +1,128 @@
...
+        a = pm.Normal("a", a_loc, sigma=a_scale)
+        b = pm.Normal("b", b_loc, sigma=b_scale)
+        obs_error = pm.HalfNormal("σ_model_fmc", obs_error)
+
+        y_model = pm.Normal("y_model", a + b * x, obs_error, observed=y_data)
+
+        idata = pm.sample(1000, tune=1000)
+        y_test = pm.sample_posterior_predictive(idata)
+
+        y_test.posterior_predictive.equlas(model.posterior_predictive)
```
twiecki, line 128: Needs an assert. And `equlas`?

pymc_experimental/tests/test_model_builder.py
```diff
@@ -0,0 +1,128 @@
+import pymc as pm
+from pymc.model_builder import ModelBuilder
+import unittest
+import pytest
+import re
+from nose.tools import *
```
twiecki, line 6: I think nose is oldschool and only pytest should be used. Are you using any functionality of nose?

pymc_experimental/tests/test_model_builder.py
5hv5hvnk: `assert` can only work with certain datatypes. `idata` itself is an `xarray.Dataset` which doesn't not work with `assert_equal` . This can be replaced using this. link for documentation of [equals](https://docs.xarray.dev/en/v0.12.3/generated/xarray.Dataset.equals.html)

twiecki:
Saving attrs does not seem to work currently: https://github.com/arviz-devs/arviz/issues/2109

pymc_experimental/model_builder.py
```diff
@@ -24,20 +24,33 @@
 
 
 class ModelBuilder(pm.Model):
-    """
+        """
```
twiecki, line 27: The indentation was correct before.

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        --------
+
+        data: Dictionary of string and either of numpy array, pandas dataframe or pandas Series
+            It is the data we need to set as idata for the model
+        x_only: bool
+            if data only contains values of x and y is not present in the data
+
+        Example: 
+
+        >>>def _data_setter(self, data : pd.DataFrame):
```
twiecki, line 75: Again, wrong indents, please be more mindful here.
```suggestion
        >>> def _data_setter(self, data : pd.DataFrame):
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        >>>    'tune': 1_000,
+        >>>    'chains': 1,
+        >>>    'target_accept': 0.95,
+        >>>    }
+
+        >>>    return data, model_config, sampler_config
+        """
+        raise NotImplementedError
+
+    def save(self, file_prefix, filepath):
```
twiecki, line 120: Change this to just `fname`

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        >>> name = 'mymodel'
+        >>> path = '.'
+        >>> imported_model = LinearModel.load(name,path)
+
+        """
+    
+        filepath = Path(str(filepath) + str(file_prefix) + ".nc")
+        data = az.from_netcdf(filepath)
+        self.idata = data
+        # if model.idata.attrs is not None:
```
twiecki, line 171: Can you link to the issue with the bug?

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        #         self = model
+        #         self.idata = data
+        #         return self
+        #     else:
+        #         raise ValueError(
+        #             f"The route '{file}' does not contain an inference data of the same model '{self.__name__}'"
+        #         )
+        return self.idata
+
+    # fit and predict methods
```
twiecki, line 182:
```suggestion
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        data: Dictionary of string and either of numpy array, pandas dataframe or pandas Series
+            It is the data we need to train the model on.
+
+        Retruns
+        --------
+        returns infernece data of the fitted model.
+
+        Example
+        -------
+        >>>data, model_config, sampler_config = LinearModel.create_sample_input()
```
twiecki, line 199: add space after `>>>`

twiecki: Also in other parts.

pymc_experimental/tests/test_model_builder.py
twiecki: But it spells `equlas`, did you run the tests to see if they pass?

michaelosthege: For the equals assert here, you don't need to run 2000 iterations.

Please do `pm.sample(tune=10, draws=20, chains=3, cores=1, step=pm.Metropolis(), compute_convergence_checks=False)` here and above, so the test runtime doesn't increase too much.

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+            post_pred_dict[key] = post_pred.posterior_predictive[key].to_numpy()[0]
+
+        return post_pred_dict
+
+    def id(self):
+        """
+        It creates a hash value to match the model version using last 16 characters of hash encoding.
+
+        Return
+        ------
```
michaelosthege, line 285:
```suggestion
        Returns
        -------
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+
+        if data_prediction is not None:  # set back original data in model
+            self._data_setter(self.data)
+
+        return post_pred
+
+    @staticmethod
+    def _extract_samples(post_pred: arviz.data.inference_data.InferenceData) -> Dict[str, np.array]:
+        """
+        Parameters
```
michaelosthege, line 266: The docstring needs a title, and an empty line before the `Parameters` section

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+        As the name suggests fit can be used to fit a model using the data that is passed as a parameter.
+        Sets attrs to inference data of the model
+
+        Parameter
+        ---------
+        data: Dictionary of string and either of numpy array, pandas dataframe or pandas Series
+            It is the data we need to train the model on.
+
+        Retruns
+        --------
```
michaelosthege, line 188:
```suggestion
        Retruns
        -------
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        Loads infernce data for the model.
+
+        Parameters
+        ----------
+        fname: string
+            This denotes the name with path from where idata should be loaded from.
+
+        Return
+        ------
```
michaelosthege, line 148:
```suggestion
        Returns
        -------
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        Saves inference data of the model.
+
+        Parameters
+        ----------
+        fname: string
+            This denotes the name with path from where idata should be saved.
+
+        Example
+        ------
```
michaelosthege, line 129:
```suggestion
        Example
        -------
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+        self.idata = None  # parameters for
+
+    def _data_setter(
+        self, data: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]], x_only: bool = True
+    ):
+        """
+        Sets new data in the model.
+
+        Parameter
+        --------
```
michaelosthege, line 66:
```suggestion
        Parameter
        ---------
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+
+        data: Dictionary of string and either of numpy array, pandas dataframe or pandas Series
+            It is the data we need to set as idata for the model
+        x_only: bool
```
michaelosthege, line 70:
```suggestion
        data : Dictionary of string and either of numpy array, pandas dataframe or pandas Series
            It is the data we need to set as idata for the model
        x_only : bool
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+            if data only contains values of x and y is not present in the data
+
+        Example:
```
michaelosthege, line 73: Please check in the core PyMC repo how example sections should be formatted.

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+    
+    """
+
+    _model_type = "BaseClass"
+    version = "None"
+
+    def __init__(self, model_config: Dict, sampler_config: Dict):
+        """
+        Initialises model configration and sampler configration for the model
+        Parameters
```
michaelosthege, line 41:
```suggestion
        Initialises model configration and sampler configration for the model
        
        Parameters
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+class ModelBuilder(pm.Model):
+    """
+    Extention of pm.Model class to improve workflow.
+
+    ModelBuilder class can be used to play around models with ease using direct API calls
+    for multiple tasks that one need to deploy a model.  
+    
+    """
```
michaelosthege, line 33:
```suggestion
    for multiple tasks that one need to deploy a model.
    """
```

5hv5hvnk:
@michaelosthege Can you go through the lastest commit and share if any edits have to be made?

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+    _model_type = "BaseClass"
+    version = "None"
+
+    def __init__(self, model_config: Dict, sampler_config: Dict):
+        """
+        Initialises model configration and sampler configration for the model
+
+        Parameters
+        ----------
+        model_confid : Dictionary
```
twiecki, line 42:
```suggestion
        model_config : Dictionary
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+            dictonary of parameters that initialise model configration. Genrated by the user defiend create_sample_input method.
```
twiecki, line 43:
```suggestion
            dictionary of parameters that initialize model configuration. Generated by the user defined create_sample_input method.
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+
+        raise NotImplementedError
+
+    def save(self, fname):
+        """
+        Saves inference data of the model.
+
+        Parameters
+        ----------
+        fname: string
```
twiecki, line 126:
```suggestion
        fname : string
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        self.idata.to_netcdf(file)
+
+    @classmethod
+    def load(cls, fname):
+        """
+        Loads infernce data for the model.
+
+        Parameters
+        ----------
+        fname: string
```
twiecki, line 145:
```suggestion
        fname : string
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        Sets attrs to inference data of the model.
+
+        Parameter
+        ---------
+        data : Dictionary of string and either of numpy array, pandas dataframe or pandas Series
+            It is the data we need to train the model on.
+
+        Retruns
+        -------
+        returns infernece data of the fitted model.
```
twiecki, line 193:
```suggestion
        returns inference data of the fitted model.
```

pymc_experimental/model_builder.py
```diff
@@ -94,7 +94,7 @@ def create_sample_input(cls):
         --------
         >>> @classmethod
         >>> def create_sample_input(cls):
-        >>>    x = np.linspace(start=1, stop=50, num=100)
+        >>>    x = np.linespace(start=1, stop=50, num=100)
```
twiecki, line 97:
```suggestion
        >>>    x = np.linspace(start=1, stop=50, num=100)
```

michaelosthege:
@5hv5hvnk please visit the open threads and close those that you have done

pymc_experimental/model_builder.py
twiecki: I see your point, but I also don't think it's terrible style in this setting.

pymc_experimental/tests/test_model_builder.py
```diff
@@ -0,0 +1,156 @@
...
+import pymc as pm
+import pytest
+from pymc.model_builder import ModelBuilder
+
+
+class test_ModelBuilder(ModelBuilder):
+    _model_type = "LinearModel"
+    version = "0.1"
+
+    @classmethod
```
twiecki, line 16:
```suggestion
```

twiecki:
@5hv5hvnk

```
************* Module pymc_experimental.tests.test_model_builder
pymc_experimental/tests/test_model_builder.py:1: [W0611(unused-import), ] Unused import unittest
pymc_experimental/tests/test_model_builder.py:2: [W0611(unused-import), ] Unused Dict imported from typing
pymc_experimental/tests/test_model_builder.py:4: [W0611(unused-import), ] Unused arviz imported as az
pymc_experimental/tests/test_model_builder.py:8: [W0611(unused-import), ] Unused import pytest
************* Module pymc_experimental.model_builder
pymc_experimental/model_builder.py:19: [W0404(reimported), ] Reimport 'arviz' (imported line 18)
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused TYPE_CHECKING imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused Any imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused Callable imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused List imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused Optional imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused Sequence imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused Tuple imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused Type imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused TypeVar imported from typing
pymc_experimental/model_builder.py:3: [W0611(unused-import), ] Unused cast imported from typing
pymc_experimental/model_builder.py:20: [W0611(unused-import), ] Unused import cloudpickle
```

twiecki:
You should install pre-commit.

twiecki:
Why aren't the tests being run? CC @ricardoV94 @michaelosthege

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,295 @@
...
+        Needs to be implemented by the user in the inherited class.
+        Returns examples for data, model_config, sampler_config.
+        This is useful for understanding the required
+        data structures for the user model.
+
+        Examples
+        --------
+        >>> @classmethod
+        >>> def create_sample_input(cls):
+        >>>    x = np.linespace(start=1, stop=50, num=100)
```
twiecki, line 93:
```suggestion
        >>>    x = np.linspace(start=1, stop=50, num=100)
```

pymc_experimental/tests/test_model_builder.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def _data_setter(self, data: pd.DataFrame):
+        with self.model:
+            pm.set_data({"x": data["input"].values})
+            if "output" in data.columns:
+                pm.set_data({"y_data": data["output"].values})
+
+    @classmethod
+    def create_sample_input(cls):
+        x = np.linespace(start=0, stop=1, num=100)
```
twiecki, line 41:
```suggestion
        x = np.linspace(start=0, stop=1, num=100)
```

pymc_experimental/tests/test_model_builder.py
```diff
@@ -0,0 +1,151 @@
...
+
+    data, model_config, sampler_config = test_ModelBuilder.create_sample_input()
+    model_2 = test_ModelBuilder(data, model_config, sampler_config)
+    model_2.idata = model_2.fit(data)
+    assert str(model_2.idata.groups) == str(idata.groups)
+
+
+def test_fit():
+    with pm.Model() as model:
+        x = np.linespace(start=0, stop=1, num=100)
```
twiecki, line 95:
```suggestion
        x = np.linspace(start=0, stop=1, num=100)
```

pymc_experimental/tests/test_model_builder.py
```diff
@@ -0,0 +1,151 @@
...
+            "chains": 3,
+            "target_accept": 0.95,
+        }
+
+        return data, model_config, sampler_config
+
+
+def test_fit():
+    with pm.Model() as model:
+        x = np.linespace(start=0, stop=1, num=100)
```
twiecki, line 66:
```suggestion
        x = np.linspace(start=0, stop=1, num=100)
```

pymc_experimental/tests/test_model_builder.py
```diff
@@ -0,0 +1,151 @@
...
+
+def test_predict():
+    x_pred = np.random.uniform(low=0, high=1, size=100)
+    prediction_data = pd.DataFrame({"input": x_pred})
+    data, model_config, sampler_config = test_ModelBuilder.create_sample_input()
+    model_2 = test_ModelBuilder(model_config, sampler_config, data)
+    model_2.idata = model_2.fit()
+    model_2.predict(prediction_data)
+    with pm.Model() as model:
+        x = np.linespace(start=0, stop=1, num=100)
```
twiecki, line 130:
```suggestion
        x = np.linspace(start=0, stop=1, num=100)
```

5hv5hvnk:
Also fixes #81

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+        ----------
+        model_config : Dictionary
+            dictionary of parameters that initialise model configuration. Generated by the user defined create_sample_input method.
+        sampler_config : Dictionary
+            dictionary of parameters that initialise sampler configuration. Generated by the user defined create_sample_input method.
+        data : Dictionary
+            It is the data we need to train the model on.
+        Examples
+        --------
+        >>> class LinearModel(ModelBuilder):
```
ferrine, line 40: there was a totally valid error in the CI
https://github.com/pymc-devs/pymc-experimental/issues/81

The identation in incorrect, the code block should look like this:

```
>>> class LinearModel(ModelBuilder):
>>>     ...
```
not 
```
>>> class LinearModel(ModelBuilder):
>>> ...
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Examples
+        --------
+        >>> @classmethod
+        >>> def create_sample_input(cls):
+        >>>    x = np.linspace(start=1, stop=50, num=100)
+        >>>    y = 5 * x + 3 + np.random.normal(0, 1, len(x)) * np.random.rand(100)*10 +  np.random.rand(100)*6.4
+        >>>    data = pd.DataFrame({'input': x, 'output': y})
+
+        >>>    model_config = {
```
ferrine, line 102: pretty identation should be applied to dictionaries

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+        Saves inference data of the model.
+
+        Parameters
+        ----------
+        fname : string
+            This denotes the name with path from where idata should be saved.
+
+        Examples
+        --------
+        >>> class LinearModel():
```
ferrine, line 132: the identation problem as mentioned above

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+        Returns the inference data that is loaded from local system.
+
+        Raises
+        ------
+        ValueError
+            If the inference data that is loaded doesn't match with the model.
+
+        Examples
+        --------
+        >>> class LinearModel():
```
ferrine, line 165: same issue

ferrine (changes requested):

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+import numpy as np
+import pandas as pd
+import pymc as pm
+
+
+class ModelBuilder(pm.Model):
+    """
+    Extension of pm.Model class to improve workflow.
+    ModelBuilder class can be used to play around models with ease using direct API calls
+    for multiple tasks that one need to deploy a model.
```
twiecki, line 15:
```suggestion
    ModelBuilder class can be used to provide an easy-to-use API (similar to scikit-learn) for models
    and help with deployment.
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+
+import arviz as az
+import numpy as np
+import pandas as pd
+import pymc as pm
+
+
+class ModelBuilder(pm.Model):
+    """
+    Extension of pm.Model class to improve workflow.
```
twiecki, line 13:
```suggestion
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+    ModelBuilder class can be used to play around models with ease using direct API calls
+    for multiple tasks that one need to deploy a model.
+    """
```
twiecki, line 16:
```suggestion
    Extends the pymc.Model class.
    """
```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,309 @@
...
+import arviz as az
+import numpy as np
+import pandas as pd
+import pymc as pm
+
+
+class ModelBuilder(pm.Model):
+    """
+    ModelBuilder class can be used to provide an easy-to-use API (similar to scikit-learn) for models
+    and help with deployment.
```
twiecki, line 14:
```suggestion
    and help with deployment.

```

pymc_experimental/model_builder.py
```diff
@@ -0,0 +1,310 @@
...
+
+import arviz as az
+import numpy as np
+import pandas as pd
+import pymc as pm
+
+
+class ModelBuilder(pm.Model):
+    """
+    ModelBuilder class can be used to provide an easy-to-use API (similar to scikit-learn) for models
```
twiecki, line 13:
```suggestion
    ModelBuilder can be used to provide an easy-to-use API (similar to scikit-learn) for models
```

twiecki:
Congrats @5hv5hvnk!!! This was a big one.

PR #66, opened 2022-08-31
Add a `DirichletProcessMixture` class

larryshamalama:
This PR shows the beginnings of a (Truncated) Dirichlet Process Mixture API whose two main ingredients are the `pm.StickBreakingWeights` distribution, also known as the GEM distribution, and the `pm.Mixture` distribution.

The truncation of the infinite sum allows for easier implementation at the cost of numerical error which shouldn't be an issue for large `K`. However, computational cost can increase; some Goldilocks principle should be applied when choosing `K`.

More to come as I work on some utility functions and polish the distribution class.


fonnesbeck:
Stale

PR #67, opened 2022-09-07
fix unitary case in prior from trace

ferrine:
fixes https://discourse.pymc.io/t/posterior-values-in-factored-pdf/10280/4


ricardoV94 (approved):

ferrine:
these PRs are related
https://github.com/pymc-devs/pymc/pull/6104
+ the one in pmx that fixes import in bart test

PR #75, opened 2022-09-09
remove bart

aloctavodia:
Removes BART from pymc-experimental as it has been moved to its own repository (https://github.com/pymc-devs/pymc-bart). As we have not released pymc-experimental I am directly removing without deprecation warning.


GarrettMooney:
Is this BART implementation being moved to another package?

GarrettMooney:
>  Is this BART implementation being moved to another package?

Nevermind, [found it](https://github.com/pymc-devs/pymc-bart) 😆

PR #80, opened 2022-09-12
Blackjax pathfinder

twiecki:
Second attempt. Not sure why CI is trying to import blackjax on windows.


ricardoV94:
Wild guess but there might be a typo in the yml (and that of PyMC): https://github.com/pymc-devs/pymc-experimental/blob/14e24063a9da3a624dff2e5c84528dedf7e522fe/.github/workflows/test.yml#L134

I think it should not have the `$` at the end of `%TEST_SUBSET%`

PR #84, opened 2022-09-26
Genextreme

ccaprani:
This is a follow-on from #4 which after rebasing closed the PR. It would be good to merge and close this out as it has been a long time. @ricardoV94 please advise if anything outstanding.


pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,218 @@
...
+            at.isclose(xi, 0),
+            -at.log(sigma) - scaled - at.exp(-scaled),
+            -at.log(sigma)
+            - ((xi + 1) / xi) * at.log1p(xi * scaled)
+            - at.pow(1 + xi * scaled, -1 / xi),
+        )
+        # bnd = mu - sigma/xi
+        return check_parameters(
+            logp_expression,
+            1 + xi * (value - mu) / sigma > 0,
```
ricardoV94, line 184: This constraint depends on mu so it should probably be in a `switch(constraint, logp, -inf)`. Usually value-dependent constraints are treated elemwise while parameter based constraints are treated blockwise with `check_parameters` (if one fails, everything fails), and we raise an error outside of sampling.

```suggestion
            1 + xi * (value - mu) / sigma > 0,
```

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Returns
+        -------
+        TensorVariable
+        """
+        scaled = (value - mu) / sigma
+        logc_expression = at.switch(
+            at.isclose(xi, 0), -at.exp(-scaled), -at.pow(1 + xi * scaled, -1 / xi)
+        )
+        return check_parameters(logc_expression, 1 + xi * scaled > 0, sigma > 0)
```
ricardoV94, line 209: Same here with the first constraint

pymc_experimental/tests/test_distributions.py
```diff
@@ -0,0 +1,50 @@
...
+    TestMatchesScipy,
+)
+
+# the distributions to be tested
+from pymc_experimental.distributions import (
+    GenExtreme,
+)
+
+
+class TestMatchesScipyX(TestMatchesScipy):
```
ricardoV94, line 32: We removed the `check_logp`/`check_logcdf` into https://github.com/pymc-devs/pymc/blob/a38bead63f71e3dac1c589c73acb98174f3b8e59/pymc/tests/distributions/util.py#L230, so you no longer need to subclass the old `TestMatchesScipy`. Just use a `TestGenExtreme` class from scratch.

pymc_experimental/tests/test_distributions_moments.py
```diff
@@ -0,0 +1,51 @@
...
+                (3, 6),
+                np.arange(6)
+                + np.arange(1, 7)
+                * ((1 + np.linspace(-0.2, 0.2, 6)) ** -np.linspace(-0.2, 0.2, 6) - 1)
+                / np.linspace(-0.2, 0.2, 6),
+            ),
+        ),
+    ],
+)
+def test_genextreme_moment(mu, sigma, xi, size, expected):
```
ricardoV94, line 48: This can go together with the other testclass mentioned above

pymc_experimental/tests/test_distributions.py
ricardoV94: Also, the tests in the main repo have been refactored, and now fall in the folder where the distribution is implemented, so in this case would be `/tests/distributions/test_continuous`

pymc_experimental/tests/test_distributions_random.py
```diff
@@ -0,0 +1,38 @@
...
+from pymc.tests.test_distributions_random import (
+    BaseTestDistribution,
+    seeded_scipy_distribution_builder,
+)
+
+# the distributions to be tested
+import pymc_experimental as pmx
+
+
+class TestGenExtreme(BaseTestDistribution):
```
ricardoV94, line 27: This would also go in the same class/file as the other tests

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,218 @@
...
+
+    """
+
+    rv_op = gev
+
+    @classmethod
+    def dist(cls, mu=0, sigma=1, xi=0, scipy=False, **kwargs):
+        # If SciPy, use its parametrization, otherwise convert to standard
+        if scipy:
+            xi = -xi
```
ricardoV94, line 150: Can you add a quick test for this kwarg? Something like this should suffice:

```python
def test_gen_extreme_scipy_kwarg():
  dist = GenExtreme.dist(xi=1, scipy=False):
  assert dist.owner.inputs[-1].eval() == 1

  dist = GenExtreme.dist(xi=1, scipy=True):
  assert dist.owner.inputs[-1].eval() == -1
```

ricardoV94 (changes requested):
Thanks for coming back to this. Looks great. The tests in the main repo have been refactored to follow the same path as the code. I suggest you update it so it's already pymc compatible. Otherwise I left a small suggestion for a direct test of the `scipy` kwarg.

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def __call__(
+        self, mu=0.0, sigma=1.0, xi=0.0, size=None, **kwargs
+    ) -> TensorVariable:
+        return super().__call__(mu, sigma, xi, size=size, **kwargs)
+
+    @classmethod
+    def rng_fn(
+        cls,
+        rng: np.random.RandomState,
```
ricardoV94, line 50: Generators are also allowed as rng (and the default actually)
```suggestion
        rng: Union[np.random.RandomState, np.random.Generator]
```

Need to import Union from typing

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,231 @@
...
+        logp_expression = at.switch(
+            at.isclose(xi, 0),
+            -at.log(sigma) - scaled - at.exp(-scaled),
+            -at.log(sigma)
+            - ((xi + 1) / xi) * at.log1p(xi * scaled)
+            - at.pow(1 + xi * scaled, -1 / xi),
+        )
+        # Confirm in valid domain
+        logp = at.switch(
+            1 + xi * (value - mu) / sigma > 0,
```
ricardoV94, line 183: Usually we use at.gt, because Aesara is  abit finicky as to what can be written like this and what cannot (e.g, `==` does not work)

```suggestion
            at.gt((1 + xi * (value - mu) / sigma), 0),  
```

Same for the logcdf

ricardoV94 (changes requested):
Two more small things

pymc_experimental/distributions/continuous.py
ricardoV94: What do you think about this one?

pymc_experimental/distributions/__init__.py
```diff
@@ -1,2 +1,26 @@
-from pymc_experimental.distributions import histogram_utils
-from pymc_experimental.distributions.histogram_utils import histogram_approximation
+#   Copyright 2020 The PyMC Developers
```
junpenglao, line 1:
```suggestion
#   Copyright 2022 The PyMC Developers
```

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,231 @@
+#   Copyright 2020 The PyMC Developers
```
junpenglao, line 1:
```suggestion
#   Copyright 2022 The PyMC Developers
```

pymc_experimental/tests/distributions/test_continuous.py
```diff
@@ -0,0 +1,108 @@
+
+#   Copyright 2020 The PyMC Developers
```
junpenglao, line 2:
```suggestion
#   Copyright 2022 The PyMC Developers
```

pymc_experimental/distributions/continuous.py
ccaprani: I've implemented this now, but not precisely sure the purpose: `check_parameters` is value-based, while `switch(constraint...` is tensor-based, right? Anyway, please have a look.

pymc_experimental/distributions/continuous.py
ccaprani: Done.

pymc_experimental/tests/test_distributions.py
ccaprani: Yes, I've moved all those functions now into one place.

However, the tests are failing. Parameters values outside of the set `Domain` are being generated for `xi`, and so it is failing at [this point](https://github.com/pymc-devs/pymc/blob/372d7c247ac55bd90ed999031a4e207fa24757be/pymc/tests/distributions/util.py#L359). I don't know why values of `xi=-2.0` are being proposed, for example. It is as if the new `check_logp` is not evaluating or respecting the set Domain? Example output:
```
Failed: test_params={'mu': array(-2.1), 'sigma': array(0.01), 'xi': TensorConstant{-2.0}}, valid_
value=-2.1
```
and yet when I implement my own test, it passes (code without obvious imports):
```python
μ,σ,ξ,x = -2.1,0.01,-2.0,-2.1
mu = at.as_tensor_variable(floatX(μ))
sigma = at.as_tensor_variable(floatX(σ))
xi = at.as_tensor_variable(floatX(ξ))
logp_pm = pmx.distributions.GenExtreme.logp(x,mu,sigma,xi).eval()
logp_sp = sp.genextreme.logpdf(x, c=-ξ, loc=μ, scale=σ)
if not np.allclose(logp_pm,logp_sp):
    print(f"Fail on {μ=}, {σ=}, {ξ=}")
```
So I need some input at this point for the tests - thanks!

pymc_experimental/distributions/continuous.py
ccaprani: Done. Thanks.

pymc_experimental/distributions/continuous.py
ricardoV94: The point is that the logp is usually defined for every possible `value` (although it can be zero for some combinations of value parameters, such as a halfnormal evaluated at -1), so it should never "error" out. Logps can however be undefined for some parameter values, such as a halfnormal with negative sigma. We do that distinction in `PyMC` by using a switch for values and check_parameters for the second.

pymc_experimental/distributions/continuous.py
ricardoV94: I marked this as unresolved. I didn't see any test targeting the `scipy` flag.

ccaprani: Line 99 of test_continuous.py.

pymc_experimental/distributions/continuous.py
ricardoV94: You can increase the test tolerange in `check_logp/ check_locdf`, with the kwarg `decimal`. Do note that you may only be checking a subset of parameter combinations when testing locally, so you can temporarily set `n_samples=-1` to make sure all parameter combinations are tested

ricardoV94: This is all detailed in the slightly outdated developer guide: https://www.pymc.io/projects/docs/en/latest/contributing/developer_guide_implementing_distribution.html#adding-tests-for-the-logp-logcdf-methods

ricardoV94: About the checks, is `xi` forced to be between -1 and 1? If so it should be added to `check_parameters`. I can't quite read your interactive output. Calling `check_parameters` should never fail, but it is a symbolic expression. If you want to see if the Assert passes for some inputs you need to `.eval()` it as well.

`at.and_` is the elemwise `and`

ccaprani: > You can increase the test tolerange in check_logp/ check_locdf, with the kwarg `decimal`.

`decimal` is an absolute check, defaulting to 6 for float64. So it compares numbers like -1.898391e+61 to a precision of 1e-6 which is 67 orders of magnitude and clearly unnecessary. A relative check like the one I mentioned would compare the ratios of the two numbers to a relative accuracy of 1e-6 instead, which makes more sense.

ccaprani: @ricardoV94 good news: Using `at.and_` in `check_parameters` did the trick for the restricted (but practical) domain of $\xi \in [-1,1]$. All checks now pass!

I can expand to the full domain of $\xi \in \mathbb{R}$ if the logp check is made relative instead of absolute as mentioned above.

In any case, thank you so much for your help on this. I think it's been 11 months in the pipeline, so hopefully we can get it merged?!

pymc_experimental/distributions/continuous.py
ricardoV94: Thanks, I missed it!

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,236 @@
...
+
+        logp = at.switch(
+            at.gt(1 + xi * scaled, 0.0),
+            logp_expression,
+            -np.inf)
+
+        return check_parameters(
+            logp,
+            sigma > 0,
+            1 + xi * scaled > 0,
```
ricardoV94, line 191: No need to duplicate
```suggestion
```

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,236 @@
...
+            at.and_(xi > -1, xi < 1),
+            msg="sigma <= 0 or 1+xi*(x-mu)/sigma <= 0")
```
ricardoV94, line 193:
```suggestion
            msg="sigma > 0, -1 < xi < 1)
```

ricardoV94 (changes requested):

pymc_experimental/distributions/continuous.py
ccaprani: Because this is done on line 184?

ccaprani: Done!

pymc_experimental/distributions/continuous.py
ccaprani: And done.

pymc_experimental/distributions/continuous.py
ricardoV94: >  A relative check like the one I mentioned would compare the ratios of the two numbers to a relative accuracy of 1e-6 instead, which makes more sense.

I have thought about this sometime ago, but then faced issues. logps vary widely (and non-linearly) between -700xs and +x. More annoying, it's not uncommon to have logps around 0, in which a relative comparison or, on the other hand, absolute comparisons with higher threshold become pretty meaningless.

If you have a more flexible solution that works in general that would be great. Certainly there are more principled ways of doing these accuracy checks.

ricardoV94 (changes requested):

pymc_experimental/distributions/continuous.py
```diff
@@ -0,0 +1,234 @@
...
+        logp = at.switch(
+            at.gt(1 + xi * scaled, 0.0),
+            logp_expression,
+            -np.inf)
+
+        return check_parameters(
+            logp,
+            sigma > 0,
+            at.and_(xi > -1, xi < 1),
+            msg="sigma <= 0 or -1 < xi < 1")
```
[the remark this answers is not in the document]
ccaprani: @ricardoV94 Ok, I've reversed them. But in many places the message is written in the opposite direction to the check it looks like, e.g. [Gumbel](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L3320), [Logistic](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L3545), [Normal](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L584), [HalfNormal](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L909), [Beta](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L1287). Others are written in the same direction, e.g. [TruncatedNormal](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L793), [Wald](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L1095), [Kumaraswamy](https://github.com/pymc-devs/pymc/blob/b12c189896e260cb743d31a47a5d4680f2358216/pymc/distributions/continuous.py#L1389) and so on.

pymc_experimental/distributions/continuous.py
ccaprani: @ricardoV94 `npt.assert_allcose` [here](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) allows for both a relative, `rtol` *and* absolute `atol` tolerance to be set. Example:
```python
>>> x = np.pi*np.array([1e2,1e-1,1e-6,1e-10,1e-100,1e-700])
>>> y = (1 + 1e-6)*x
>>> npt.assert_allclose(x,y)  # will assert
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 5 / 6 (83.3%)
Max absolute difference: 0.00031416
Max relative difference: 9.99999e-07
 x: array([3.141593e+002, 3.141593e-001, 3.141593e-006, 3.141593e-010,
       3.141593e-100, 0.000000e+000])
 y: array([3.141596e+002, 3.141596e-001, 3.141596e-006, 3.141596e-010,
       3.141596e-100, 0.000000e+000])
```
but this does not assert:
```python
>>> npt.assert_allclose(x,y,rtol=1e-6,atol=1e-10)
```

ccaprani:
@ricardoV94 it fails the tests for float32 due to differences between scipy and pymc underflows. Shall I add:
```
    @pytest.mark.skipif(
        condition=(aesara.config.floatX == "float32"),
        reason="PyMC underflows earlier than scipy on float32",
    )
```
to the tests for logp and logc as per other distributions?

ricardoV94:
> @ricardoV94 it fails the tests for float32 due to differences between scipy and pymc underflows. Shall I add:
> ```
>     @pytest.mark.skipif(
>         condition=(aesara.config.floatX == "float32"),
>         reason="PyMC underflows earlier than scipy on float32",
>     )
> ```
> to the tests for logp and logc as per other distributions?

Sure

pymc_experimental/distributions/continuous.py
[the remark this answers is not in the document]
ricardoV94: We should standardize it then... Do you mind opening an issue on the main repo for that?

pymc_experimental/distributions/continuous.py
ricardoV94: Do want to try to open a PR on the main repo to change the logic then?

pymc_experimental/distributions/continuous.py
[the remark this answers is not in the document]
ccaprani: Doine :-) https://github.com/pymc-devs/pymc/issues/6158

pymc_experimental/distributions/continuous.py
ccaprani: Done: https://github.com/pymc-devs/pymc/pull/6159

ccaprani:
The test on the windows machine is failing with message:

```
E           AssertionError: 
E           Arrays are not almost equal to 6 decimals
E           {'mu': array(0.01), 'sigma': array(0.01), 'xi': array(0.), 'value': array(-1.)}
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference: 9.90352031e+27
E           Max relative difference: 1.35533585e-16
E            x: array(-7.30706e+43)
E            y: array(-7.30706e+43)
```
You can see why I reckon relative tolerance would be good! https://github.com/pymc-devs/pymc/pull/6159 Can this be accepted without this test passing as is?

ricardoV94:
> The test on the windows machine is failing with message:
> 
> ```
> E           AssertionError: 
> E           Arrays are not almost equal to 6 decimals
> E           {'mu': array(0.01), 'sigma': array(0.01), 'xi': array(0.), 'value': array(-1.)}
> E           Mismatched elements: 1 / 1 (100%)
> E           Max absolute difference: 9.90352031e+27
> E           Max relative difference: 1.35533585e-16
> E            x: array(-7.30706e+43)
> E            y: array(-7.30706e+43)
> ```
> 
> You can see why I reckon relative tolerance would be good! [pymc-devs/pymc#6159](https://github.com/pymc-devs/pymc/pull/6159) Can this be accepted without this test passing as is?

Let's just extend the `check_logp/logcdf` to allow to use rtol already. Your PR on PyMC does not allow it yet right?

ricardoV94:
Those are also very extreme values (pun not intended), tweaking the grid of `mu`, `sigma` would certainly avoid, so if you want to wrap this up quicker that's another option.

ccaprani:
@ricardoV94 that's a fantastic idea - thank you. I've just implemented that now.

ricardoV94:
Looks good, I pushed a commit that separates the logp and logcdf tests and adds an xfail only for float32. I tested with `n_samples=-1` so it should be stable from now on

ccaprani:
Wonderful to get this finally done @ricardoV94 ; the first PR on this was [a year ago](https://github.com/pymc-devs/pymc/pull/5085)! The [corresponding example](https://github.com/pymc-devs/pymc-examples/pull/241) was also just merged, so many thanks for all your help indeed!

ricardoV94:
Hopefully the next one will be smoother. Thanks for sticking around

ccaprani:
No worries - a big learning curve with long gaps between efforts! Thanks for your patience!

PR #90, opened 2022-10-14
added predict_posterior

### LINKED ISSUE #88: Tweaks to `predict` method of `ModelBuilder`
Remove `point_estimate` kwarg of the `ModelBuilder.predict` method. Instead have two distinct methods... 
* `predict` (which returns the full idata) and 
* `predict_point` which returns the point estimate.

Tagging @michaelosthege, @twiecki.

5hv5hvnk:
fixes #88
@twiecki @drbenvincent

### DIFF
--- a/pymc_experimental/model_builder.py
+++ b/pymc_experimental/model_builder.py
@@ -287,7 +287,7 @@ class ModelBuilder(pm.Model):
         data_prediction: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]] = None,
     ):
         """
-        Uses model to predict samples on unseen data.
+        Uses model to predict samples on unseen data.`
 
         Parameters
         ---------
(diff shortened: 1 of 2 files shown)


twiecki:
As we didn't need to adapt any of the tests it signals we didn't have enough coverage. Could you add some tests for the new method?

5hv5hvnk:
Yes sure let me add tests for the new method and edit the previous tests

pymc_experimental/model_builder.py
```diff
@@ -287,7 +287,7 @@ def predict_posterior(
         data_prediction: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]] = None,
     ):
         """
-        Uses model to predict samples on unseen data.
+        Uses model to predict samples on unseen data.`
```
twiecki, line 290:
```suggestion
        Uses model to predict samples on unseen data.
```

PR #91, opened 2022-11-14
Automatic marginalization of finite discrete variables in the logp

ricardoV94:
Added a Model transformation, to automatically marginalize discrete variables from the logp graph. 

I first attempted to modify the original model in-place so that the `FiniteDiscreteMarginalRVs` are introduced in the model RV graph, but this led to many issues, in the `initial_point`, `step sampler assignment` and so on, which don't really need to know about the modified logp graph.

My latest approach is to clone the underlying model, modify that in place and request its logp, everytime the logp of the user model is requested.

TODO
- [x] Marginalize finite discrete RVs and replace by SymbolicRV subgraph
- [x] Suppress annoying RVs missing in graph
  - Fixed by making the marginalized RV part of the subgraph 
- [x] Test nested marginalized RVs
- [x] Test multiple dependent RVs
  -  Aeppl is too restrictive in that it expects one logp-term per value variable, can we group them in a tuple?
- [x] Test transformed dependent RV 
  - Currently not supported due to Aeppl limitations 
- [x] ~~Dispatch default_transform and moment~~ 
  -  No longer needed because we don't modify the original model
- [ ] Add sampling procedure to retrieve marginalized RVs from prior and posterior
  - Will do in follow up #94 
- [x] Allow Deterministics that don't depend on marginalized RVs
- [x] Allow Potentials that don't depend on marginalized RVs
- [x] ~~Fix graphviz for marginalized RVs~~ 
  - No longer a problem because we don't modify the original model
- [x] ~~Fix string representation for marginalized RVs~~ 
  - No longer a problem because we don't modify the original model
- [x] Finish tests
- [x] Finish docstrings
- [x] Update PyMC dependency


ricardoV94:
This is ready for review, I decided to leave the posterior recovery of marginalized RVs for a subsequent PR as this one is already big as is

notebooks/marginalized_changepoint_model.ipynb
```diff
@@ -0,0 +1,821 @@
+{
```
twiecki, line 1: <p>Why is disasters_missing still being sampled?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymcx/pull/91/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

ricardoV94: Those variables were not marginalized, only the switchpoint

ricardoV94: Also worth pointing that models with bernoulli/categoricals/shorter discrete_uniform benefit more, because the logp is not much slower than the unmarginalized model. Here the slowdown comes from having to use a Scan. I just used this example because it was already out there, and discussed among the STAN community.

Also models where you get rid of more RVs should see better benefits on the NUTS side, rather than a model with a single one like this.

twiecki: would be nice to have an example where we can use full NUTS after marginalization.

aloctavodia: A simple, dirty fix will be to add the missing  observations to the observed data.

One question. How difficult is to recover the marginalized parameters?

ricardoV94: Yes, we don't need to sample the missing years, we could also just infer/recover them in posterior predictive.

Still the original model with metropolis samples pretty well, and I see no reason why a full NUTS is going to do much better.

***

About recovering the marginalized parameters, it's not that bad. We need to sample from the marginalized elemwise logp conditioned on the remaining posterior parameters. Just some Aesara wrangling...

ricardoV94: I'm all for a more interesting example. Let's open an issue for that?

pymc_experimental/marginal_model.py
```diff
@@ -0,0 +1,440 @@
...
+
+    input_rvs = [*marginalized_rv_input_rvs, *dependent_rvs_input_rvs]
+    rvs_to_marginalize = [rv_to_marginalize, *dependent_rvs]
+
+    outputs = rvs_to_marginalize
+    # Clone replace inner RV rng inputs so that we can be sure of the update order
+    # replace_inputs = {rng: rng.type() for rng in updates_rvs_to_marginalize.keys()}
+    # Clone replace outter RV inputs, so that their shared RNGs don't make it into
+    # the inner graph of the marginalized RVs
+    # FIXME: This shouldn't be needed!
```
zaxtax, line 334: Should this FIXME be addressed before merging?

pymc_experimental/marginal_model.py
```diff
@@ -0,0 +1,440 @@
...
+
+    # Some inputs are not root inputs (such as transformed projections of value variables)
+    # Or cannot be used as inputs to an OpFromGraph (shared variables and constants)
+    inputs = list(inputvars(inputs))
+
+    rvs_to_values = {}
+    dummy_marginalized_value = marginalized_rv.clone()
+    rvs_to_values[marginalized_rv] = dummy_marginalized_value
+    rvs_to_values.update(zip(dependent_rvs, values))
+    logps_dict = factorized_joint_logprob(rv_values=rvs_to_values, **kwargs)
```
zaxtax, line 385: It might be worth explaining what lines below do.

pymc_experimental/marginal_model.py
```diff
@@ -0,0 +1,440 @@
...
+    marginalized_rvs_node = op.make_node(*inputs)
+    marginalized_rv, *dependent_rvs = clone_replace(
+        op.inner_outputs,
+        replace={u: v for u, v in zip(op.inner_inputs, marginalized_rvs_node.inputs)},
+    )
+
+    # Some inputs are not root inputs (such as transformed projections of value variables)
+    # Or cannot be used as inputs to an OpFromGraph (shared variables and constants)
+    inputs = list(inputvars(inputs))
+
```
zaxtax, line 380: It might be worth explaining what lines below do.

pymc_experimental/marginal_model.py
ricardoV94: It's an Aesara thing, so definitely not in this PR

pymc_experimental/marginal_model.py
ricardoV94: Definitely, I will add some

notebooks/marginalized_changepoint_model.ipynb
zaxtax: Definitely should be another issue. What about a Cormack-Jolly-Seber capture-recapture model? Austin's blog post basically does what this feature could allow you to do automatically.

https://mc-stan.org/docs/stan-users-guide/mark-recapture-models.html
https://austinrochford.com/posts/2018-01-31-capture-recapture.html

Not a blocker for merging, but could be good first contribution!

ricardoV94: If that kind of model is supported by the current marginalization, it would be nice. Perhaps as a notebook for pymc-examples?

pymc_experimental/marginal_model.py
ricardoV94: I added some comments explaining what's going on there. Also docstrings to the class

ricardoV94:
Need an approving review :)

zaxtax (approved):
Looks good to me.

twiecki (approved):

PR #92, opened 2022-11-15
Make xhistogram dependency optional

### LINKED ISSUE #82: Make xhistogram optional
Given the wide range of utilities we want to add to pymc-experimental it would be great if any dependencies that are not bundled with PyMC were optional. If I don't intend to use the histogram functionality, I would rather not have to install xhistogram.

CC @ferrine

ricardoV94:
Closes #82

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,5 +81,9 @@ if __name__ == "__main__":
         classifiers=classifiers,
         python_requires=">=3.8",
         install_requires=install_reqs,
-        extras_requires=dict(dask=["dask[all]"]),
+        extras_requires=dict(
+            dask=["dask[all]"], 
+            histogram=["xhistogram"], 
+            complete=["dask[all]", "xhistogram"]
+        ),
     )
--- a/setup.py
+++ b/setup.py
@@ -12,10 +12,11 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import itertools
 import re
 from codecs import open
 from os.path import dirname, join, realpath
-import itertools
+
 from setuptools import find_packages, setup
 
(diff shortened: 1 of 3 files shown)


ferrine:
let's add to a setup.py an extention

```python
extras=dict(histogram=[...], complete=[...])
```

So one can install a specific set of dependencies

ricardoV94:
> let's add to a setup.py an extention
> 
> ```python
> extras=dict(histogram=[...], complete=[...])
> ```
> 
> So one can install a specific set of dependencies

Do you want to push that into this PR? I have no idea how that's done but sounds quite good

setup.py
```diff
@@ -81,5 +81,9 @@ def get_version():
         classifiers=classifiers,
         python_requires=">=3.8",
         install_requires=install_reqs,
-        extras_requires=dict(dask=["dask[all]"]),
+        extras_requires=dict(
+            dask=["dask[all]"], 
+            histogram=["xhistogram"], 
+            complete=["dask[all]", "xhistogram"]
```
ricardoV94, line 87: Should it be? Or does `complete` have a special meaning

```suggestion
            all=["dask[all]", "xhistogram"]
```

ferrine:
looks good to me

ferrine (approved):

ricardoV94:
Thanks @ferrine!

ferrine:
I noticed dask requirement on its own does not make sense so I've incremented the dependency list

ferrine:
+ there is warning
```
WARNING: dask 2022.11.0 does not provide the extra 'all'
```

ferrine (approved):

PR #93, opened 2022-11-18
add default to argument to prior util

ferrine:


ricardoV94:
What's the point of calling this function with an empty model?

ferrine:
> What's the point of calling this function with an empty model?

if you pass everything to kwargs for example

ricardoV94:
Will this function ever do anything with empty var_names?

ferrine:
> Will this function ever do anything with empty var_names?

kwargs + var_names combined matter, not just var_names

ferrine:
> > > Will this function ever do anything with empty var_names?
> > 
> > 
> > kwargs + var_names combined matter, not just var_names
> 
> So why this change to allow kwargs + no var_names? If that case can only end up with an empty dictionary?

he change is required to use the function with kwargs only

ricardoV94 (approved):

PR #100, opened 2022-12-16
add DiscreteMarkovChainRV

jessegrabowski:
Add a distribution for discrete-state Markov chains. Support for deterministic and random initial conditions.  Motivated by two recent threads on the discourse asking about this type of model: [here](https://discourse.pymc.io/t/help-understanding-pm-categorical/10960) and [here](https://discourse.pymc.io/t/simple-markov-chain-with-aesara-scan/10745). This work is derivative of the work presented by @junpenglao and @ricardoV94 in those threads.

Still a lot of work to do so I'm marking this as a draft. In no particular order:

1. I use `.eval()` methods in several places to validate inputs, this strikes me as bad but I didn't know what else to do.
2. Using `dims` currently breaks the model
3. The time series dimension is currently specified by the `steps` argument, but it seems more natural if it were instead set by the `size` or `dims` argument (this is related to why dims breaks the model i think)
4. `pm.sample` automatically assigns the markov chain RV to `pm.Metropolis`, don't know how to direct it to BinaryMetropolis or CategoricalMetropolis, depending on the size of the state space.
5. There's some janky stuff with the `steps` argument. I internally subtract 1 to account for the fact that `x0` will be appended to the scan -- otherwise the resulting chain is length steps + 1, which I don't think will match the user expectation. It gives the right answer but it doesn't feel very clean.


pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+from pytensor.tensor.random.op import RandomVariable
+from pytensor.graph.basic import Node
+
+from pymc.distributions.shape_utils import (
+    _change_dist_size,
+    get_support_shape_1d,
+)
+
+
+def validate_transition_matrix(P):
```
ricardoV94, line 25: Instead of eagerly evaluating P,  you can wrap it in a PyTensor assert

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+    if not row_sums_all_one.eval():
+        raise ValueError('All rows of P must sum to 1.')
+
+
+class DiscreteMarkovChainRV(SymbolicRandomVariable):
+    default_output = 1
+    _print_name = ('DiscreteMC', '\\operatorname{DiscreteMC}')
+
+    def update(self, node: Node):
+        # TODO: Do I need this?
```
ricardoV94, line 45: Yes, otherwise the categorical in the scan won't be properly seeded

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+        Length of the markov chain
+    x0: tensor or RandomVariable
+        Intial state of the system. If tensor, treated as deterministic.
+    """
+
+    rv_type = DiscreteMarkovChainRV
+
+    def __new__(cls, *args, steps, **kwargs):
+        # TODO: Allow steps to be None and infer chain length from shape?
+        # TODO: Dims breaks the RV
```
ricardoV94, line 74: Why does it break? Mind adding a test?

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+        P = at.as_tensor_variable(P)
+        validate_transition_matrix(P)
+
+        # TODO: Can this eval be avoided?
+        n_states = P.shape[0].eval()
+
+        if not isinstance(x0, TensorVariable):
+            x0 = at.as_tensor_variable(x0).astype(intX)
+
+            # TODO: Can this eval be avoided?
```
ricardoV94, line 113: If the code wont just break loudly, you can wrap it in an Assert as well.

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+                    'x0 has support over a range of values larger than the number of states in the Markov Chain')
+
+        return super().dist([P, logit_P, steps, x0], **kwargs)
+
+    @classmethod
+    def rv_op(cls, P, logit_P, steps, x0, size=None):
+        if size is not None:
+            batch_size = size
+        else:
+            batch_size = at.broadcast_shape(x0)
```
ricardoV94, line 135: This broadcast doesn't do anything

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+@_logprob.register(DiscreteMarkovChainRV)
+def discrete_mc_logp(
+        op, values, P, x0, steps, state_rng, **kwargs
+):
+
+    (value,) = values
+    # GARCH11 swaps the time axis to the front, not sure why this is necessary
+    mc_logprob = at.log(P[value[..., :-1], value[..., 1:]]).sum(axis=-1)
```
ricardoV94, line 189: I guess scan logp was too painful :P?

ricardoV94 (commented):
This looks amazing 🤩 Left some comments below that hopefully you find helpful.

ricardoV94:
For the sampler, I think there's a way to make it so your distribution goes with Categorical sampler by default.

pymc_experimental/distributions/timeseries.py
jessegrabowski: Yes, I copied it from GARCH11, but then I didn't know what to do with it. I guess it's supposed to ensure that if i pass a vector to the parameters, the shape of those parameters will be added as a batch_dim? I'll figure out how to make sure this matches the other time series distributions in handling that case.

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+
+    return mc_logprob
+
+@_moment.register(DiscreteMarkovChainRV)
+def discrete_markov_chain_moment(
+        op, rv, P, x0, steps, state_rng
+):
+    # TODO: What is the mean of the chain?
+    return at.zeros_like(rv)
```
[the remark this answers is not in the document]
jessegrabowski: Is that the only place the moment is used?

The only moment-like summary statistic I know for these markov chains is the stationary distribution. Would computing that be useful anywhere? I don't think so but I also have no idea.

pymc_experimental/distributions/timeseries.py
jessegrabowski: Always T__T

pymc_experimental/distributions/timeseries.py
jessegrabowski: Yes i need to add a test suite in general but I wasn't sure how. I can just copy tests used in the other time series distributions?

ricardoV94: Sure

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
ricardoV94: Might be useful for users but wouldn't be used internally anywhere no. It doesn't even matter for observed variables.

So definitely not a blocker

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+        if logit_P is not None:
+            P = pm.math.softmax(logit_P, axis=1)
+        P = at.as_tensor_variable(P)
+        validate_transition_matrix(P)
+
+        # TODO: Can this eval be avoided?
+        n_states = P.shape[0].eval()
+
+        if not isinstance(x0, TensorVariable):
+            x0 = at.as_tensor_variable(x0).astype(intX)
```
ricardoV94, line 111: If x0 is accounted for in the logp, you should treat it like init_dist in other timeseries: check it's not registered in the model, resize it to match the size (batched dims) and call `ignore_logprob` so PyMC doesn't complain when it sees a "loose" RV without a value.

jessegrabowski: x0 isn't included in the logp since you can't get the probability of x0 out of the transition matrix (no associated previous value). The first term of the logp is `log(P[state_0, state_1])`, which is P(state_1 | state_0).

I had hoped with this setup everything would magically work. It doesn't complain about loose RVs, and the x0 probability should be automatically handled by the prior on x0. If you check the notebook I did have an issue where the individual logp terms didn't add up as expected, though -- not sure if this is related.

ricardoV94: Okay. You can consider including it, so that users don't have to worry about shaping issues. Then they would call `DiscreteMarkov(init_dist=Categorical.dist(), P=..., shape=...)` And because you have control over the size of init_dist, some of the checks can probably go away.

ricardoV94: On the other hand it makes it difficult to have constant x0, if that's desired.

jessegrabowski: I split them on the basis of "clean coding": since the x0 distribution isn't, strictly speaking, part of the markov chain, there's no reason for the markov chain class to be creating it. 

What's more important though is that this matches the API of other distributions that ask for an initial condition. I would change this to match the way you and the other core devs consider too be the most "PyMC-like".

ricardoV94: The integrated init_dist is more in line with our timeseries. I don't particularly love the API, but I guess it's better to stick with it until someone thinks of a comprehensive alternative.

jessegrabowski: I'll make the change. IMO the comprehensive alternative is to explicitly demand the user make their own initial distributions, since all of these distributions are written mathematically with "x0 given" .

pymc_experimental/distributions/timeseries.py
jessegrabowski: I followed the example set other distributions and wrapped the `logp` return with `check_parameters`, which no longer requires eager evaluation.

pymc_experimental/distributions/timeseries.py
jessegrabowski: Thanks!

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+        if P is not None and logit_P is not None:
+            raise ValueError('Must specify only one of either P or logit_P parameter')
+
+        if logit_P is not None:
+            P = pm.math.softmax(logit_P, axis=1)
+        P = at.as_tensor_variable(P)
+        validate_transition_matrix(P)
+
+        # TODO: Can this eval be avoided?
+        n_states = P.shape[0].eval()
```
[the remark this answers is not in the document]
jessegrabowski: Resolved by moving the shape checks to the `logp` function

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: Is it worth returning zeros just to silence the warning? All zeros is a valid markov chain under any parameterization.

pymc_experimental/distributions/timeseries.py
jessegrabowski: I added a check for a dtype kwarg in the `dist` function and set it to `pytensor.config.floatX.replace('float', 'int')` if it's None. I was getting an error because the scan was returning `int32` variables, while the `init_dist` was returning int64. Is there a better way to handle this?

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,198 @@
...
+
+        else:
+            x0_probs = x0.owner.inputs[-1].eval()
+            n_cats = 1 if x0_probs.ndim == 0 else len(x0_probs)
+
+            if not n_cats <= n_states:
+                raise ValueError(
+                    'x0 has support over a range of values larger than the number of states in the Markov Chain')
+
+        return super().dist([P, logit_P, steps, x0], **kwargs)
```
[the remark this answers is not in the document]
jessegrabowski: I set the internal representation to use `P` only.

I think `dims` are breaking for another reason, I am checking how `steps` are handled in the `AR` distribution for guidance.

pymc_experimental/distributions/timeseries.py
jessegrabowski: x0 parameter removed, added `init_dist` parameter that takes an unnamed distribution as an argument. Docstring and type checks were copied from `pymc.distributions.timeseries.AR`

pymc_experimental/distributions/timeseries.py
jessegrabowski: The break was a result of the `batch_dims` stuff in `rv_op` that did nothing. I refactored this section following the `AR` distribution and any combination of `steps` plus `size` or `dims` now works as expected.

pymc_experimental/distributions/timeseries.py
jessegrabowski: This was the source of the issue with `dims`, it has been resolved, see my comment above.

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
ricardoV94: What about (Pt+1=0|Pt=0)=0?

Maybe you can just return rv? Maybe just set default `initval` in `__new__` to be `prior`? That should do the same without the warning

pymc_experimental/tests/distributions/test_discrete_markov_chain.py
```diff
@@ -0,0 +1,91 @@
...
+    P = at.zeros((3, 3))
+    chain = DiscreteMarkovChain.dist(P=P, steps=3)
+    with pytest.raises(ParameterValueError):
+        pm.logp(chain, np.zeros((3,))).eval()
+
+
+def test_logp_with_default_init_dist():
+    P = at.full((3, 3), 1 / 3)
+
+    chain = DiscreteMarkovChain.dist(P=P, steps=3)
```
ricardoV94, line 29: `with pytest.warns(...):`?

pymc_experimental/tests/distributions/test_discrete_markov_chain.py
```diff
@@ -0,0 +1,91 @@
...
+
+# general imports
+import pytensor.tensor as at
+import pytest
+from pymc.logprob.utils import ParameterValueError
+
+from pymc_experimental.distributions.timeseries import DiscreteMarkovChain
+
+
+def test_fail_if_P_not_square():
```
ricardoV94, line 12: Can you group these tests in a class?

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+                    f"got {type(init_dist)}"
+                )
+            check_dist_not_registered(init_dist)
+            if init_dist.owner.op.ndim_supp > 1:
+                raise ValueError(
+                    "Init distribution must have a scalar or vector support dimension, ",
+                    f"got ndim_supp={init_dist.owner.op.ndim_supp}.",
+                )
+        else:
+            warnings.warn(
```
ricardoV94, line 118: This initial distribution seems reasonable enough that we might not need the warning?

pymc_experimental/tests/distributions/test_discrete_markov_chain.py
```diff
@@ -0,0 +1,91 @@
+import numpy as np
+import pymc as pm
+
+# general imports
```
ricardoV94, line 4: Can you add a test for the random and moment functions?

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+        Matrix of tranisiton logits. Converted to probabilities via Softmax activation.
+        One of P or P_logits must be provided.
+    steps: tensor, optional
+        Length of the markov chain. Only needed if state is not provided.
+    init_dist : unnamed distribution, optional
+        Vector distribution for initial values. Unnamed refers to distributions
+        created with the ``.dist()`` API. Distribution should have shape n_states.
+        If not, it will be automatically resized. Defaults to pm.Categorical.dist(p=np.full(n_states, 1/n_states)).
+        .. warning:: init_dist will be cloned, rendering it independent of the one passed as input.
+    """
```
ricardoV94, line 56: I suggest you add an example code snippet

ricardoV94 (commented):
Great progress. Some comments/requests below.

Would be so cool to implement logp marginalization of these variables in a future PR! Should be as simple as putting the right DiscreteMarginalizedRV inside Scan? Should come out the same as the forward algorithm.

Another future improvement would be multiple let dependencies? Anyway I am getting carried away, this is already very exciting!

pymc_experimental/distributions/timeseries.py
ricardoV94: I think you can set it to int64 as all our discrete distributions are of that type?

pymc_experimental/distributions/timeseries.py
ricardoV94: > Yes, I copied it from GARCH11, but then I didn't know what to do with it. I guess it's supposed to ensure that if i pass a vector to the parameters, the shape of those parameters will be added as a batch_dim? 

When you don't specify size explicit, it's implied by the parameters shape. If there's more than one parameter, it's determined by the broadcasted batched dimensions. When you have a single parameter there's no need to broadcast.

pymc_experimental/distributions/timeseries.py
ricardoV94: By comprehensive I mostly meant user friendly :)

pymc_experimental/distributions/timeseries.py
ricardoV94: And an entry to the docs

ricardoV94:
> For the sampler, I think there's a way to make it so your distribution goes with Categorical sampler by default.

Taking of bad APIs...

Here is a solution? Like the test added here you can override the default list of step methods: https://github.com/pymc-devs/pymc/commit/2dd4c8c5401121b3e48fffd558838fe2d547a4c2

You could add a subclass of the categorical metropolis sampler that assigns ideal competence to this variable.

We should definitely refactor that approach, perhaps with dispatching, but for now that is the step assignment API for external libraries

jessegrabowski:
> > For the sampler, I think there's a way to make it so your distribution goes with Categorical sampler by default.
> 
> Speaking of bad APIs...
> 
> Here is a solution? Like the test added here you can override the default list of step methods: [pymc-devs/pymc@2dd4c8c](https://github.com/pymc-devs/pymc/commit/2dd4c8c5401121b3e48fffd558838fe2d547a4c2)
> 
> You could add a subclass of the categorical metropolis sampler that assigns ideal competence to this variable.
> 
> We should definitely refactor that approach, perhaps with dispatching, but for now that is the step assignment API for external libraries...

I was thinking if I could subclass from Categorical then it would be automatically assigned correctly, but I got some errors doing this.

pymc_experimental/distributions/timeseries.py
jessegrabowski: I guess to me the alternative is that the init_dist parameters want actual random variables, not `.dist`. They can still be concatenated to the time series output internally -- your point Pro (1) is very crucial I think and any implementation should have this property. I would argue that Con (1) applies more to my proposal, because you have to do a bunch of type and shape checking on the distribution being provided. But these checks could be farmed out to a function common to all time series distributions, so the individual distributions wouldn't need to know anything about it.

To me the biggest advantage is allowing the graph to naturally handle the extra x0 logp, rather than having to manually ignore it in the `.dist` method and then re-compute it in the `.logp` method.

jessegrabowski:
re: multiple lag dependencies, I think that would be nice as well and should be easy enough to add. 
re: Marginalization, I was really hoping this would integrate with that PR as well. The HMM model in particular is quite difficult to sample (see the example notebook as well as discussion [here](https://discourse.pymc.io/t/how-to-marginalized-hidden-markov-model-with-categorical/2230)), and I wonder if automatic marginalization could help the sampler along.

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: >What about (Pt+1=0|Pt=0)=0?

In principal we should probably dis-allow absorbing states. I have no proof, but I suspect they would cause numerical headaches.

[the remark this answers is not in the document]
ricardoV94: This is not an absorbing state, it's a forced switch state. Eg `P=[[0, 0.5, 0.5], [1/3, 1/3, 1/3], [1/3, 1/3, 1/3]]`

ricardoV94:
Marginalizing HMMs usually sample much better in my experience. We would need to increase the marginalization to support it, though. Right now it only works for vanilla discrete variables, not scans or this specific wrapped scan op either.

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: Yes you are absolutely correct! My mistake.

pymc_experimental/distributions/timeseries.py
jessegrabowski: I defer to your judgement. I tend to err on the side of "give too much information" in my code -- reasonable or not, it's something we're injecting into the user's model.

ricardoV94: No strong opinion, go with your preference

pymc_experimental/distributions/timeseries.py
jessegrabowski: Yeah i'm not sure what I thought the problem is. Removed all this and the code runs fine.

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: Removed the moment method, added `initval='prior'` as a default argument to `__new__`.

jessegrabowski:
One of the scan re-writes is throw an error now, but only if I compile the model outside of a model context (e.g. with `pm.draw`). Curious if you know what might be causing it?

```python
ERROR (pytensor.graph.rewriting.basic): Rewrite failure due to: save_mem_new_scan
ERROR (pytensor.graph.rewriting.basic): node: for{cpu,scan_fn}(TensorConstant{(1,) of 9}, IncSubtensor{Set;:int64:}.0, RandomGeneratorSharedVariable(<Generator(PCG64) at 0x151E8EE40>), Softmax{axis=1}.0)
ERROR (pytensor.graph.rewriting.basic): TRACEBACK:
ERROR (pytensor.graph.rewriting.basic): Traceback (most recent call last):
  File "/Users/jessegrabowski/mambaforge/envs/pymc_dev/lib/python3.9/site-packages/pytensor/graph/rewriting/basic.py", line 1933, in process_node
    replacements = node_rewriter.transform(fgraph, node)
  File "/Users/jessegrabowski/mambaforge/envs/pymc_dev/lib/python3.9/site-packages/pytensor/graph/rewriting/basic.py", line 1092, in transform
    return self.fn(fgraph, node)
  File "/Users/jessegrabowski/mambaforge/envs/pymc_dev/lib/python3.9/site-packages/pytensor/scan/rewriting.py", line 1628, in save_mem_new_scan
    subtens = Subtensor(nw_slice)
  File "/Users/jessegrabowski/mambaforge/envs/pymc_dev/lib/python3.9/site-packages/pytensor/tensor/subtensor.py", line 692, in __init__
    self.idx_list = tuple(map(index_vars_to_types, idx_list))
  File "/Users/jessegrabowski/mambaforge/envs/pymc_dev/lib/python3.9/site-packages/pytensor/tensor/subtensor.py", line 592, in index_vars_to_types
    slice_a = index_vars_to_types(a, False)
  File "/Users/jessegrabowski/mambaforge/envs/pymc_dev/lib/python3.9/site-packages/pytensor/tensor/subtensor.py", line 613, in index_vars_to_types
    raise AdvancedIndexingError("Invalid index type or slice for Subtensor")
pytensor.tensor.exceptions.AdvancedIndexingError: Invalid index type or slice for Subtensor
```

pymc_experimental/distributions/timeseries.py
jessegrabowski: Verbose for now then, to match AR. Open to changing it in the future.

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+        return super().dist([P, steps, init_dist], **kwargs)
+
+    @classmethod
+    def rv_op(cls, P, steps, init_dist, size=None):
+        if size is not None:
+            batch_size = size
+        else:
+            batch_size = (1,)
+
+        if init_dist.owner.op.ndim_supp == 0:
```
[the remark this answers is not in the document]
jessegrabowski: you mean the entire `init_dist.ower.op.ndim_supp` check can be removed?

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+        .. warning:: init_dist will be cloned, rendering it independent of the one passed as input.
+    """
+
+    rv_type = DiscreteMarkovChainRV
+
+    def __new__(cls, *args, steps=None, **kwargs):
+        # Subtract 1 step to account for x0 given, better match user expectation of
+        # len(markov_chain) = steps
+        if steps is not None:
+            steps -= 1
```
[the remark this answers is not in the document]
jessegrabowski: `support_shape_offset` will handle the cases where the time dimension is defined via size, but not if it is defined via steps. For example, `pm.AR.dist(rho=0.99, steps=10).eval().shape` gives `(11,)`, but  `pm.AR.dist(rho=0.99, shape=10).eval().shape` gives `(10,)`. This correction to steps unifies the output shape to `(10,)` in all cases.

pymc_experimental/tests/distributions/test_discrete_markov_chain.py
jessegrabowski: The moment function was removed, but tests were added for random draws and change_dist_size.

pymc_experimental/distributions/timeseries.py
jessegrabowski: Where would I update the docs? I don't see anything in pymc-experimental/docs

jessegrabowski:
I marked it as ready for review, although the problem of sampler assignment still isn't solved. If it was OK'd for the main PyMC code base, I'd just add the distribution to the list of competencies for `CategoricalGibbsMetropolis` I guess.

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
ricardoV94: Why not make steps behave like other timeseries? A GRW with shape 10, has 9 steps. Most users will never see steps, since we advertise the shape/dims/observed API 

It is indeed 9 steps right?

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
ricardoV94: Yes

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+            init_dist = pm.Categorical.dist(p=pt.full((k,), 1 / k), shape=batch_size)
+
+        # We can ignore init_dist, as it will be accounted for in the logp term
+        init_dist = ignore_logprob(init_dist)
+
+        return super().dist([P, steps, init_dist], **kwargs)
+
+    @classmethod
+    def rv_op(cls, P, steps, init_dist, size=None):
+        batch_size = size or (1,)
```
ricardoV94, line 145: The RandomVariable API, would say something like `batch_size = size if size is not None else at.broadcast_shapes(P.shape[:-2], init_dist.shape[:-1])[0]`

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+        init_dist = change_dist_size(init_dist, init_dist_size)
+
+        init_dist_ = init_dist.type()
+        P_ = P.type()
+        steps_ = steps.type()
+
+        state_rng = pytensor.shared(np.random.default_rng())
+
+        def transition(previous_state, transition_probs, old_rng):
+            p = transition_probs[previous_state]
```
ricardoV94, line 161:
```suggestion
            p = transition_probs[..., previous_state, :]
```

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+            non_sequences=[P_, state_rng],
+            outputs_info=[init_dist_],
+            n_steps=steps_,
+        )
+
+        (state_next_rng,) = tuple(state_updates.values())
+
+        discrete_mc_ = pt.moveaxis(
+            pt.concatenate([init_dist_[None, ...], markov_chain], axis=0), 0, -1
+        ).squeeze()
```
ricardoV94, line 176: Why the squeeze?

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: I have no idea what a "step" is I guess. It has 9 "state transitions", are those steps? I guess personally I interpreted "steps" to mean "time steps', which would be the length of the chain

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+                    f"got ndim_supp={init_dist.owner.op.ndim_supp}.",
+                )
+        else:
+            warnings.warn(
+                "Initial distribution not specified, defaulting to "
+                "`Categorical.dist(p=pt.full((k_states, ), 1/k_states), shape=...)`. You can specify an init_dist "
+                "manually to suppress this warning.",
+                UserWarning,
+            )
+            k = P.shape[0]
```
ricardoV94, line 135:
```suggestion
            k = P.shape[..., 0, :]
```

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: Anyway it's easy enough to remove this and just keep everything consistent.

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+            init_dist = pm.Categorical.dist(p=pt.full((k,), 1 / k), shape=batch_size)
```
ricardoV94, line 136: You don't need to worry about batch shapes, just the core shape. You will resize it anyway in `rv_op`

```suggestion
            init_dist = pm.Categorical.dist(p=pt.full((k,), 1 / k))
```

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+
+    return DiscreteMarkovChain.rv_op(
+        *dist.owner.inputs[:-1],
+        size=new_size,
+    )
+
+
+@_logprob.register(DiscreteMarkovChainRV)
+def discrete_mc_logp(op, values, P, steps, init_dist, state_rng, **kwargs):
+    n, k = P.shape
```
ricardoV94, line 202:
```suggestion
    n, k = P.shape[-2:]
```

ricardoV94:
> Also still need to add support for multiple lags.

That need not be a blocker unless you want it to

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
ricardoV94: We need steps internally because we use size for RVs. Size does not contain info about support shape, which for univariate timeseries is the last one.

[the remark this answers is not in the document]
ricardoV94: Steps is the number of steps in the scan argument

pymc_experimental/distributions/timeseries.py
ricardoV94: Here I think?

https://github.com/pymc-devs/pymc-experimental/blob/main/docs/api_reference.rst

ricardoV94: CC @OriolAbril

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: Of course. But then it gets x0 stuck onto it, which is the source of my confusion -- i wanted to adjust for that.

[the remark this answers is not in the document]
ricardoV94: The final distribution is now x0 plus the rest? I guess this made sense when x0 was not part of the distribution?

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,214 @@
...
+    --------
+    .. code-block:: python
+        # Create a Markov Chain of length 100 with 3 states
+        with pm.Model() as markov_chain:
+            # The transition probability matrix should be square with rows that sum to 1
+            # The number of states in the markov chain is given by the shape of P, 3 in this example
+            P = pm.Dirichlet("P", a=[1, 1, 1], size=(3,))
+            # The initial state probabilities should have size = n_states, or 3 in this case.
+            init = pm.Categorical.dist(p = np.full(3, 1 / 3))
+            markov_chain = pm.DiscreteMarkovChain("markov_chain", P=P, init_dist=init, steps=100)
```
ricardoV94, line 69:
```suggestion
            markov_chain = pm.DiscreteMarkovChain("markov_chain", P=P, init_dist=init, shape=(100,))
```

pymc_experimental/distributions/timeseries.py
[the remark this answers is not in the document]
jessegrabowski: Right, if x0 counts and I ask for 10 steps, i get back 11 -- 10 from the scan and 1 from x0. I thought if x0 is part of the distribution, it should scan for 1 less to account for the fact we already have the first "step".

jessegrabowski:
> That need not be a blocker unless you want it to

It shouldn't be too hard to do, i just need to figure out how to fancy index from a symbolic vector. 

To get the transition probabilities for an arbitrary number of lags in numpy, I'd write:

```python
P[tuple([output[i, t-j] for j in range(n_lags, 0, -1)])]
```

But I don't know how to implement this pytensor (short of adding another scan...), since `output` is a symbolic vector.

pymc_experimental/distributions/timeseries.py
jessegrabowski: It comes out as `(1, steps)` otherwise

pymc_experimental/distributions/timeseries.py
jessegrabowski: In the ndim > 2 case every dimension needs to be the same, so it should be changed to something like `pt.all(pt.eq(P.shape, P.shape[0])`

pymc_experimental/distributions/timeseries.py
jessegrabowski: This broke my tests, the logic is to allow potential batch dims on P, e.g. in a hierarchical case? Could we just take `P[-1]` instead of `P[0]`?

pymc_experimental/distributions/timeseries.py
OriolAbril: the current api docs look a bit strange (and limited?). The main question is "how should users import this distribution?", then I'll be able to indicate how to document it.

jessegrabowski: Strange how? I tried to follow the style in the other timeseries.py distributions -- eager to fix it if it's off the mark.

pymc_experimental/distributions/timeseries.py
jessegrabowski: Actually if there's the potential for batch dimensions on `P` the whole logp function needs some tweaking, not just the shape check.

pymc_experimental/distributions/timeseries.py
OriolAbril: > Strange how?

There are 3 modules _documented_, and each has only one object inside of it. Yet there are 2 section headings seemingly indicating the existence of 2 modules only which are a superset of the documented modules.

My hunch is that what should be documented are two modules (trusting the headings instead of the `automodule` directives), `distributions` and `utils` with elements being imported from there, not from their specific file (e.g. `pmx.utils.prior_from_idata` instead of `pmx.utils.prior.prior_from_idata`). It also seems clear that there are several objects missing from the API reference.

pymc_experimental/distributions/timeseries.py
jessegrabowski: This gives me a cannot broadcast shapes message. Should I be adding a batch dim to P and init_dist?

pymc_experimental/distributions/timeseries.py
ricardoV94: There's an open issue to fix it: https://github.com/pymc-devs/pymc-experimental/issues/101

Should we not even try to include your distribution in the docs in this PR?

pymc_experimental/distributions/timeseries.py
ricardoV94: No I don't think so, what shape is it trying to broadcast? Maybe the [0] index is wrong? Maybe the function returns a single shape ad output? I was assuming it would return one for each input but that's arguably silly since they should be the same

pymc_experimental/distributions/timeseries.py
ricardoV94: Yes that's the logic, If `P.shape=[2, 3, 3]` I was trying to get 3. What you say is the correct thing. I thought we were indexing on P directly not on it's shape

pymc_experimental/distributions/timeseries.py
ricardoV94: Not all dims need to be the same. P.shape=[2, 3, 3] would mean 2 independent chains, each with 3 states. Only the last two shapes need to match (be square)

ricardoV94:
> > That need not be a blocker unless you want it to
> 
> It shouldn't be too hard to do, i just need to figure out how to fancy index from a symbolic vector.
> 
> To get the transition probabilities for an arbitrary number of lags in numpy, I'd write:
> 
> ```python
> P[tuple([output[i, t-j] for j in range(n_lags, 0, -1)])]
> ```
> 
> But I don't know how to implement this pytensor (short of adding another scan...), since `output` is a symbolic vector.

PyTensor can do advanced indexing just fine. Should work just like Numpy. If you have 2 lags, The P matrix should be n x n x n? The transition in the scan step should have probability `P[state_m2, state_m1]`? Your init dist must have shape (..., 2) and you can use Scan taps functionality so the inner function gets all the lags you need

pymc_experimental/distributions/timeseries.py
ricardoV94: When I looked at the logp, I thought it would work out of the box.

ricardoV94:
> I marked it as ready for review, although the problem of sampler assignment still isn't solved. If it was OK'd for the main PyMC code base, I'd just add the distribution to the list of competencies for CategoricalGibbsMetropolis I guess.

Actually I think we really need a subclass because the sampler wouldn't know how to infer the number of possible states (it assumes it's assigned a categorical variable, where P is the fourth input). We need to override the init

pymc_experimental/distributions/timeseries.py
jessegrabowski: As currently implemented, the scan isn't written to handle independent chains. I just assumed P would always be a square tensor. I will experiment to see what changes are required, if any.

jessegrabowski:
Understood re: sampler. I'm surprised though, because just manually assigning the variable to `pm.CategoricalGibbsMetropolis` seemed to work fine in my example (it struggles to sample but I assumed this was just a feature of an HMM). How do discrete RVs "communicate" their support to the sampler? Is that something I can provide somewhere, e.g. in `__new__`? It seems odd that we would need to subclass an entire step function just for one RV.

For the advanced indexing, I was making it more complex than it needed to be. Thanks for your guidance!

ricardoV94:
> How do discrete RVs "communicate" their support to the sampler? Is that something I can provide somewhere, e.g. in `__new__`? It seems odd that we would need to subclass an entire step function just for one RV.

They don't communicate. The Binary and Categorical assume you assign only a Binary or Categorical, and the latter infers the support similarly to how you did it from the P matrix.

Otherwise Metropolis sampler knows nothing about the variables and it's happy to propose values outside of the domain (tangential issue here: https://github.com/pymc-devs/pymc/pull/6102)

pymc_experimental/distributions/timeseries.py
jessegrabowski: What I came up with in the end was pretty ugly:

```python

            batch_size = pt.broadcast_shape(
                P[tuple([...] + [0] * (n_lags + 1))], pt.atleast_1d(init_dist)[..., 0]
            )
```

This will slice off all the square parts of P (which has an assumed structure of `(batch_dims) + (n_states) * (n_lags +1)` and match just the batch_dims of P with those of the user-provided `init_dist`. Not sure if you have a more elegant suggestion?

jessegrabowski:
I updated `rv_op` to allow for more lags, but it's still not finished because I'm figuring out how to pass around the `n_lags` information between `dist` and `rv_op`. It needs to remain a python integer coming into `rv_op`, because I can't pass a symbolic tensor to taps. On the other hand, it needs to be a tensor object, because otherwise I can't pass it to the `OpFromGraph` function that will define the information set of downstream functions (logp, change_shape)

ricardoV94:
You can make it a property of the Op. Like `order` in the AR distribution

pymc_experimental/distributions/timeseries.py
OriolAbril: We can already add it in the distributions section as pymc_experimental.distributions.timeseries.DiscreteMarkovChain

pymc_experimental/distributions/timeseries.py
```diff
@@ -0,0 +1,215 @@
...
+    ----------
+    P: tensor
+        Matrix of transition probabilities between states. Rows must sum to 1.
+        One of P or P_logits must be provided.
+    P_logit: tensor, optional
+        Matrix of transition logits. Converted to probabilities via Softmax activation.
+        One of P or P_logits must be provided.
+    steps: tensor, optional
+        Length of the markov chain. Only needed if state is not provided.
+    init_dist : unnamed distribution, optional
```
junpenglao, line 53: IIUC, this is a DiscreteMarkovChain in the sense that the emission distribution is a Categorical - if so you might want to restrict the init_dist to  pm.Categorical.dist? Or even better only allow kwarg of `p: Tensor = np.full(n_states, 1/n_states)`

junpenglao:
Are you planning to add the Viterbi algorithm for inference?

pymc_experimental/distributions/timeseries.py
jessegrabowski: Categorical makes the most sense but in principle it could be any discrete distribution no?

jessegrabowski:
I wasn't planning to add Viterbi but I have no objection to doing it. It seems out of scope for the distribution itself (since it operations on the markov chain but isn't an intrinsic property of it), but otoh I'm not sure where in the code base it would fit.

It would make the most sense if we had an HMM class wrapper that people could easily use, especially if we could get the forward algorithm by combining in automatic marginalization of the hidden states.  IMO that would be the natural place for things like Viterbi (and beam search for that matter)

jessegrabowski:
I fixed the shape problems, so I think all the core functionality is done now

pymc_experimental/distributions/timeseries.py
junpenglao: Depending on how you are going to deal with unbounded (Poisson) discrete distribution or component with different support - I dont quite see it from the docstring and your implementation, but that's something we can think about it more in follow up.

jessegrabowski: Yes, as it stands it would allow you to use the unbounded init_dist, then fail on sampling when a state outside of the support came along. Is there a way to quickly check the support of a discrete distribution? In an earlier version I did a check like:

```python
x0_probs = init_dist.owner.inputs[-1].eval()
n_cats = 1 if x0_probs.ndim == 0 else len(x0_probs)
```

Which checked the number of categories in the categorical. The if/else logic was there to check for a Bernoulli, but it wouldn't catch a Poisson. 

Anyway, I think you suggestion is probably right, but my last doubt is that all the other distributions in `distributions.timeseries` use `init_dist` in their signature and expect an unnamed distribution, and I think there is value in consistency. I could add a check (and update the docstring) to enforce that the distribution passed must be categorical (for now at least).

ricardoV94:
@jessegrabowski wanna try to push this through the finish line?

pymc_experimental/distributions/timeseries.py
jessegrabowski: I tried to split the baby by adding a check that `init_dist` must be `pm.Categorical`, and it raises an error if not. I really wanted the API to be consistent with the other time series distributions. If you think that's not valuable, I'll change it so that `init_dist` becomes `init_p` instead.

pymc_experimental/distributions/timeseries.py
jessegrabowski: What do I need to do to resolve this?

jessegrabowski:
I think I addressed pretty much everything. I changed an import in `marginal_model.py` because I did `git rebase` on my fork's `main` but not the branch i set up for this PR; I guess that was pretty stupid. I needed to change the imports for `logp`. I hope it's not too much of a hassle to fix it.

pymc_experimental/distributions/timeseries.py
ricardoV94: Just list it here? https://github.com/pymc-devs/pymc-experimental/blob/main/docs/api_reference.rst#distributions

twiecki:
The NB should go to pymc-examples.

jessegrabowski:
The NB was a bit of an afterthought, just to show everyone the distribution works as intended. I could doll it up a bit and make a separate PR into `pymc-examples`?

twiecki:
@jessegrabowski I updated my comment after I saw that this was pymc-experimental, not pymc. So I think a NB here is a good idea, and you can later doll it up for pymc-examples. But the title needs to be fixed.

pymc_experimental/distributions/timeseries.py
ricardoV94: I am not sure about the categorical constraint. I might want to pass a diracdelta for posterior predictive or use a discrete uniform because it's simpler. Anyway, doesn't need to be a blocker.

jessegrabowski: I'm glad you flagged it, I was about to push a change to let Bernoulli in, but the more I think about it the more I think it's better to just assume the user will be responsible?

jessegrabowski: I removed the check.

docs/api_reference.rst
```diff
@@ -30,6 +30,7 @@ Distributions
 
    GenExtreme
    histogram_utils.histogram_approximation
+   timeseries.DiscreteMarkovChain
```
ricardoV94, line 33: If you add it to `distributions.__init__.py` `__all__` the short path should work (and it will be easier for users to find)
 
```suggestion
   DiscreteMarkovChain
```

jessegrabowski: Done

ricardoV94 (approved):

PR #104, opened 2023-01-15
Updated load method in ModelBuilder

5hv5hvnk:
@twiecki


pymc_experimental/model_builder.py
```diff
@@ -157,8 +157,7 @@ def save(self, fname):
         file = Path(str(fname))
         self.idata.to_netcdf(file)
 
-    @classmethod
-    def load(cls, fname):
+    def load(self, fname):
```
twiecki, line 160: I think we should keep this a class-method.

pymc_experimental/model_builder.py
```diff
@@ -187,18 +186,14 @@ def load(cls, fname):
...
-        #         )
-        return idata
+        if idata.attrs is not None:
+            if self.id() == idata.attrs["id"]:
+                self.idata = idata
+            else:
+                raise ValueError(
+                    f"The route '{file}' does not contain an inference data of the same model '{self.__name__}'"
+                )
+        # return idata
```
twiecki, line 196: I think we require something like 
```python
self = cls(idata.attrs["sample_config"], idata.attrs["model_config"])
self.idata = idata
return self
```

cls is the class being passed above, so we need to instantiate it using the right args from the idata.

pymc_experimental/model_builder.py
```diff
@@ -238,8 +233,8 @@ def fit(self, data: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]] = None
         self.idata.attrs["id"] = self.id()
         self.idata.attrs["model_type"] = self._model_type
         self.idata.attrs["version"] = self.version
-        self.idata.attrs["sample_conifg"] = self.sample_config
-        self.idata.attrs["model_config"] = self.model_config
+        self.idata.attrs["sample_conifg"] = tuple(self.sample_config)
```
twiecki, line 236:
```suggestion
        self.idata.attrs["sample_config"] = tuple(self.sample_config)
```

pymc_experimental/model_builder.py
5hv5hvnk:
```
self = cls(idata.attrs["sample_config"], idata.attrs["model_config"])
self.idata = idata
```

should be in the if condition right?

pymc_experimental/model_builder.py
```diff
@@ -158,7 +158,7 @@ def save(self, fname):
         self.idata.to_netcdf(file)
 
     @classmethod
-    def load(cls, fname):
+    def load(cls, self, fname):
```
twiecki, line 161:
```suggestion
    def load(cls, fname):
```

pymc_experimental/model_builder.py
```diff
@@ -187,18 +187,15 @@ def load(cls, fname):
...
-        #     if model.idata.attrs['id'] == self.idata.attrs['id']:
-        #         self = model
-        #         self.idata = data
-        #         return self
-        #     else:
-        #         raise ValueError(
-        #             f"The route '{file}' does not contain an inference data of the same model '{self.__name__}'"
-        #         )
-        return idata
+        if idata.attrs is not None:
```
twiecki, line 190: if arviz has that patch let's remove this.

pymc_experimental/model_builder.py
```diff
@@ -187,18 +187,15 @@ def load(cls, fname):
...
+            if self.id() == idata.attrs["id"]:
+                self = cls(idata.attrs["sample_config"], idata.attrs["model_config"])
```
twiecki, line 192: No this needs to be outside. First create the object `self`, put in the idata, then make the check for id.

pymc_experimental/model_builder.py
5hv5hvnk: it has but I added it to make sure one can't load a `.nc` file with no attrs, If it's not required lets just remove it

pymc_experimental/model_builder.py
5hv5hvnk: This statement is to check if the self and the inference data from `.nc` has same id. If we create the self with the loaded data then how will we check if idata from the file and the current model has same id

pymc_experimental/model_builder.py
twiecki: I think we run into problems if we can't save attrs, we need it.

pymc_experimental/model_builder.py
twiecki: Yes, but we first need to create the `self` object, we can only check `self.id()` after.

twiecki: You can check by trying to run this.

twiecki: Does this make sense:
```python
        filepath = Path(str(fname))
        data = az.from_netcdf(filepath)
        idata = data
        self = cls(idata.attrs["sample_config"], idata.attrs["model_config"])
        self.idata = idata
        if self.id() != idata.attrs["id"]:
            raise ValueError(
                f"The route '{file}' does not contain an inference data of the same model '{self.__name__}'"
            )

        return self
```

pymc_experimental/model_builder.py
```diff
@@ -185,20 +185,18 @@ def load(cls, fname):
...
+        idata = az.from_netcdf(filepath)
+        self = cls(
+            dict(zip(idata.attrs["model_config_keys"], idata.attrs["model_config_values"])),
+            dict(zip(idata.attrs["sample_config_keys"], idata.attrs["sample_config_values"])),
+            idata.data,
+        )
+        self.idata = idata
+        if self.id() != idata.attrs["id"]:
+            raise ValueError(
+                f"The route '{fname}' does not contain an inference data of the same model '{self._model_type}'"
```
twiecki, line 197:
```suggestion
                f"The file '{fname}' does not contain an inference data of the same model or configuration as '{self._model_type}'"
```

pymc_experimental/model_builder.py
```diff
@@ -238,8 +236,11 @@ def fit(self, data: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]] = None
         self.idata.attrs["id"] = self.id()
         self.idata.attrs["model_type"] = self._model_type
         self.idata.attrs["version"] = self.version
-        self.idata.attrs["sample_conifg"] = self.sample_config
-        self.idata.attrs["model_config"] = self.model_config
+        self.idata.attrs["sample_config_keys"] = tuple(self.sample_config.keys())
+        self.idata.attrs["sample_config_values"] = tuple(self.sample_config.values())
+        self.idata.attrs["model_config_keys"] = tuple(self.model_config.keys())
+        self.idata.attrs["model_config_values"] = tuple(self.model_config.values())
+        self.idata.add_groups(data=self.data.to_xarray())
```
twiecki, line 243:
```suggestion
        self.idata.add_groups(constant_data=self.data.to_xarray())
```

twiecki (dismissed):

twiecki:
@5hv5hvnk Still need to fix CI.

pymc_experimental/model_builder.py
```diff
@@ -238,8 +236,11 @@ def fit(self, data: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]] = None
         self.idata.attrs["id"] = self.id()
         self.idata.attrs["model_type"] = self._model_type
         self.idata.attrs["version"] = self.version
-        self.idata.attrs["sample_conifg"] = self.sample_config
-        self.idata.attrs["model_config"] = self.model_config
+        self.idata.attrs["sample_config_keys"] = tuple(self.sample_config.keys())
+        self.idata.attrs["sample_config_values"] = tuple(self.sample_config.values())
+        self.idata.attrs["model_config_keys"] = tuple(self.model_config.keys())
+        self.idata.attrs["model_config_values"] = tuple(self.model_config.values())
+        self.idata.add_groups(constant_data=self.data.to_xarray())
```
twiecki, line 243:
```suggestion
        self.idata.add_groups(fit_data=self.data.to_xarray())
```

twiecki (dismissed):

twiecki:
@marcogorelli Can't use tempfiles on windows CI? getting a permission error.

MarcoGorelli:
ah I saw this recently, here's a fix someone submitted in case it's useful https://github.com/pandas-dev/pandas/pull/51244

PR #105, opened 2023-02-03
Update API reference listings

### LINKED ISSUE #101: List MarginalModel in API reference

ricardoV94:
Probably I have syntax issues: CC @OriolAbril 

Not sure what GP stuff should be listed: CC @bwengals 

Closes #101

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -51,4 +51,3 @@ Utils
 
    spline.bspline_interpolation
    prior.prior_from_idata
-
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -4,51 +4,51 @@ API Reference
 This reference provides detailed documentation for all modules, classes, and
 methods in the current release of PyMC experimental.
 
+.. currentmodule:: pymc_experimental
+.. autosummary::
+   :toctree: generated/
 
-.. toctree::
-  :maxdepth: 2
+   marginal_model.MarginalModel
+   model_builder.ModelBuilder
 
+Inference
+=========
 
-:mod:`pymc_experimental`
-=============================
+.. currentmodule:: pymc_experimental.inference
+.. autosummary::
+   :toctree: generated/
 
-.. automodule:: pymc_experimental.marginal_model
-   :members: MarginalModel
-
-.. automodule:: pymc_experimental.model_builder
-   :members: ModelBuilder
-
-
-:mod:`pymc_experimental.inference`
-=============================
-
-.. automodule:: pymc_experimental.inference.fit
-   :members: fit
+   fit
 
 
-:mod:`pymc_experimental.distributions`
-=============================
+Distributions
+=============
 
-.. automodule:: pymc_experimental.distributions.continuous
-   :members: GenExtreme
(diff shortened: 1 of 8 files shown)


docs/api_reference.rst
```diff
@@ -9,13 +9,40 @@ methods in the current release of PyMC experimental.
   :maxdepth: 2
 
 
+:mod:`pymc_experimental`
```
OriolAbril, line 12: `:mod:` should not be used for the headings. Used like this it is a cross-referencing role that links to the pymc_experimental module (that is, to the place where the `.. module:: pymc_experimental` or `.. automodule:: pymc_experimental` directive is, as this doesn't exist for now right now it does nothing and the cross-reference falls back to raw text, so we can use the raw text. Even if we added the directive it would basically be a link pointing basically to the same place.

docs/api_reference.rst
```diff
@@ -9,13 +9,40 @@ methods in the current release of PyMC experimental.
...
+=============================
+
+.. automodule:: pymc_experimental.marginal_model
```
OriolAbril, line 15: Do you want all the docstrings for all the functions rendered in the same page? (like it currently happens) or listing the functions in the api page and then have individual pages for each docstring or submodule?

docs/api_reference.rst
```diff
@@ -9,13 +9,40 @@ methods in the current release of PyMC experimental.
...
+   :members: MarginalModel
```
OriolAbril, line 16: This is probably the most important comment. I am not sure how I am supposed to import the MarginalModel class.

It will be documented here as `pymc_experimental.marginal_model.MarginalModel`, but it is also inside the "pymc_experimental" section which makes me want to try using `pymc_experimental.MarginalModel`

docs/api_reference.rst
ricardoV94: Individual pages is probably better in the long term. Is it too much trouble if I ask you to push the needed changes?

docs/api_reference.rst
ricardoV94: For now I would emphasize/require the verbose imports

docs/api_reference.rst
OriolAbril: I can try to do it tomorrow

OriolAbril:
Should be fixed now. I tried running pre-commit locally but something failed generating my environment. Can you finish and merge @ricardoV94?

Look with this PR: https://pymcio--105.org.readthedocs.build/projects/experimental/en/105/api_reference.html

imatge

OriolAbril (approved):

ricardoV94:
Pre-commit seems to be trying to run on Python 3.11 Could that be why it's failing?

OriolAbril:
I have no idea, and didn't have time to dive into the issue the other day

ricardoV94:
Just pre-commit failing now. Anyone knows what it could be? Maybe something that was fixed in one of the other repos already?

ricardoV94:
I rebased from main, I think the CI is passing now. @OriolAbril any change we should do or is it good to merge?

OriolAbril (approved):
Everything good on my end

Issue #109, opened 2023-02-16
consistent naming in model_builder.py


michaelraczycki:
I wanted to propose few minor changes to increase consistency in the code. What I noticed is that all the functions are named with verbs, as usual, but then there's 'id' function. In my opinion it would be better to stick with the rest of the convention and call it 'generate_id'. Second thing is that other functions have type hinting, and in case of id function even though the comment inside the function itself contains description of what type it returns, the hinting is not added. If you agree with those changes I'll create a pull request right away

twiecki:
Maybe we could turn `id` into a `property`?

michaelraczycki:
If you mean @property decorator, then it would also work as well, the only problem I have with those is that if the function is meant to be used also outside o this very file, it creates confusion. Not everyone will know that it's a a function, and from code perspective it will look like a variable.

twiecki:
yeah, `@property` but note that this is part of the internal API anyway.

michaelraczycki:
In this case I'll create new PR to change it to @property

PR #110, opened 2023-02-17
changed id of model_builder to property, added type hinting

michaelraczycki:
As discussed in #109 I've added @property decorator to id function from model_builder
Additionally added type hinting to it


twiecki:
Thanks!

PR #111, opened 2023-02-20
Add utility to convert Model to and from FunctionGraph

ricardoV94:
```python
import pymc as pm
from pytensor import dprint
from pymc_experimental.utils.model_fgraph import fgraph_from_model, model_from_fgraph

with pm.Model (coords={"trial": range(3)}) as m:
    y_obs = pm.MutableData("y_obs", [1., 2., 3.], dims=("trial",))
    mu = pm.Normal("mu")
    sigma = pm.HalfNormal("sigma")
    y = pm.Normal("y", mu, sigma, observed=y_obs, dims=("trial",))

pm.model_to_graphviz(m)

fg = fgraph_from_model(m)
dprint(fg)

... # Manipulate model

new_m = model_from_fgraph(fg)
pm.model_to_graphviz(new_m)
```

### DIFF
--- a/pymc_experimental/utils/model_fgraph.py
+++ b/pymc_experimental/utils/model_fgraph.py
@@ -1,4 +1,4 @@
-from typing import Optional, Sequence, Tuple
+from typing import Dict, Optional, Sequence, Tuple
 
 import pytensor
 from pymc.logprob.transforms import RVTransform
@@ -109,16 +109,19 @@ def local_remove_identity(fgraph, node):
 remove_identity_rewrite = out2in(local_remove_identity)
 
 
-def fgraph_from_model(model: Model) -> FunctionGraph:
+def fgraph_from_model(model: Model) -> Tuple[FunctionGraph, Dict[Variable, Variable]]:
     """Convert Model to FunctionGraph.
 
-    Create a FunctionGraph that includes a copy of model variables, wrapped in dummy `ModelVar` Ops.
-
-    PyTensor rewrites can be used to transform the FunctionGraph.
-
-    It should be possible to reconstruct a valid PyMC model using `model_from_fgraph`.
-
     See: model_from_fgraph
+
+    Returns
+    -------
+    fgraph: FunctionGraph
+        FunctionGraph that includes a copy of model variables, wrapped in dummy `ModelVar` Ops.
+        It should be possible to reconstruct a valid PyMC model using `model_from_fgraph`.
+
+    memo: Dict
+        A dictionary mapping original model variables to the equivalent nodes in the fgraph.
     """
 
     if any(v is not None for v in model.rvs_to_initial_values.values()):
@@ -202,7 +205,19 @@ def fgraph_from_model(model: Model) -> FunctionGraph:
             new_var = var
         new_vars.append(new_var)
 
-    toposort_replace(fgraph, tuple(zip(vars, new_vars)))
+    replacements = tuple(zip(vars, new_vars))
+    toposort_replace(fgraph, replacements)
+
+    # Reference model vars in memo
+    inverse_memo = {v: k for k, v in memo.items()}
+    for var, model_var in replacements:
+        if isinstance(
+            model_var.owner is not None and model_var.owner.op, (ModelDeterministic, ModelNamed)
+        ):
+            # Ignore extra identity that will be removed at the end
+            var = var.owner.inputs[0]
+        original_var = inverse_memo[var]
+        memo[original_var] = model_var
 
     # Remove value variable as outputs, now that they are graph inputs
     first_value_idx = len(fgraph.outputs) - len(value_vars)
@@ -212,7 +227,7 @@ def fgraph_from_model(model: Model) -> FunctionGraph:
     # Now that we have Deterministic dummy Ops, we remove the noisy `Identity`s from the graph
     remove_identity_rewrite.apply(fgraph)
 
-    return fgraph
+    return fgraph, memo
 
 
 def model_from_fgraph(fgraph: FunctionGraph) -> Model:
@@ -282,7 +297,7 @@ def model_from_fgraph(fgraph: FunctionGraph) -> Model:
     return model
 
 
-def clone_model(model: Model) -> Model:
+def clone_model(model: Model) -> Tuple[Model]:
     """Clone a PyMC model.
 
     Recreates a PyMC model with clones of the original variables.
@@ -310,4 +325,4 @@ def clone_model(model: Model) -> Model:
             z = pm.Deterministic("z", clone_x + 1)
 
     """
-    return model_from_fgraph(fgraph_from_model(model))
+    return model_from_fgraph(fgraph_from_model(model)[0])
--- a/pymc_experimental/utils/model_fgraph.py
+++ b/pymc_experimental/utils/model_fgraph.py
@@ -1,0 +1,313 @@
+from typing import Optional, Sequence, Tuple
+
+import pytensor
+from pymc.logprob.transforms import RVTransform
+from pymc.model import Model
+from pymc.pytensorf import find_rng_nodes
+from pytensor import Variable
+from pytensor.graph import Apply, FunctionGraph, Op, node_rewriter
+from pytensor.graph.rewriting.basic import out2in
+from pytensor.scalar import Identity
+from pytensor.tensor.elemwise import Elemwise
+
+from pymc_experimental.utils.pytensorf import StringType
+
+
+class ModelVar(Op):
+    """A dummy Op that describes the purpose of a Model variable and contains
+    meta-information as additional inputs (value and dims).
+    """
+
+    def make_node(self, rv, *dims):
+        assert isinstance(rv, Variable)
+        dims = self._parse_dims(rv, *dims)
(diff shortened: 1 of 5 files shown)


pymc_experimental/utils/model_fgraph.py
```diff
@@ -0,0 +1,174 @@
...
+            return Apply(self, [rv, *dims], [rv.type()])
+
+    def infer_shape(self, fgraph, node, inputs_shape):
+        return inputs_shape[0]
+
+    def do_constant_folding(self, fgraph, node):
+        return False
+
+    def perform(self, *args, **kwargs):
+        raise RuntimeError("ValuedRVs should never be evaluated!")
```
ricardoV94, line 40:
```suggestion
        raise RuntimeError("ModelVars should never be in a final graph!")
```

pymc_experimental/utils/model_fgraph.py
```diff
@@ -0,0 +1,174 @@
...
+
+import pytensor
+from pymc.model import Model
+from pytensor.graph import Apply, FunctionGraph, Op
+from pytensor.tensor import TensorVariable
+
+from pymc_experimental.utils.pytensorf import StringType
+
+
+class ModelVar(Op):
```
lucianopaz, line 11: My main concern is that you are wrapping every random variable with variants of this `Op`. That might make it harder for rewrites to work. For example, imagine a model that uses the non centered parametrization and you want to rewrite it into the centered parametrization. You would need to look at many things in the rewrite: whether the two entities that are being summed together are `FreeRV`s or if its a single `FreeRV` plus some product of another thing. This extra `Op` layer, in my opinion, complicates the operation of the rewrites.

From what I understood, the main advantage of these new `Op`s is that they track two things: the kind of variable within the model, and the dimension information of the variable. I know that it's a bit of a crappy thing to ask, but have you considered using `tag`s to keep track of the variable kind instead of `Op`s?

pymc_experimental/utils/model_fgraph.py
```diff
@@ -0,0 +1,174 @@
...
+    # We start the `dict` with mappings from the value variables to themselves,
+    # to prevent them from being cloned.
+    memo = {v: v for v in values}
+
+    fgraph = FunctionGraph(
+        outputs=rvs + potentials + deterministics,
+        clone=True,
+        memo=memo,
+        copy_orphans=False,
+        copy_inputs=False,
```
lucianopaz, line 105: Does this mean that each `RandomGenerator` will get its own copy of the shared random stream? Or will that underlying memory still be shared? If yes, that isn't what I think that we want. I think that we want the new variables to use their own copied versions of the random streams or generators. By copies I mean that they should be two different generator objects that have the exact same initial random state. That way, they would yield the same sequence of random numbers, but getting random numbers from the new model would not alter the state of the old model.

pymc_experimental/utils/model_fgraph.py
```diff
@@ -0,0 +1,174 @@
...
+            model.observed_RVs.append(var)
+            model.create_value_var(var, transform=None, value_var=value)
+        elif isinstance(model_var.owner.op, Potential):
+            var, *dims = model_var.owner.inputs
+            model.potentials.append(var)
+        elif isinstance(model_var.owner.op, Deterministic):
+            var, *dims = model_var.owner.inputs
+            model.deterministics.append(var)
+        else:
+            continue  # Raise?
```
lucianopaz, line 166: How would you ever reach this?

lucianopaz (changes requested):
@ricardoV94, thanks for being able to kick this off. I have a bunch of comments, but I feel that most of the design decisions should be discussed elsewhere. I've opened [this discussion](https://github.com/pymc-devs/pymc-experimental/discussions/112) to track the design stuff. My main critique is that the new `Op` that you introduced will make it harder to apply rewrites, because they will have to weigh what to do about the `FreeRV` or `ObservedRV` or `Deterministic` wrapping `Op`. I think that we can continue that discussion in the thread that I linked.

pymc_experimental/utils/model_fgraph.py
ricardoV94: > whether the two entities that are being summed together are FreeRVs or if its a single FreeRV plus some product of another thing. This extra Op layer, in my opinion, complicates the operation of the rewrites.

I responded in the discussion. I think we want to have to look at what is a `FreeRV` and what's not, because you want to explicitly state what old `FreeRV` is to be replaced by the new centered one (that is move the value_variable one node lower). I don't think the rewrite is any more hard, the rule would be:
```
shift + FreeRV(Normal(mu, sigma)) * scale  -> FreeRV(Normal(mu + shift, sigma * scale)
```
A bit of a forced case, but you wouldn't want this rewrite to do this, on top of the following
```
shift + ObservedRV(Normal(mu, sigma) * scale
```
So you would always need to inspect the Value variable type anyway. Then you might have it in the graph already.

pymc_experimental/utils/model_fgraph.py
ricardoV94: I don't know what happens to shared variables, here I was thinking of not copying the non-shared value variables. Will have to take a look. I think we should copy them, just so that two compiled random functions don't interfere with each other.

About the initial state, that doesn't matter. Whenever we compile a function with `pytensorf.compile_pymc` we override the RNG values based on `random_seed` anyway. The current global values are never taken into consideration (and I think they shouldn't as it creates a very cumbersome internal state to our PyMC models)

pymc_experimental/utils/model_fgraph.py
ricardoV94: I was just dumping code. I guess if a user defined a new type of `ModelVar`

pymc_experimental/utils/model_fgraph.py
ricardoV94: Actually we probably want to reuse shared variables other than RNGs (eg mutableData), so maybe I will do a replace rngs after cloning but otherwise reuse the same shared variables

pymc_experimental/utils/model_fgraph.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # TODO: Do something with these
+    dim_lengths = model.dim_lengths
+    rvs_to_transforms = model.rvs_to_transforms
+
+    # Not supported yet
+    if any(v is not None for v in model.rvs_to_total_sizes.values()):
+        raise NotImplementedError("Cannot convert models with total_sizes")
+    if any(v is not None for v in model.rvs_to_initial_values.values()):
+        raise NotImplementedError("Cannot convert models with non-default initial_values")
```
ricardoV94, line 94: If model has parent raise. We don't support nested models specifically yet

pymc_experimental/utils/model_fgraph.py
```diff
@@ -0,0 +1,174 @@
...
+    for var in vars:
+        dims = vars_to_dims.get(var.name, None)
+        if var in free_rvs_to_values:
+            new_var = free_rv(var, free_rvs_to_values[var], dims)
+        elif var in observed_rvs_to_values:
+            new_var = observed_rv(var, observed_rvs_to_values[var], dims)
+        elif var in potentials:
+            new_var = potential(var, dims)
+        elif var in deterministics:
+            new_var = deterministic(var, dims)
```
ricardoV94, line 126: We might want to pull Deterministics out of the graph in `Fgraph` (so they don't bother us during rewrites) and back into the graph in `Model` (so they print nicely in graphviz).

pymc_experimental/utils/model_fgraph.py
```diff
@@ -0,0 +1,174 @@
...
+            raise RuntimeError(f"Variable is not RV, Potential nor Deterministic: {new_var}")
+        new_vars.append(new_var)
+
+    toposort_replace(fgraph, tuple(zip(vars, new_vars)))
+    return fgraph
+
+
+def model_from_fgraph(fgraph: FunctionGraph) -> Model:
+
+    model = Model(coords=getattr(fgraph, "coords", None))
```
ricardoV94, line 137: Fail if we are in a model context, or this would be a submodel of the parent

pymc_experimental/utils/model_fgraph.py
ricardoV94: Decided to just raise a warning, it shouldn't fail

pymc_experimental/utils/model_fgraph.py
ricardoV94: Replaced RNGs now

ricardoV94:
Ready for review!

lucianopaz (commented):
This looks really great @ricardoV94. I’ve got two comments:
1. This is a small nit, but could you add a todo list with the things that still aren’t covered by the transformation to fgraph. At the moment I can think of the initval, mutable dimensions, but there must be some more.
2. I can’t wrap my head around the special care you take with deterministics. It looks like the rebuilt model might have a different structure of the deterministics than the original one? At least the test that you wrote with `y_` and `y__` make it seem that way to me

lucianopaz (approved):
I think this looks well enough to merge. Let’s try it out on the actual applications to see if we need to adapt anything

PR #119, opened 2023-03-07
Model Builder refactoring

michaelraczycki:
In order to make the model_builder a template class for other model-related pymc classes we need to restructure its constructor. So far it was assuming presence of build_model function in child class which is not always going to be the case.

changes:
build() function call was moved out of the constructor, and now is called from fit function. This change allows inheritance from ModelBuilder by abstract classes like baseMMM


michaelraczycki:
@twiecki this is just a starter of final PR, let's not merge it yet as I might find other necessary changes while trying to integrate ModelBuilder with BaseMMM.

michaelraczycki:
As recently we've came to a conclusion that for variety of projects it will be unsuitable to have Model Builder as a child of pm.Model, this PR will be used for refactoring tasks that will prepare new structure of ModelBuilder. 

General changes will include:
removing the inheritance from pm.Model
adding class attribute containing an pm.Model object
adaptation of existing functions to accommodate the changes
adaptation of tests

pymc_experimental/tests/test_model_builder.py
```diff
@@ -24,12 +28,13 @@ class test_ModelBuilder(ModelBuilder):
     _model_type = "LinearModel"
     version = "0.1"
 
-    def build_model(self, model_config, data=None):
-
+    def build_model(self, model_instance, model_config, data=None):
+        model_instance.model_config = model_config
```
twiecki, line 32: I think we can just keep the `model_config` and `data` in the base class (instance `self`). The init() method should also take care of that.

pymc_experimental/tests/test_model_builder.py
```diff
@@ -24,12 +28,13 @@ class test_ModelBuilder(ModelBuilder):
...
+        model_instance.data = data
         self.model_config = model_config
         self.data = data
 
-        with pm.Model() as self.model:
```
twiecki: I think this should stay. Ultimately we want:
1. `ModelBuilder` base-class that contains all the config dicts and data
2. `model_builder.model` which is an instance of `pm.Model()`.

michaelraczycki: that's the outdated version. I've replaced it, but now build_model receives instance of a ModelBuilder as an argument, so it can assign the created pm.Model() directly as the instance_model.model

twiecki:
If you rebase from main again, https://github.com/pymc-devs/pymc-experimental/pull/125 should have fixed CI.

pymc_experimental/model_builder.py
```diff
@@ -37,8 +37,8 @@ class ModelBuilder:
     def __init__(
         self,
         model_config: Dict,
```
twiecki, line 39: make model_config also optional and move down.

michaelraczycki:
@twiecki it seem that after rebasing the original issue with test is solved, but a lot of not-related tests fail both in ubuntu and windows

twiecki:
Hm, I think something went wrong with the rebasing, those errors should be fixed.

michaelraczycki:
Ok, tried to do rebase again, hopefully runs good this time

docs/_templates/autosummary/base.rst
```diff
@@ -0,0 +1,5 @@
+{{ objname | escape | underline }}
```
twiecki, line 1: I think we made it worse :P, this hsouldn't show up here. Maybe start with a fresh PR?

michaelraczycki: ok, I'll try to just copy the changes to new branch, branched from fresh main and see if it helps

PR #125, opened 2023-03-29
Test on last release of PyMC instead of main

ricardoV94:
Otherwise it's a bit of a mess when we are changing API across releases

We might even undo changes before a release, and we shouldn't be patching the code here until it becomes "fixed".

Related to #124
Closes #123

PR #127, opened 2023-03-30
Install pymc via conda

ricardoV94:
Follow up to try and remove last commit of #125


michaelraczycki:
@ricardoV94 I believe that the build is failing because actually anaconda doesn't contain any system specific distribution for pymc>4.1
Screenshot 2023-04-28 at 14 05 30
I'm searching for possible solutions now

wd60622:
Seems like this is good to close after #403

PR #131, opened 2023-03-31
Model Builder refactoring

michaelraczycki:
New branch containing changes from #119, created because of unexpected problems with rebasing done in order to cover CI test bugs.


pymc_experimental/model_builder.py
```diff
@@ -38,8 +37,8 @@ class ModelBuilder(pm.Model):
     def __init__(
         self,
         model_config: Dict,
```
twiecki, line 39: optional

pymc_experimental/model_builder.py
```diff
@@ -187,22 +192,29 @@ def load(cls, fname):
 
         filepath = Path(str(fname))
         idata = az.from_netcdf(filepath)
-        self = cls(
-            json.loads(idata.attrs["model_config"]),
-            json.loads(idata.attrs["sampler_config"]),
-            idata.fit_data.to_dataframe(),
+        if "sampler_config" in idata.attrs:
```
twiecki, line 195: I think we should assume that it's always there.

pymc_experimental/model_builder.py
```diff
@@ -187,22 +192,29 @@ def load(cls, fname):
...
+            sampler_config = json.loads(idata.attrs["sampler_config"])
+        else:
+            sampler_config = {}
+        model_builder = cls(
+            model_config=json.loads(idata.attrs["model_config"]),
+            sampler_config=sampler_config,
```
twiecki, line 201:
```suggestion
            sampler_config=json.loads(idata.attrs["sampler_config"]),
```

pymc_experimental/model_builder.py
```diff
@@ -225,35 +237,40 @@ def fit(self, data: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]] = None
...
-
-        if self.basic_RVs == []:
             self.build()
+            self._data_setter(data)
 
-        with self:
-            self.idata = pm.sample(**self.sample_config)
+        with self.model:
+            if self.sampler_config:
+                self.idata = pm.sample(**self.sampler_config)
```
twiecki, line 245: This will work even if `sampler_config` is an empty dict, so you can just remove the if check.

pymc_experimental/model_builder.py
```diff
@@ -225,35 +237,40 @@ def fit(self, data: Dict[str, Union[np.ndarray, pd.DataFrame, pd.Series]] = None
...
+            else:
+                self.idata = pm.sample()
             self.idata.extend(pm.sample_prior_predictive())
             self.idata.extend(pm.sample_posterior_predictive(self.idata))
 
         self.idata.attrs["id"] = self.id
         self.idata.attrs["model_type"] = self._model_type
         self.idata.attrs["version"] = self.version
-        self.idata.attrs["sampler_config"] = json.dumps(self.sample_config)
+        if self.sampler_config:
```
twiecki, line 254: same, just remove

pymc_experimental/model_builder.py
michaelraczycki: ah okey. Actually you're right, I don't know why I assumed that if we agreed to make he sampler config and model config as optional then we'd not assign empty attribute spot to sampler config. But I agree that to make it more logically continuous we should add whatever is there, whether is a valid config dict or just an empty dict.

pymc_experimental/model_builder.py
michaelraczycki: same as the previous one, I wrongly assumed that in case the user wouldn't pass the config we'd just fill it with empty one, ignoring the fact that actually default value of config should be preserved in the saved form of model

pymc_experimental/model_builder.py
michaelraczycki: done

michaelraczycki:
@ricardoV94 I'll make sure to make more granular commits from now on and make them more descriptive

pymc_experimental/model_builder.py
michaelraczycki: covered

michaelraczycki:
@twiecki requested changes are in. Do you think we should keep this PR as draft and keep adding on top of the present changes, or keep merging and opening new PR's?

pymc_experimental/model_builder.py
```diff
@@ -60,20 +59,30 @@ def __init__(
         """
 
         super().__init__()
```
twiecki, line 61:
```suggestion
```

pymc_experimental/model_builder.py
```diff
@@ -60,20 +59,30 @@ def __init__(
...
-    def build(self):
+    def build(self) -> None:
         """
         Builds the defined model.
         """
 
-        with self:
-            self.build_model(self.model_config, self.data)
+        self.build_model(
+            model_instance=self,
```
twiecki, line 79: Why does this need to be passed another `self`?

michaelraczycki: It doesn't. It's something left from very early stages of refactoring and forgot to replace it with just selfs afterwards

pymc_experimental/model_builder.py
```diff
@@ -60,20 +59,30 @@ def __init__(
...
-        self.idata = None  # inference data object
+        self.sampler_config = sampler_config  # parameters for sampling
         self.data = data
-        self.build()
+        self.idata = (
+            None  # inference data object placeholder, idata is generated during build execution
+        )
 
-    def build(self):
+    def build(self) -> None:
```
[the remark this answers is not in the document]
michaelraczycki: I've removed it and used the direct call of build_model in fit function

twiecki:
That looks great now!

twiecki (approved):

Issue #102, opened 2022-12-18
Fail tests with uncaught warnings 


ricardoV94:
The earlier we start, the easier it is :)

michaelraczycki:
do you mean to change the pytest settings to strict? Because I believe that by default pytest should fail test if it has uncaught exception

ricardoV94:
Oh I mean uncaught warnings

michaelraczycki:
I'd like to do it then, do you have something specific in mind? I'd just create the pytest.ini file and use "addopts = -W error" global setting, then all warnings should fail the test. Later on we can exclude specific types of warnings if need be

ricardoV94:
According to pytest you can specify

```
[pytest]
filterwarnings =
    error
```

https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html

Yes that should be a good start.

PR #133, opened 2023-04-03
Make pytest fail on uncaught warnings

michaelraczycki:
introduced config file (pytest.ini) in a root of the repository, added configuration failing tests if containing uncaught warnings. Added ignores for already known and ofter appearing warnings that won't be resolved in the near future 
(namespace deprecation warning and arviz inference data warning that warns about inference data not being registered in InferenceData scheme

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+filterwarnings =
+    error
+    ignore:.*?(\b(pkg_resources\.declare_namespace)\b).*:DeprecationWarning
+    ignore::UserWarning:arviz.data.inference_data


michaelraczycki:
addresses changes requested in #102

michaelraczycki:
I guess it's one of the few situations where the pytest / ubuntu failing might signify that my work was a success :)

michaelraczycki:
@ricardoV94 if you will make a list of which warnings we want to ignore for now, and not make them fail tests I can add regex exceptions for now in pytest.ini. I think this step will be quite a shock, as every single test has some warnings, and we will get test failing here unless we'd go one by one and adapt all the tests (did you think of that as within a scope of that issue?)

ricardoV94:
The first step would be to fix the FutureWarnings in the codebase directly. For instance, the first one I see is pretty straightforward:

```python
=================================== FAILURES ===================================
_______________________________ test_pathfinder ________________________________

    @pytest.mark.skipif(sys.platform == "win32", reason="JAX not supported on windows.")
    def test_pathfinder():
        # Data of the Eight Schools Model
        J = 8
        y = np.array([28.0, 8.0, -3.0, 7.0, -1.0, 1.0, 18.0, 12.0])
        sigma = np.array([15.0, 10.0, 16.0, 11.0, 9.0, 11.0, 10.0, 18.0])
    
        with pm.Model() as model:
    
            mu = pm.Normal("mu", mu=0.0, sigma=10.0)
            tau = pm.HalfCauchy("tau", 5.0)
    
            theta = pm.Normal("theta", mu=0, sigma=1, shape=J)
            theta_1 = mu + tau * theta
            obs = pm.Normal("obs", mu=theta, sigma=sigma, shape=J, observed=y)
    
>           idata = pmx.fit(method="pathfinder", iterations=100)

pymc_experimental/tests/test_pathfinder.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pymc_experimental/inference/fit.py:37: in fit
    return fit_pathfinder(**kwargs)
pymc_experimental/inference/pathfinder.py:141: in fit_pathfinder
    for var_name, dims in model.RV_dims.items()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pymc.model.Model object at 0x7f90a684d2b0>

    @property
    def RV_dims(self) -> Dict[str, Tuple[Union[str, None], ...]]:
        """Tuples of dimension names for specific model variables.
    
        Entries in the tuples may be ``None``, if the RV dimension was not given a name.
        """
>       warnings.warn(
            "Model.RV_dims is deprecated. User Model.named_vars_to_dims instead.",
            FutureWarning,
        )
E       FutureWarning: Model.RV_dims is deprecated. User Model.named_vars_to_dims instead.
```

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+filterwarnings =
+    error
+    ignore:.*?(\b(pkg_resources\.declare_namespace)\b).*:DeprecationWarning
```
ricardoV94, line 4: What are we ignoring here?

michaelraczycki: it's a warning about a way namespaces are declared in pymc-experimental, it's purely python convention related, so I thought that for now it shouldn't be the highest priority

michaelraczycki: pkg_resources.declare_namespace() is throwing a warning

michaelraczycki:
It's hard to say, because a lot of errors break the test suite on collection, in model_builder related tests every test was breaking because of the namespace and arviz warning, that's why I added the ignores. I contacted @OriolAbril about the arviz warning, and also got a confirmation that it's nothing to worry about now

pytest.ini
ricardoV94: Sounds good to ignore

ricardoV94:
That's good so it seems only the failing test needs to be addressed and this PR is good to go?

michaelraczycki:
I can fix it tomorrow, and hopefully it's good to go

michaelraczycki:
@ricardoV94 it seems like 2 things were still failing (I can change the pathfinder test code but I can't check it locally, as it requires x86_64 architecture and me with Mac have arm64 I think we should extend the pytest.mark.skipif for Macs as well.

pymc_experimental/tests/test_pathfinder.py
```diff
@@ -21,6 +22,7 @@
 import pymc_experimental as pmx
 
 
+@pytest.mark.skipif(platform.machine() != "x86_64", reason="Test requires x86_64 architecture")
```
ricardoV94, line 25: Why was this needed? If there's a github issue can you add as inline comment?

ricardoV94: Can you share the error message?

ricardoV94: Then I think we should remove the skip. The AVX instructions are a bit more CPU specific than just "x86_64"

michaelraczycki: https://en.wikipedia.org/wiki/Advanced_Vector_Extensions
AVX were used only in x86, whether by intel or amd ( at least that's what I keep finding)

ricardoV94: Sorry I read the other way around, that you had a x86 without AVX. 

But then are you saying JAX cannot run on M1 Macs at all, or just your installation?

ricardoV94:
Thanks @michaelraczycki

PR #135, opened 2023-04-04
adapted the release setup to mimic one used in pymc-marketing

michaelraczycki:
As pointed out in #134  by @ricardoV94, pymc-experimental doesn't use the same versioning system as main pymc, so the solution taken from there wouldn't work. This PR contains changes that adopt the solution used in pymc-marketing


michaelraczycki:
I'm not quite understanding the numpy import error here, @ricardoV94 , @lucianopaz , @twiecki maybe you've seen something like this before? It seems like there's no reason numpy would be required for building sdist

ricardoV94:
You might be building it the wrong way, maybe a difference between using `pyproject.toml` and `setup.py`. 

Have you tried to do it locally?

michaelraczycki:
@ricardoV94 , after the changes I've introduced now it runs locally without issues, but here still seems to be failing. I'll keep searching for some solution but for now it's going to be guessing game a bit

ricardoV94 (approved):
Let's give it a try?

PR #138, opened 2023-04-12
Model builder adaptations

michaelraczycki:


docs/_templates/autosummary/class.rst
```diff
@@ -0,0 +1,29 @@
+{{ objname | escape | underline}}
```
twiecki, line 1: Should this stuff be showing up here?

PR #142, opened 2023-04-16
adaptations to integrate with mmm

michaelraczycki:
changes to ModelBuilder that are required for full integration with MMM, the final shape is still open for discussion.
main changes:
-build_model moved out of constructor of baseMMM and called in fit() in model_builder
-model is not build at initialisation of MMM, but with first fit, reason is that chain calling constructors was causing issues with object creation timing resulting in not reliable inheritance

### DIFF
--- a/pymc_experimental/model_builder.py
+++ b/pymc_experimental/model_builder.py
@@ -23,6 +23,8 @@ import arviz as az
 import numpy as np
 import pandas as pd
 import pymc as pm
+from pymc.util import RandomState
+from traitlets import Any
 
 
 class ModelBuilder:
@@ -191,7 +193,7 @@ class ModelBuilder:
             data=idata.fit_data.to_dataframe(),
         )
         model_builder.idata = idata
-        model_builder.build()
+        model_builder.build_model()
         if model_builder.id != idata.attrs["id"]:
             raise ValueError(
                 f"The file '{fname}' does not contain an inference data of the same model or configuration as '{cls._model_type}'"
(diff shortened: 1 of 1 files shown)


ricardoV94 (dismissed):
LGTM, any specific thing you need input on?

pymc_experimental/model_builder.py
```diff
@@ -114,12 +115,15 @@ def create_sample_input():
...
 
         >>>    model_config = {
-        >>>       'a_loc': 7,
-        >>>       'a_scale': 3,
-        >>>       'b_loc': 5,
-        >>>       'b_scale': 3,
-        >>>       'obs_error': 2,
-        >>>    }
+        >>>          'a' : {
+        >>>              'a_loc': 7,
```
twiecki, line 119: Should this not just be `loc`, the `a` seems redundant.

pymc_experimental/model_builder.py
```diff
@@ -191,7 +220,7 @@ def load(cls, fname: str):
             data=idata.fit_data.to_dataframe(),
         )
         model_builder.idata = idata
-        model_builder.build()
+        model_builder.idata = model_builder.fit()
```
twiecki, line 223: What if the model was already fit?

pymc_experimental/model_builder.py
michaelraczycki: true, good catch

twiecki:
:+1:

PR #147, opened 2023-04-20
Model builder docstrings update

michaelraczycki:
Updates the docstrings from model_builder.py, ensures they follow the numpydoc conventions.


michaelraczycki:
forgot to switch to up-to-date branch before creating new branch, so the PR would carry commits that are already in main

PR #148, opened 2023-04-20
docstrings update in model_builder.py

michaelraczycki:
Updates the docstrings in model_builder.py , ensures that they follow the numpydoc conventions

### DIFF
--- a/pymc_experimental/model_builder.py
+++ b/pymc_experimental/model_builder.py
@@ -131,9 +131,15 @@ class ModelBuilder:
         >>>       'chains': 1,
         >>>       'target_accept': 0.95,
         >>>    }
-        >>>    return data, model_config, sampler_config
+        Returns
+        -------
+        data : dict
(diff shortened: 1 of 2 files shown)


pymc_experimental/model_builder.py
```diff
@@ -131,9 +131,15 @@ def create_sample_input():
         >>>       'chains': 1,
         >>>       'target_accept': 0.95,
         >>>    }
-        >>>    return data, model_config, sampler_config
```
twiecki, line 134: I think this should stay.

michaelraczycki: oh, my bad. I was quite sure I added the Return section without deleting the example

michaelraczycki: it's back again

twiecki:
Conflict, but LGTM otherwise.

twiecki (approved):

PR #149, opened 2023-04-20
removed unused *args from fit, added named args to sampler_config

michaelraczycki:
I noticed that named parameters added in last model_builder PR were not added in the sampler_config during fit() execution, 
which would result in them not being preserved in the saved version of the model. This PR fixes this


twiecki (approved):

michaelraczycki:
@twiecki the windows based pytest failed, but it looks like something that the PR change couldn't influence. Should we re-run the test or just ignore it and merge regardless?

PR #161, opened 2023-04-30
Initial draft of version of model_builder to work with scikit-learn.

### LINKED ISSUE #155: model_builder scikit-learn integration
I am working on making an existing scikit-learn model pipeline produce probabilistic output. To do that, I used model_builder to make a pymc model that could integrate into a scikit-learn `Pipeline`, including standardization of inputs and outputs. However, I find that the current API doesn't seem suitable for this. I made my own modifications to the `ModelBuilder` class and [example](https://www.pymc.io/projects/examples/en/latest/howto/model_builder.html) `LinearModel` subclass to get it to work. I think the main change was to have the `fit` and `predict` methods take X and y as separate parameters rather than as members of a data dict with specially-named keys. My reference for the scikit-learn estimator API is the scikit-learn [documentation](https://scikit-learn.org/stable/developers/develop.html) and  [template](https://github.com/scikit-learn-contrib/project-template/blob/master/skltemplate/_template.py) for TemplateEstimator.

I very well might be one the wrong track (or at least on a different one than what model_builder intends), but what I came up with seems to work for being able to apply `sklearn.preprocessing.StandardScaler` to inputs and to point outputs using `sklearn.compose.TransformedTargetRegressor`. These seem like reasonable goals for `ModelBuilder` subclasses to be able to integrate with, so maybe tests and/or examples of such would be good.

Any thoughts? I'm happy to contribute what I can.

pdb5627:
Resolves #155

### DIFF
--- a/pymc_experimental/bayesian_estimator_linearmodel.py
+++ b/pymc_experimental/bayesian_estimator_linearmodel.py
@@ -431,8 +431,8 @@ class LinearModel(BayesianEstimator):
         # Actual data will be set by _data_setter when fitting or evaluating the model.
         # Data array size can change but number of dimensions must stay the same.
         with pm.Model() as self.model:
-            x = pm.MutableData("x", np.zeros((1,)), dims='observation')
-            y_data = pm.MutableData("y_data", np.zeros((1,)), dims='observation')
+            x = pm.MutableData("x", np.zeros((1,)), dims="observation")
+            y_data = pm.MutableData("y_data", np.zeros((1,)), dims="observation")
 
             # priors
             intercept = pm.Normal(
@@ -442,10 +442,17 @@ class LinearModel(BayesianEstimator):
             obs_error = pm.HalfNormal("σ_model_fmc", cfg["obs_error"])
 
             # Model
-            y_model = pm.Deterministic("y_model", intercept + slope * x, dims='observation')
+            y_model = pm.Deterministic("y_model", intercept + slope * x, dims="observation")
 
             # observed data
-            y_hat = pm.Normal("y_hat", y_model, sigma=obs_error, shape=x.shape, observed=y_data, dims='observation')
+            y_hat = pm.Normal(
+                "y_hat",
+                y_model,
+                sigma=obs_error,
+                shape=x.shape,
+                observed=y_data,
+                dims="observation",
+            )
             self.output_var = "y_hat"
 
     def _data_setter(self, X, y=None):
--- a/pymc_experimental/bayesian_estimator_linearmodel.py
+++ b/pymc_experimental/bayesian_estimator_linearmodel.py
@@ -10,19 +10,19 @@ import pymc as pm
 import xarray as xr
 from pymc.util import RandomState
 
-from .model_builder import ModelBuilder
+from pymc_experimental.model_builder import ModelBuilder
 
 # If scikit-learn is available, use its data validator
 try:
-    from sklearn.utils.validation import check_X_y, check_array
+    from sklearn.utils.validation import check_array, check_X_y
 # If scikit-learn is not available, return the data unchanged
 except ImportError:
+
     def check_X_y(X, y, **kwargs):
         return X, y
 
     def check_array(X, **kwargs):
         return X
-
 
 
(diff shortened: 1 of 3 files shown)


twiecki:
This looks great! I wonder if perhaps we should just make the base class directly use this API.

michaelraczycki:
Nice job on the fit refactoring! From what I see the model_builder was already compatible with scikit-learn or you didn't finish pushing the changes yet?

pdb5627:
> Nice job on the fit refactoring! From what I see the model_builder was already compatible with scikit-learn or you didn't finish pushing the changes yet?

@michaelraczycki  No, it wasn't compatible. I made a new file `bayesian_estimator_linearmodel.py` for a subclass of `ModelBuilder` and an example implemention subclass.  I introduced new tests in a new file `tests/test_bayesian_estimator_linearmodel.py`. The refactoring in `ModelBuilder` was to put the sampling into a function that wouldn't be overridden in the subclass. I also pulled out the attr-setting code so that it could be reused by a function for sampling the prior predictive distribution.

pdb5627:
Here's a comparison of the APIs:

| ModelBuilder                                                  | BayesianEstimator           | Comment                                                                                                          |
|---------------------------------------------------------------|-----------------------------|------------------------------------------------------------------------------------------------------------------|
| __init__                                                      | __init__                    | Remove data from constructor. Default args come from properties.                                                 |
|                                                               | default_model_config ***    |                                                                                                                  |
|                                                               | default_sampler_config ***  |                                                                                                                  |
|                                                               | _validate_data              |                                                                                                                  |
| _data_setter ***                                              | _data_setter ***            | Adapted for X,y rather than data                                                                                 |
| create_sample_input ***                                       | create_sample_input ***     | Creates just data, not configuration. Default args from properties.                                              |
| build_model ***                                               | build_model ***             | Takes no arguments. Instead of setting MutableData values, puts in dummy data of the right number of dimensions. |
| sample_model                                                  |                             |                                                                                                                  |
| save                                                          |                             |                                                                                                                  |
| load                                                          | load                        | Adapted for different constructor arguments                                                                      |
| fit                                                           | fit                         | Adapted for X,y rather than data                                                                                 |
| predict                                                       | predict                     | Adapted for X,y rather than data                                                                                 |
| predict_posterior                                             | predict_posterior           | Adapted for X,y rather than data                                                                                 |
|                                                               | predict_proba               |                                                                                                                  |
|                                                               | sample_prior_predictive     |                                                                                                                  |
|                                                               | sample_posterior_predictive |                                                                                                                  |
| _serializable_model_config                                    |                             |                                                                                                                  |
| id                                                            |                             |                                                                                                                  |
                                                                                                               
 *** = abstract method requiring implementation in subclasses.

michaelraczycki:
Oh, I focused much more on the changes you've introduced to model_builder itself, I didn't go into the implementation of Bayesian Estimator that much in detail as It doesn't affect other projects so far. I really like the ideas you're proposing, could we schedule a short call to discuss Bayesian Estimator a bit in depth? I'm working on making the model_builder to a standard used by other pymc-related projects, and I wanted to discuss with you which of the API changes you've suggested are complete must-have for sklearn compatibility

pdb5627:
> Oh, I focused much more on the changes you've introduced to model_builder itself, I didn't go into the implementation of Bayesian Estimator that much in detail as It doesn't affect other projects so far. I really like the ideas you're proposing, could we schedule a short call to discuss Bayesian Estimator a bit in depth? I'm working on making the model_builder to a standard used by other pymc-related projects, and I wanted to discuss with you which of the API changes you've suggested are complete must-have for sklearn compatibility

@michaelraczycki Ok, sounds good. How would you like to set it up? I'm in UTC+3.

michaelraczycki:
I'm in UTC+0, so anything after your lunch should work

michaelraczycki:
@pdb5627 is there something else you'd like to add here?

michaelraczycki (approved):
Amazing job, thank you for your contribution!

twiecki:
Indeed, this is great @pdb5627!

pdb5627:
Thanks, guys. I'm glad I could contribute.

Issue #160, opened 2023-04-30
Docs broken, possibly due to versioneer


ricardoV94:
The API can't be accessed anymore https://www.pymc.io/projects/experimental/en/latest/index.html

michaelraczycki:
the api reference works again, solved in #166

PR #162, opened 2023-05-01
fixing __init__.py to use versioneer

michaelraczycki:
fix for #160

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -41,4 +41,4 @@ See the `GitHub contributor page <https://github.com/pymc-devs/pymc-experimental
 .. toctree::
    :hidden:
 
-   api_reference
+   api_reference <https://www.pymc.io/projects/experimental/en/latest/api_reference.html>
--- a/.pylintrc
+++ b/.pylintrc
@@ -370,3 +370,6 @@ int-import-graph=
 # Exceptions that will emit a warning when being caught. Defaults to
 # "Exception"
 overgeneral-exceptions=Exception
+
+[pylint.'pymc_experimental._version']
+disable=E0601
--- a/pymc_experimental/__init__.py
+++ b/pymc_experimental/__init__.py
@@ -13,8 +13,6 @@
 #   limitations under the License.
 import logging
 
-import versioneer
-
 _log = logging.getLogger("pmx")
 
 if not logging.root.handlers:
@@ -23,8 +21,9 @@ if not logging.root.handlers:
         handler = logging.StreamHandler()
         _log.addHandler(handler)
 
-from pymc_experimental import _version, distributions, gp, utils
+from pymc_experimental import distributions, gp, utils
+from pymc_experimental._version import get_versions
 from pymc_experimental.inference.fit import fit
 from pymc_experimental.marginal_model import MarginalModel
 
-__version__ = versioneer.get_version()
+__version__ = get_versions()["version"]
--- a/pymc_experimental/__init__.py
+++ b/pymc_experimental/__init__.py
@@ -13,6 +13,8 @@
 #   limitations under the License.
 import logging
 
+import versioneer
+
 _log = logging.getLogger("pmx")
 
 if not logging.root.handlers:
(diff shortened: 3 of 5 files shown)


twiecki (approved):

twiecki (changes requested):
Successfully installed arviz-0.15.1 cachetools-5.3.0 cloudpickle-2.2.1 cons-0.4.5 contourpy-1.0.7 cycler-0.11.0 etuples-0.3.8 fastprogress-1.0.3 filelock-3.12.0 fonttools-4.39.3 h5netcdf-1.1.0 h5py-3.8.0 importlib-resources-5.12.0 kiwisolver-1.4.4 logical-unification-0.4.5 matplotlib-3.7.1 miniKanren-1.0.3 multipledispatch-0.6.0 numpy-1.24.3 packaging-23.1 pandas-2.0.1 pillow-9.5.0 pymc-5.3.1 pymc-experimental-0+untagged.108.g1c2d064 pyparsing-3.0.9 pytensor-2.11.2 python-dateutil-2.8.2 pytz-2023.3 scipy-1.10.1 setuptools-67.7.2 six-1.16.0 toolz-0.12.0 typing-extensions-4.5.0 tzdata-2023.3 xarray-2023.1.0 xarray-einstats-0.5.1 zipp-3.15.0
WARNING: You are using pip version 22.0.4; however, version 23.1.2 is available.
You should consider upgrading via the '/home/runner/work/pymc-experimental/pymc-experimental/test-sdist/venv-sdist/bin/python -m pip install --upgrade pip' command.
Checking import and version number (on release)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/runner/work/pymc-experimental/pymc-experimental/test-sdist/venv-sdist/lib/python3.8/site-packages/pymc_experimental/__init__.py", line 16, in <module>
    import versioneer
ModuleNotFoundError: No module named 'versioneer'

michaelraczycki:
yes, I need to modify the action to install versioneer before trying to run setup. I'm right now also setting up the docker env to be able to run actions locally without pushing to repo to see the result, that's why it's taking so long but I guess might be useful for the future if we'd like to introduce more workflows

ricardoV94:
The API can't still be accessed in the RTD preview, is that expected? https://pymcio--162.org.readthedocs.build/projects/experimental/en/162/

ricardoV94:
What was the reason we didn't go with the simpler `pymc-marketing` strategy where we hard-code the version number in a txt file?

michaelraczycki:
that was my (maybe unfortunate) decision. I am a big fan of using the same solutions in all related libraries, so I wanted to implement the same solution that was used in pymc, since pymc-experimental is an extension of it

ricardoV94:
I have the hunch the pymc / pytensor solution brings more (maintenance) problems than it solves. So maybe we should just go back and try the simple solution?

docs/index.rst
```diff
@@ -41,4 +41,4 @@ See the `GitHub contributor page <https://github.com/pymc-devs/pymc-experimental
 .. toctree::
    :hidden:
 
-   api_reference
+   api_reference <https://www.pymc.io/projects/experimental/en/latest/api_reference.html>
```
OriolAbril, line 44:
```suggestion
   api_reference
```

This should not be changed. We want this to generate a link to the api reference of the version for which the docs are being built. Changing that we'd be hardcoding the link to always point to latest while at the same time excluding the actual api_reference doc from the toctree. Trying to see why the docs are broken in latest, but this would break them even more without fixing anything.

OriolAbril (changes requested):

docs/index.rst
OriolAbril: To fix docs:

* move `conf.py` back to `docs` folder, not `docs/source` folder.
* delete docs/source folder
* update readthedocs yaml and makefile to have the right path

michaelraczycki:
In the end we decided to go with other versioning system, with version.py and version.txt

PR #165, opened 2023-05-10
Merging BayesianEstimator into ModelBuilder

michaelraczycki:
Complete merge of BayesianEstimator and it's functionalities into ModelBuilder:

Implementation details:
-splitting create_sample_input into 3 new functions:
    *generate_model_data - a method that takes over data generation/pre-formatting 
    *default_sampler_config - property function that allows access to default sampler config 
    *default_model_config - property function that allows access to default model config

Split as above, allows for simplification of ModelBuilder initialisation in the default mode, allowing user easy access to the data structures, while making all parameters optional maintains full customisation option.

All model-related functions now take X and y as input, to mimic scikit-learn model input parameters. That was intended to allow usage of sklearn Transformers and Pipelines with PyMC models. 
For now, data can be provided still on each of the steps that it could have been provided in the past (class init, build_model function). This allows for a live test period, in which it will be determined whether the combined data provisioning system will stay, or we'll move into sklearn-like way.
Fit now takes an optional parameter 'predictor_names' which allows custom naming of the predictors in fit_data, in case the predictors were given in 2d array, in case of providing parameters in pandas Dataframe, the predictor names will be taken from the data frame itself.


michaelraczycki:
I decided to adapt some of the favourite solution ideas that were introduced with BayesianEstimator class, also I decided to not fully skip their implementation to avoid possible runtime issues that I've encountered in last development.

michaelraczycki:
the pytest exception needed to be added because of numpy 1.24. introducing deprecation warning for bool8, which is used in bokeh library, was causing all test to fail on collection for any envs with the latest numpy

pymc_experimental/model_builder.py
```diff
@@ -193,6 +256,34 @@ def sample_model(self, **kwargs):
...
+        RuntimeError
+            If no InferenceData object is provided.
+
+        Returns
+        -------
+        None
+
+        Examples
+        --------
+        >>> import arviz as az
```
twiecki, line 278: These are pretty weird examples, more like tests. This should give the user info on how to use the method to achieve something.

pymc_experimental/model_builder.py
```diff
@@ -205,22 +296,33 @@ def set_idata_attrs(self, idata=None):
...
-        >>>     ...
-        >>> data, model_config, sampler_config = LinearModel.create_sample_input()
-        >>> model = LinearModel(model_config, sampler_config)
-        >>> idata = model.fit(data)
-        >>> name = './mymodel.nc'
-        >>> model.save(name)
+        This method is meant to be overridden and implemented by subclasses.
+        It should not be called directly on the base abstract class or its instances.
+
+        >>> class MyModel(ModelBuilder):
```
twiecki, line 320: This is a good example.

pymc_experimental/model_builder.py
michaelraczycki: Doctests provide an option to perform additional checks based on the docstrings, they might look a bit odd, but you're right, this one is not the best. I'll try to refactor it. Maybe putting some comment on the last line that is actually both doctest and the user manual (in this case 280) would help?

pymc_experimental/model_builder.py
```diff
@@ -275,10 +275,10 @@ def set_idata_attrs(self, idata=None):
 
         Examples
         --------
-        >>> import arviz as az
-        >>> idata = az.InferenceData()
-        >>> self.set_idata_attrs(idata=idata)
-        >>> assert "id" in idata.attrs
+        >>> model = YourModel(ModelBuilder)
```
twiecki, line 278: Maybe keep it consistent between `MyModel`, `YourModel`.

michaelraczycki: My mistake, didn't notice that one. It's fixed now.

pymc_experimental/bayesian_estimator_linearmodel.py
```diff
@@ -60,6 +60,7 @@ class BayesianEstimator(ModelBuilder):
 
     def __init__(
         self,
+        data: Union[np.ndarray, pd.DataFrame, pd.Series] = None,
```
twiecki, line 63: Why are you adding this back in?

michaelraczycki: As discussed, I want to experiment with adding a no-parameter initialization that will act as a sandbox for the new users, but keeping the data in the constructor helps to reduce the number of adaptations before going into implementing model builder in other classes. Also it has low impact, and should be easy to remove later if we don't like it

michaelraczycki:
please do not squash merge it, the commits were split on purpose to make it reversible in case it's needed

michaelraczycki:
@twiecki @ricardoV94 any suggestions to the current state?

.pylintrc
```diff
@@ -46,7 +46,6 @@ enable=import-self,
        used-before-assignment,
        cell-var-from-loop,
        global-variable-undefined,
-       dangerous-default-value,
```
ricardoV94, line 49: Didn't have time to check the rest yet, but this doesn't seem right

michaelraczycki: There's a problem with pymc-marketing and pymc-expental, as they don't use the same linting rules. So It's something I forgot to 'undo' after installing locally pymc-experimental from my local repo.

PR #168, opened 2023-05-17
Implement observe and do transformations

ricardoV94:
```python
import pymc as pm

with pm.Model() as m:
    x = pm.Normal("x", 0, 1)
    y = pm.Normal("y", x, 1)
    z = pm.Normal("z", y + x, 1)

# Dummy posterior, same as calling `pm.sample`
idata_m = az.from_dict({rv.name: [pm.draw(rv, draws=500)] for rv in [x, y, z]})

# Replace `y` by a constant `100.0`
m_do = do(m, {y: 100.0})
with m_do:
    idata_do = pm.sample_posterior_predictive(idata_m, var_names="z")
```

### DIFF
--- a/pymc_experimental/model_transform/conditioning.py
+++ b/pymc_experimental/model_transform/conditioning.py
@@ -4,6 +4,7 @@ from pymc import Model
 from pymc.pytensorf import _replace_vars_in_graphs
 from pytensor.tensor import TensorVariable
 
+from pymc_experimental.model_transform.basic import prune_vars_detached_from_observed
 from pymc_experimental.utils.model_fgraph import (
     ModelDeterministic,
     ModelFreeRV,
@@ -113,7 +114,9 @@ def replace_vars_in_graphs(graphs: Sequence[TensorVariable], replacements) -> Li
     return replaced_graphs
 
 
-def do(model: Model, vars_to_interventions: Dict[Union["str", TensorVariable], Any]) -> Model:
+def do(
+    model: Model, vars_to_interventions: Dict[Union["str", TensorVariable], Any], prune_vars=False
+) -> Model:
     """Replace model variables by intervention variables.
 
     Intervention variables will either show up as `Data` or `Deterministics` in the new model,
@@ -126,6 +129,9 @@ def do(model: Model, vars_to_interventions: Dict[Union["str", TensorVariable], A
         Dictionary that maps model variables (or names) to intervention expressions.
         Intervention expressions must have a shape and data type that is compatible
         with the original model variable.
+    prune_vars: bool, defaults to False
+        Whether to prune model variables that are not connected to any observed variables,
+        after the interventions.
 
     Returns
     -------
@@ -196,4 +202,7 @@ def do(model: Model, vars_to_interventions: Dict[Union["str", TensorVariable], A
     # Replace variables by interventions
     toposort_replace(fgraph, tuple(replacements.items()))
 
-    return model_from_fgraph(fgraph)
+    model = model_from_fgraph(fgraph)
+    if prune_vars:
+        return prune_vars_detached_from_observed(model)
+    return model
--- a/pymc_experimental/model_transform/conditioning.py
+++ b/pymc_experimental/model_transform/conditioning.py
@@ -1,0 +1,199 @@
+from typing import Any, Dict, List, Sequence, Union
+
+from pymc import Model
+from pymc.pytensorf import _replace_vars_in_graphs
+from pytensor.tensor import TensorVariable
+
+from pymc_experimental.utils.model_fgraph import (
+    ModelDeterministic,
+    ModelFreeRV,
+    extract_dims,
+    fgraph_from_model,
+    model_deterministic,
+    model_from_fgraph,
+    model_named,
+    model_observed_rv,
+    toposort_replace,
+)
+from pymc_experimental.utils.pytensorf import rvs_in_graph
+
+
+def observe(model: Model, vars_to_observations: Dict[Union["str", TensorVariable], Any]) -> Model:
+    """Convert free RVs or Deterministics to observed RVs.
+
+    Parameters
+    ----------
+    model: PyMC Model
+    vars_to_observations: Dict of variable or name to TensorLike
+        Dictionary that maps model variables (or names) to observed values.
+        Observed values must have a shape and data type that is compatible
+        with the original model variable.
+
+    Returns
+    -------
+    new_model: PyMC model
+        A distinct PyMC model with the relevant variables observed.
+        All remaining variables are cloned and can be retrieved via `new_model["var_name"]`.
+
+    Examples
+    --------
+
+    .. code-block:: python
+
+        import pymc as pm
+        from pymc_experimental.model_transform.conditioning import observe
+
+        with pm.Model() as m:
+            x = pm.Normal("x")
+            y = pm.Normal("y", x)
+            z = pm.Normal("z", y)
+
+        m_new = observe(m, {y: 0.5})
+
+    Deterministic variables can also be observed.
+    This relies on PyMC ability to infer the logp of the underlying expression
+
+    .. code-block:: python
+
+        import pymc as pm
+        from pymc_experimental.model_transform.conditioning import observe
+
+        with pm.Model() as m:
+            x = pm.Normal("x")
+            y = pm.Normal.dist(x, shape=(5,))
+            y_censored = pm.Deterministic("y_censored", pm.math.clip(y, -1, 1))
+
+        new_m = observe(m, {y_censored: [0.9, 0.5, 0.3, 1, 1]})
+
+
+    """
+    vars_to_observations = {
+        model[var] if isinstance(var, str) else var: obs
+        for var, obs in vars_to_observations.items()
+    }
+
+    valid_model_vars = set(model.free_RVs + model.deterministics)
+    if any(var not in valid_model_vars for var in vars_to_observations):
+        raise ValueError(f"At least one var is not a free variable or deterministic in the model")
+
+    fgraph, memo = fgraph_from_model(model)
+
+    replacements = {}
+    for var, obs in vars_to_observations.items():
+        model_var = memo[var]
+
+        # Just a sanity check
+        assert isinstance(model_var.owner.op, (ModelFreeRV, ModelDeterministic))
+        assert model_var in fgraph.variables
+
+        var = model_var.owner.inputs[0]
+        var.name = model_var.name
+        dims = extract_dims(model_var)
+        model_obs_rv = model_observed_rv(var, var.type.filter_variable(obs), *dims)
+        replacements[model_var] = model_obs_rv
+
+    toposort_replace(fgraph, tuple(replacements.items()))
+
+    return model_from_fgraph(fgraph)
+
+
+def replace_vars_in_graphs(graphs: Sequence[TensorVariable], replacements) -> List[TensorVariable]:
+    def replacement_fn(var, inner_replacements):
+        if var in replacements:
+            inner_replacements[var] = replacements[var]
+
+        # Handle root inputs as those will never be passed to the replacement_fn
+        for inp in var.owner.inputs:
+            if inp.owner is None and inp in replacements:
+                inner_replacements[inp] = replacements[inp]
+
+        return [var]
+
+    replaced_graphs, _ = _replace_vars_in_graphs(graphs=graphs, replacement_fn=replacement_fn)
+    return replaced_graphs
+
+
+def do(model: Model, vars_to_interventions: Dict[Union["str", TensorVariable], Any]) -> Model:
+    """Replace model variables by intervention variables.
+
+    Intervention variables will either show up as `Data` or `Deterministics` in the new model,
+    depending on whether they depend on other RandomVariables or not.
+
+    Parameters
+    ----------
+    model: PyMC Model
+    vars_to_interventions: Dict of variable or name to TensorLike
+        Dictionary that maps model variables (or names) to intervention expressions.
+        Intervention expressions must have a shape and data type that is compatible
+        with the original model variable.
+
+    Returns
+    -------
+    new_model: PyMC model
+        A distinct PyMC model with the relevant variables replaced by the intervention expressions.
+        All remaining variables are cloned and can be retrieved via `new_model["var_name"]`.
+
+    Examples
+    --------
+
+    .. code-block:: python
+
+        import pymc as pm
+        from pymc_experimental.model_transform.conditioning import do
+
+        with pm.Model() as m:
+            x = pm.Normal("x", 0, 1)
+            y = pm.Normal("y", x, 1)
+            z = pm.Normal("z", y + x, 1)
+
+        # Dummy posterior, same as calling `pm.sample`
+        idata_m = az.from_dict({rv.name: [pm.draw(rv, draws=500)] for rv in [x, y, z]})
+
+        # Replace `y` by a constant `100.0`
+        m_do = do(m, {y: 100.0})
+        with m_do:
+            idata_do = pm.sample_posterior_predictive(idata_m, var_names="z")
+
+    """
+    do_mapping = {}
+    for var, obs in vars_to_interventions.items():
+        if isinstance(var, str):
+            var = model[var]
+        try:
+            do_mapping[var] = var.type.filter_variable(obs)
+        except TypeError as err:
+            raise TypeError(
+                "Incompatible replacement type. Make sure the shape and datatype of the interventions match the original variables"
+            ) from err
+
+    if any(var not in model.named_vars.values() for var in do_mapping):
+        raise ValueError(f"At least one var is not a named variable in the model")
+
+    fgraph, memo = fgraph_from_model(model, inlined_views=True)
+
+    # We need the interventions defined in terms of the IR fgraph representation,
+    # In case they reference other variables in the model
+    ir_interventions = replace_vars_in_graphs(list(do_mapping.values()), replacements=memo)
+
+    replacements = {}
+    for var, intervention in zip(do_mapping, ir_interventions):
+        model_var = memo[var]
+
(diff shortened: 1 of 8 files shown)


pymc_experimental/model_transform/conditioning.py
```diff
@@ -0,0 +1,103 @@
...
+    fgraph_from_model,
+    model_from_fgraph,
+    model_named,
+    model_observed_rv,
+    toposort_replace,
+)
+
+
+def observe(model: Model, vars_to_observations: Dict[Union["str"], Any]) -> Model:
+    """Convert free RVS to observed RVs.
```
twiecki, line 19:
```suggestion
    """Convert free RVs to observed RVs.
```

pymc_experimental/tests/model_transform/test_conditioning.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_do():
+    with pm.Model() as m_old:
+        x = pm.Normal("x", 0, 1e-3)
+        y = pm.Normal("y", x, 1e-3)
+        z = pm.Normal("z", y + x, 1e-3)
+
+    assert -5 < pm.draw(z) < 5
+
+    m_new = do(m_old, {y: x + 100})
```
twiecki, line 49: 🔥

pymc_experimental/model_transform/conditioning.py
```diff
@@ -0,0 +1,103 @@
...
+
+    This function returns a new model.
+    """
+    vars_to_observations = {
+        model[var] if isinstance(var, str) else var: obs
+        for var, obs in vars_to_observations.items()
+    }
+
+    # Note: Since PyMC can infer logprob expressions we could also allow observing Deterministics
```
lucianopaz, line 28: Since we already are in experimental, do you want to actually try this out?

pymc_experimental/model_transform/conditioning.py
```diff
@@ -0,0 +1,103 @@
...
+    replacements = {}
+    for var, obs in vars_to_observations.items():
+        model_free_rv = memo[var]
+
+        # Just a sanity check
+        assert isinstance(model_free_rv.owner.op, ModelFreeRV)
+        assert model_free_rv in fgraph.variables
+
+        rv, vv, *dims = model_free_rv.owner.inputs
+        model_obs_rv = model_observed_rv(rv, rv.type.filter_variable(obs), *dims)
```
lucianopaz, line 43: Could `obs` have also depended on some nodes from the original model graph? If yes, maybe we should check and replace with memo?

lucianopaz (commented):
This looks fine to me. I’d only add more expansions and examples to the docstrings

pymc_experimental/model_transform/conditioning.py
ricardoV94: Sure :D

ricardoV94:
> Something like unobserve? 

Yeah I thought about it... but I am not sure what an unobserved variable should be. A Deterministic? A FreeRV?

> I don’t think that a do actually has a well defined inverse. People could use do again to get the original model back though

Yeah that can't have an inverse because it's just a constant. The user would have to tell us what RV to replaced it with and whether it's free or observed. Which I think should have it's own name? `replace_by_var`?

I think I would focus on just these two for now. See if the approach is even useful and then expand with the rest of the logical subspace.

drbenvincent (commented):
Once you get your new (mutilated) model back from `pm.do`, what's the next step? Calling `pm.sample_posterior_predictive`? What would happen if you call `pm.sample_prior_predictive()`, would you get the same result? It _might_ be useful to include that step as an example in the docstring maybe also a test?

ricardoV94:
> Once you get your new (mutilated) model back from `pm.do`, what's the next step? Calling `pm.sample_posterior_predictive`? What would happen if you call `pm.sample_prior_predictive()`, would you get the same result? It _might_ be useful to include that step as an example in the docstring maybe also a test?

I'll add the example in the docstring. If you call prior predictive you ignore everything you learned about the parameters that are still in the model.

pymc_experimental/model_transform/conditioning.py
```diff
@@ -0,0 +1,179 @@
...
+
+        import pymc as pm
+        from pymc_experimental.model_transform.conditioning import observe
+
+        with pm.Model() as m:
+            x = pm.Normal("x")
+            y = pm.Normal("y", x)
+            z = pm.Normal("z", y)
+
+        m_new = observe(m, {y: 0.5})
```
drbenvincent, line 49: Missing quotes around `y`? I didn't realise you could use a `TensorVariable` type as a key in a dictionary

pymc_experimental/model_transform/conditioning.py
```diff
@@ -0,0 +1,179 @@
...
+
+        import pymc as pm
+        from pymc_experimental.model_transform.conditioning import do
+
+        with pm.Model() as m_old:
+            x = pm.Normal("x", 0, 1)
+            y = pm.Normal("y", x, 1)
+            z = pm.Normal("z", y + x, 1)
+
+        m_new = do(m, {y: 100.0})
```
drbenvincent, line 145: Add a line here to indicate the common case of calling `pm.sample_posterior_predictive(m_new, idata)`. Presumably will need to update example so `idata` is available.

pymc_experimental/tests/model_transform/test_conditioning.py
```diff
@@ -0,0 +1,91 @@
...
+        x = pm.Normal("x")
+        pm.Censored("y_censored", pm.Normal.dist(x), lower=-1, upper=1, observed=y_censored_obs)
+
+    np.testing.assert_allclose(
+        m_new.compile_logp()({"x": 0.9}),
+        m_ref.compile_logp()({"x": 0.9}),
+    )
+
+
+def test_do():
```
drbenvincent, line 61: Recommend adding a test case which also calls `pm.sample_posterior_predictive`

pymc_experimental/model_transform/conditioning.py
ricardoV94: You can ;). Both ways are accepted by the function

pymc_experimental/tests/model_transform/test_conditioning.py
ricardoV94: Added. Actually there was some issue with the way we were doing things. Should be solved by https://github.com/pymc-devs/pymc/pull/6741

pymc_experimental/model_transform/conditioning.py
ricardoV94: Added a test for Minibatch observations. This should cover the whole range of "allowed" observed variables.

ricardoV94:
Tests are passing

drbenvincent (commented):
As far as I understand, at the moment the `do` operator only operates in the situation where you want to replace a random variable with observed data. This is fine, but this is only one use-case. **In this case, a user would implement the do operator with `pm.do()` then `pm.sample_posterior_predictive`**

For example, if you take the potential outcomes approach to confounder adjustment OR take the SCM approach with do-calculus to calculate an adjustment set (with backdoor criterion), then in both cases you basically end up with a linear regression where you enter in the variables you decide to condition upon. In this situation, these are defined in the model as `pm.MutableData`. **In this case, a user should implement the do operator with `pm.set_data()` then `pm.sample_posterior_predictive`**

From an implementation point of view, I can see that we might want different functions to implement these different things (ie. replace observed with observed vs replace RV with observed). But from a user-facing point of view, they could see it as frustrating that they have to remember which they have to use (`pm.set_data` or `pm.do`) when in both cases they want to "do".

drbenvincent:
* It could be better if the newly injected `ConstantData` node could inherit the dims from the RV that it replaces

* I'd also vote for fully removing the parent nodes from any nodes that have been intervened on to make the graphviz simpler. See pics below

Example before
Screenshot 2023-06-01 at 12 09 31

Example after
Screenshot 2023-06-01 at 12 09 45

ricardoV94:
> As far as I understand, at the moment the do operator only operates in the situation where you want to replace a random variable with observed data. 

This is not true. You can replace the variables by anything you want. Check the test where we replace it by an expression with a shared variable that acts as a switch:

https://github.com/pymc-devs/pymc-experimental/blob/ff32a66d1e2eaf96493cc7c29e99da029320e9f0/pymc_experimental/tests/model_transform/test_conditioning.py#L108-L110

I just mentioned you didn't have to, but you can certainly replace constant data by other constant data if you want to use the same method

ricardoV94:
> It could be better if the newly injected ConstantData node could inherit the dims from the RV that it replaces

That was actually supposed to work. Gonna try and fix it

drbenvincent:
> I just mentioned you didn't have to, but you can certainly replace constant data by other constant data if you want to use the same method

So when I try to use do to replace mutable data then I run into issues

```python
with pm.Model(coords_mutable=coords) as model_ancova:
    # data
    z = pm.MutableData("z", df_obs["z"], dims="obs_id")
    c = pm.MutableData("c", df_obs["c"], dims="obs_id")
    y = pm.MutableData("y", df_obs["y"], dims="obs_id")
    # priors
    beta0 = pm.Normal("beta0")
    beta_tau = pm.Normal("beta_tau")
    beta2 = pm.Normal("beta2")
    sigma = pm.HalfNormal("sigma", 1)
    # likelihood
    mu = pm.Deterministic("mu", beta0 + beta_tau * z + beta2 * c, dims="obs_id")
    pm.Normal("y_obs", mu=mu, sigma=sigma, observed=y, dims="obs_id")
    idata_ancova = pm.sample()

N = len(df_obs)

# P(y|c,do(z=0))
model_control = do(model_ancova, {"z": np.zeros(N, dtype='int32')})
with model_control:
    idata_z0 = pm.sample_posterior_predictive(idata_ancova, predictions=True, var_names=["mu"])
```
Results in this error
```
ValueError                                Traceback (most recent call last)
Cell In[14], line 4
      1 N = len(df_obs)
      3 # P(y|c,do(z=0))
----> 4 model_control = do(model_ancova, {"z": np.zeros(N, dtype='int32')})
      5 with model_control:
      6     idata_z0 = pm.sample_posterior_predictive(idata_ancova, predictions=True, var_names=["mu"])

File ~/git/pymc-experimental/pymc_experimental/model_transform/conditioning.py:160, in do(model, vars_to_interventions)
    157     do_mapping[var] = var.type.filter_variable(obs)
    159 if any(var not in (model.basic_RVs + model.deterministics) for var in do_mapping):
--> 160     raise ValueError(f"At least one var is not a variable or deterministic in the model")
    162 fgraph, memo = fgraph_from_model(model)
    164 # We need the interventions defined in terms of the IR fgraph representation,
    165 # In case they reference other variables in the model

ValueError: At least one var is not a variable or deterministic in the model
```

drbenvincent:
Oh, you have to pass in MutableData object, not the raw data

ricardoV94:
Hmm so we actually had a check that the var being replaced had to be a Deterministic or RV... we could relax that to being Constant/Mutable data as well, if you think it's useful

drbenvincent:
Still no joy with this:

```python
with model_ancova:
    do_z0 = pm.MutableData("do_z0", np.zeros(N, dtype='int32'))

model_control = do(model_ancova, {"z": do_z0})
idata_z0 = pm.sample_posterior_predictive(idata_ancova, predictions=True, var_names=["mu"])
```

Error
```
ValueError                                Traceback (most recent call last)
Cell In[16], line 4
      1 with model_ancova:
      2     do_z0 = pm.MutableData("do_z0", np.zeros(N, dtype='int32'))
----> 4 model_control = do(model_ancova, {"z": do_z0})
      5 idata_z0 = pm.sample_posterior_predictive(idata_ancova, predictions=True, var_names=["mu"])

File ~/git/pymc-experimental/pymc_experimental/model_transform/conditioning.py:160, in do(model, vars_to_interventions)
    157     do_mapping[var] = var.type.filter_variable(obs)
    159 if any(var not in (model.basic_RVs + model.deterministics) for var in do_mapping):
--> 160     raise ValueError(f"At least one var is not a variable or deterministic in the model")
    162 fgraph, memo = fgraph_from_model(model)
    164 # We need the interventions defined in terms of the IR fgraph representation,
    165 # In case they reference other variables in the model

ValueError: At least one var is not a variable or deterministic in the model
```

drbenvincent:
> Hmm so we actually had a check that the var being replaced had to be a Deterministic or RV... we could relax that to being Constant/Mutable data as well, if you think it's useful

From my point of view, relaxing that would make it better for the user. Being able to just say "I want to go in an intervene on a node, so I'll use the `pm.do`" would be good, regardless of whether that node is RV or Determinist or Constant/Mutable data.

As a bonus, it would be pretty awesome if they could just pass in numerical data and that automatically be turned into a MutableData or ConstantData object. But not sure how hard that is. Maybe they forgot control over naming the replacement variable and it keeps the original name?

ricardoV94:
> As a bonus, it would be pretty awesome if they could just pass in numerical data and that automatically be turned into a MutableData or ConstantData object. But not sure how hard that is. Maybe they forgot control over naming the replacement variable and it keeps the original name?

If you pass constants it becomes `ConstantData`. You just have to pass constants with valid types (can't pass an int64 to an int32 variable)

drbenvincent:
Sounds reasonable

drbenvincent:
Actually, it would need the new model context block there I believe:
```python
# P(y|c,do(z=0))
model_control = do(model_ancova, {"z": np.zeros(N, dtype='int32')})
with model_control:
    idata_z0 = pm.sample_posterior_predictive(idata_ancova, predictions=True, var_names=["mu"])
```

ricardoV94:
@drbenvincent dims should now be working. I updated it so that the interventions in the model show either as `Data` or `Deterministic` depending on whether the replacement is Constant or depends on other model variables

drbenvincent (commented):
As far as I can tell the newly injected data now does take on the coords of the item it replaces.

I can also replace a RV with data with 
```python
model_control = do(model_scm, {"z": np.zeros(N, dtype='int32')})
with model_control:
    idata_scm_z0 = pm.sample_posterior_predictive(idata_scm, predictions=True, var_names=["mu"])
```

The only thing I'm not sure about at the moment is whether we remove all ancestors of nodes that have been intervened on. My initial impulse is that we should remove them because the resulting graphviz would me much simpler. But I can see a situation where an ancestor of an intervened variable is also the parent of some other node in the graph and it should still be included. 

So perhaps an alternative would be to remove all ancestors of nodes that have been intervened on AND are orphans?

drbenvincent (commented):
As far as I can tell the newly injected data now does take on the coords of the item it replaces.

I can also replace a RV with data with 
```python
model_control = do(model_scm, {"z": np.zeros(N, dtype='int32')})
with model_control:
    idata_scm_z0 = pm.sample_posterior_predictive(idata_scm, predictions=True, var_names=["mu"])
```

The only thing I'm not sure about at the moment is whether we remove all ancestors of nodes that have been intervened on. My initial impulse is that we should remove them because the resulting graphviz would me much simpler. But I can see a situation where an ancestor of an intervened variable is also the parent of some other node in the graph and it should still be included. 

So perhaps an alternative would be to remove all ancestors of nodes that have been intervened on AND are orphans?

ricardoV94:
> So perhaps an alternative would be to remove all ancestors of nodes that have been intervened on AND are orphans?

That is not a trivial graph question (or at least it didn't look for me). I just pushed a commit that does something more simple: prune any variables that are not related to any likelihood term. I think this is easy enough to reason about and should cover most cases?

drbenvincent:
I tried it out.
BEFORE
Screenshot 2023-06-02 at 13 45 29
AFTER 
```python model_control = do(model_scm, {"z": np.zeros(N, dtype='int32')}, prune_vars=True)```
Screenshot 2023-06-02 at 13 46 03
Only thing that looks like it might be an issue is the status of `y_data` has changed.

ricardoV94:
>  Only thing that looks like it might be an issue is the status of y_data has changed.

Looks like a bug. Slowly but surely we're getting there xD

drbenvincent:
> Does it also change status when `prune_vars=False`?

No. Only when `prune_vars=True`

twiecki:
Given that we've now tested this quite a bit, shouldn't we just put this into pymc proper?

ricardoV94:
> Given that we've now tested this quite a bit, shouldn't we just put this into pymc proper?

No.
The underlying functionality (model->fgraph) was changed like 20x in the course of this PR, and it really helps to be able to break it and start from scratch without worries about breaking user compat.

ricardoV94:
> No. Only when `prune_vars=True`

I think it's fixed now!

drbenvincent:
> I think it's fixed now!

Yes - certainly for the examples I was looking at the mutilated graph with `prune_vars=True` looks good.

drbenvincent:
Let me know if there's anything else you want me to test. Otherwise I'm happy to approve

ricardoV94:
> Let me know if there's anything else you want me to test. Otherwise I'm happy to approve 

If you think this covers all the use cases for the blogpost we can merge it (need to rebase once more first).

drbenvincent (approved):
I believe we now have the functionality needed for a blog post. Ideally we get a bit more road testing, and eyeballs from other people, to catch any issues. But moving into the pymc repo reasonably soon would be good.

ricardoV94:
> But moving into the pymc repo reasonably soon would be good.

I'll be honest, I don't want to do that super soon. Not because of the `do` which is pretty self-contained (although I wouldn't be surprised if we need more tweaks once people try this out in real cases), but because of the fgraph stuff.

I don't get the rush either

drbenvincent:
There's some anticipation because it's cool and would be good to get out there. But I agree, if it relies on stuff that is still experimental then there's no need to rush. Getting a blog post out there which calls on pymc-experimental would sate the desire I think.

ricardoV94:
Tests should now pass again. Need a green review to merge. We can cut a release after

twiecki:
Also, congrats @ricardoV94, this is majorly cool new functionality.

PR #174, opened 2023-05-29
Merge statespace module from http://github.com/jessegrabowski/pymc_statespace

jessegrabowski:
This is the promised statespace PR. It's a WIP, for several reasons, but it already represents a MVP that can be used. I'll try to explain what I was going for with everything here. I expect that a lot needs to be changed, though, and I look forward to working hard to improve it!

# Overall design goal

Statespace models are a pair of linear equations of the form:

$$\begin{align}x_{t+1} &= T_tx_t + c_t + R_t\epsilon_t, &\quad \epsilon_t &\sim N(0, Q_t) \\
y_t &= Z_tx_t + d_t + \eta_t &\quad \eta_t &\sim N(0, H_t) \end{align}$$

Where the first equation is called the "transition equation" and the second equation is called the "observation equation". The state vector $x_t$ contain both observed and unobserved states, while the observation vector $y_t$ contains only the states observed in data.

Since everything is linear and gaussian, the posterior distribution over the observed states is just a multivariate normal. Actually it can be shown that this is a special case of a guassian process. But, since we might have a lot of states or a lot of times, we can compute the posterior mean and covariance more efficiently via recursion. This is the Kalman Filter.

## Sidebar 1: Why not just scan?

Given the advancements in scan lately, it's fair to ask why bother with filtering at all, and just let PyMC automatically infer the logp of these two equations. This might be possible, and it's something I'd like to explore. It will also let us push out beyond gaussian errors, which would be great. There was already a user on the forum looking for poisson distributed observations, for example. 

For now it's not possible, because $Q$ might not be fully rank -- that is, not all states in $x_t$ need to be stochastic. We could solve this with clever slicing, though. A second problem is that the quantity $R_t\epsilon_t$ is not measurable. For gaussian errors this isn't a problem, because we can just fold it into the covariance matrix, and write $x_{t+1} = T_tx_t + \epsilon_t, \quad \epsilon_t \sim N(0, R_tQ_tR_t^T)$, but this identity doesn't hold generally.

## Sidebar over

Basically, this PR has the following goals:

1. Abstract away the matrices and allow users to just focus on model parameters.
2. Offer computationally efficient solutions for computing the logp of these models
3. Allow the models to be quickly sampled
4. Alleviate headaches associated with post-estimation tasks, especially forecasting

#  Modules in this PR

I will briefly introduce what I've done in this PR, and try to justify my choices. I am confident they can be improved.

## Core

The core module is responsible for representing arbitrary statespace models. There are two files, `representation.py` and `statespace.py`. 

### representation.py

This is the lowest-level object in the module. It is responsible for initializing and storing the matrices c, d, T, Z, R, H, Q, along with initial state and covariance x0 and P0. The module overloads `__getitem__` and `__setitem__` to allow the user to slice into matrices by name, for example `mod['transition', 0, 0]` gets the `[0, 0]` position of the `T` matrix.

In addition, it also has a bunch of checks and logic to handle time-varying matrices. If a user wants the state-space matrices to vary over time, it is necessary to duplicate them and store the whole stack to scan over later. If the model is not time varying, it automatically slices around the time dimension so the user is never confronted with it.

Users should never have to touch this, it's all just low-level machinery.

### statespace.py

This is the base class for all statespace models. It's responsible for combining the statespace matrices with a kalman filter to make a logp graph. This is accomplished by the `gather_required_random_variables`, `update`, and `build_statespace_graph` methods. Only `update` will vary between models, and needs to be implemented. `update` is responsible for taking a flat vector of parameters and shuttling them to the correct places in the statespace matrices. 

The property `param_names` also needs to be set for each model. This defines the names of the parameters that `gather_required_random_variables` will look for in the pymc model. It also defines the order of the flat parameter vector that will be passed into `update`. 

## Filters

Filters holds all the implementations for the Kalman Filter. 

### distributions.py

Work in progress. Eventually, this should implement a PyMC distribution wrapper around the kalman filter, so that we can directly sample from it. There's just a lot of wrinkles to iron out.

### kalman_filter.py

The actual kalman filters. Currently there are 5 implemented: standard, univariate, single time series, cholesky, and steady state. These need to be bench marked against each other. Single time series should be used when there is only a single observed state (e.g. ARIMA), otherwise use standard. Cholesky is supposd to be faster, but in my limited testing it's not. Not sure why. In principle the univariate filter is the most robust, but it has a scan in a scan, so it's quite slow. I haven't benchmarked it in JAX, though.

### kalman_smoother.py

Pytensor implementation of the Kalman smoother. Good for hidden state inference, but split out because not all users will need it. It's a post-estimation thing.

### numpy_filter.py

This is a re-implementation of the cholesky kalman filter in pure numpy. Potentially for use in `distribution.py`. I'm not sure what `pm.draw` should return when called on a statespace distribution, so it's there as an option. 

### utilities.py

Shared functions between modules. Currently just holds a helper function to sort scan inputs into sequence and non-sequence. This is needed because if matrices are time-varying, they are sequences, otherwise they are non-sequences. I guess I could always make them sequences and just copy the matrix a bunch of times if its not time-varying, but this is more memory efficient (cope because I already sunk a lot of time into doing it this way?)

## Models

This module will hold actual implementations of state space models that users can call. Right now they are fully contained models, following the setup of statsmodels, but I could imagine a better, more modular API. Right now I have VARMA, ARMIA, and local level. 

There's also a utility file for shared functions, right now it's a little function that's used in the `update` function to keep track of slicing up the flat parameter vector.

## Utils

A hodge-podge of stuff. 

### numba_linalg.py

This holds numba implements linear algebra routines with no overload. Currently it's just `scipy.linalg.block_diag`. 

### pytensor_scipy.py

This holds a pytensor `Op`s that should be split off and pulled into `pytensor.tensor.slinalg` for solving the Discrete Algebraic Riccati equation. This is currently used in the SteadyStateFilter, and can save a lot of time by pre-computing and re-using a single matrix inverse for all time-steps in the kalman filter. It doesn't have a jaxified version, though, so it's not actually that useful right now. Solving AREs is useful in general, though.

### simulation.py

This holds numbafied routines for posterior predictive simulation. Since the current implementation returns a `pm.Potential` for the logp, it's not possible to use the usual posterior predictive sampling machinery, so I resorted to this. I hope it can be removed in the future. It contains separate functions for conditional and unconditional simulation. 

Unconditional simulation just applies the observation and state transition equations to an initial state. This is useful for computing theoretical moments of the system, and also for forecasting. 

Conditional simulation draws statespace matrices and runs the data through the kalman filter (and, optionally, smoother). This is useful for hidden state inference and missing data interpolation.

# Summary and To Do

Basically that's it. There's still a lot to do, but this is at least a start. I hope it can be useful to the community, and we can get it to be a super fast, reliable alternative to the statsmodels statespace module.

Here's a quick, non-exhaustive list of to-dos:

1. Get rid of the `pm.potential` term via a distribution wrapper around the kalman filter
2. Write numba overloads for relevant linear algebra operations so we can use nutpie
3. Consider a more modular API. Users should, in principle, be able to combine statespace models via block-diagonal concatenation. Then we could have a "seasonal" model, "ARIMA" model, local level model, deterministic trend model, etc., and users could combine them as `mod = seasonal_part + trend_part + arima_part`.
4. Add better support for forcasting
5. Add support for computing theoritical moments 
6. Add support for computing and plotting IRFs
7. Add parameter transformations relevant to time series models 
8. Plotting functions?
9. Support for time-series model comparison?


pymc_experimental/statespace/core/statespace.py
```diff
@@ -0,0 +1,441 @@
...
+
+        self.kalman_filter = FILTER_FACTORY[filter_type.lower()]()
+        self.kalman_smoother = KalmanSmoother()
+
+        if verbose:
+            # TODO: Ask Ricardo how to log this to console without a print
+            # print(
+            #     "Model successfully initialized! The following parameters should be assigned priors inside a PyMC "
+            #     f'model block: {", ".join(self.param_names)}'
+            # )
```
ricardoV94, line 83: You can get a logger that is a child of the `pymc` one and use that

```python
_log = logging.getLogger("pymc.experimental.statespace")

...
    _log.info(...)
```

pymc_experimental/statespace/core/statespace.py
```diff
@@ -0,0 +1,441 @@
...
+        with pymc_model:
+            for param_name in self.param_names:
+                for param in pymc_model.rvs_to_values:
+                    if param.name == param_name:
+                        theta.append(param.ravel())
+                        found_params.append(param.name)
+                for param in pymc_model.deterministics:
+                    if param.name == param_name:
+                        theta.append(param.ravel())
+                        found_params.append(param.name)
```
ricardoV94, line 139: To get a variable: `param = pymc_model["param_name"]`, then check it's a free/observed RV depending on the constraints `assert `param in pymc_model.basic_RVs`

pymc_experimental/statespace/core/statespace.py
```diff
@@ -0,0 +1,441 @@
...
+        Returns
+        -------
+        simulations: ArrayLike
+            A numpy array of shape (n_simulations x prior_samples, n_timesteps, n_states) with simulated trajectories.
+
+        """
+        validate_filter_arg(filter_output)
+        pymc_model = modelcontext(None)
+        with pymc_model, catch_warnings():
+            simplefilter("ignore", category=UserWarning)
```
ricardoV94, line 232: Why this blank `ignore` in a couple of places?

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -0,0 +1,121 @@
...
+#         if size is not None:
+#             raise NotImplementedError()
+
+#         (filtered_states, predicted_states, filtered_cov,
+#             predicted_cov, log_likelihood) = kalman_filter(data, x0, P0, c, d, T, Z, R, H, Q)
+
+#         return filtered_states, predicted_states, filtered_cov, predicted_cov
+
+
+_kalman_filter_base = _KalmanFilterBaseRV()
```
ricardoV94, line 56: If you implemented a `RandomVariable` you can skip the `SymbolicRandomVariable` / `SymbolicDistribution` altogether. The latter are just for when you want the random method defined directly with PyTensor expressions.

In `Disttribution` just set that to `rv_op` (instead of `rv_type`). Check how `pm.Normal` is defined

pymc_experimental/statespace/utils/numba_linalg.py
```diff
@@ -0,0 +1,17 @@
...
+    n, rows, cols = nd_array.shape
+
+    out = np.zeros((n * rows, n * cols))
+
+    r, c = 0, 0
+    for i, (rr, cc) in enumerate([(rows, cols)] * n):
+        out[r : r + rr, c : c + cc] = nd_array[i]
+        r += rr
+        c += cc
+    return out
```
ricardoV94, line 17: Make sure `numba` is still an optional dependency for users of `pymc_experimental`

pymc_experimental/tests/statespace/test_VARMAX.py
```diff
@@ -0,0 +1,101 @@
...
+import pytensor
+import pytensor.tensor as pt
+import pytest
+import statsmodels.api as sm
+from numpy.testing import assert_allclose
+
+from pymc_experimental.statespace import BayesianVARMAX
+
+floatX = pytensor.config.floatX
+ROOT = Path(__file__).parent.absolute()
```
ricardoV94, line 19: Can you just use the long path `pymc_experimental/tests/.../*.csv`?

ricardoV94 (commented):
Couple of small comments

pymc_experimental/statespace/utils/numba_linalg.py
jessegrabowski: Yeah hopefully this can all just go away by making the filter a distribution.

Do you have any idea how to handle the fact that there are several types of prior/posterior you can sample out of the model: conditional vs unconditional, then the different filter outputs (1-step ahead predictions, filtered, and smoothed), then mean vs covariance? I was playing around with it a bit and it seemed like the random method can't return a tuple -- is that incorrect?

ricardoV94: It has to return a numpy array. Are you trying to do a multi-output Op?

jessegrabowski: The graph for the kalman filter returns 5 values: a sequence of mean vectors and covariance matrices taking into account information up to time t ("kalman predictions"), and the same sequence also incorporating information from time t itself ("kalman filter"), and the log-likelihood of the data at time t, given the kalman predictions. I'll call these mu_p_t, Sigma_p_t, mu_f_t, Sigma_f_t, and ll_p_t.

So there are two sequences of distributions returned: MVN(mu_p_t, Sigma_p_t), and MVN(mu_f_t, Sigma_f_t). This block_diagonal function arises because I use it to sample an entire trajectory all in one go: MVN(concat(mu_p_t), block_diag(Sigma_p_t)), for example. In principle pymc could also compute ll_p_t this way, but it's a bit wasteful because it requires inversion of a bigger covariance matrix. 

But for the random function, either or both of these stacked MVN distributions could be returned, depending on the situation. Or an "unconditional" distribution could be returned, that just iterates the transition and observation equations without incorporating any observations at all.

For the logp function, on the other hand, only the predicted distribution makes sense. My current thought is to split into 4 distributions: KalmanUnconditionalDist, KalmanFilterDist, KalmanPredictionDist, and KalmanSmootherDist, and use a wrapper that returns all 4. Only the PredicitonDist would have an implemented logp function. Thoughts?

Also, is it necessary to re-write the filter in numpy for the random function? Or can I just use the pytensor code I already have?

ricardoV94: > For the logp function, on the other hand, only the predicted distribution makes sense. My current thought is to split into 4 distributions: KalmanUnconditionalDist, KalmanFilterDist, KalmanPredictionDist, and KalmanSmootherDist, and use a wrapper that returns all 4. Only the PredicitonDist would have an implemented logp function. Thoughts?

Sounds good

> Also, is it necessary to re-write the filter in numpy for the random function? Or can I just use the pytensor code I already have?

Should be fine to use PyTensor code, but then you wouldn't create a `RandomVariable`. That's the exact case for `SymbolicDisttribution`s.

pymc_experimental/statespace/filters/distributions.py
jessegrabowski: I'm trying to work towards a random variable; in that spirit I'm re-writing these filtering functions to explicitly return random variables instead of tensors of parameters. I think I am making some mistakes with the updates though:

```python
from pymc_statespace.filters.kalman_filter import MVN_CONST, BaseFilter
from pymc.pytensorf import collect_default_updates

def get_loc_scale_from_distribution(d):
    *_, loc, scale = d.owner.inputs
    return loc, scale

def make_Z_matrix(y):
    is_nan = pt.isnan(y)
    missing_idx = pt.nonzero(is_nan, return_matrix=False)[0]
    obs_idx = pt.nonzero(pt.neq(is_nan, np.array(True)), return_matrix=False)[0]
    
    n_missing = missing_idx.shape[0]
    n_obs = obs_idx.shape[0]
    n = n_missing + n_obs
    
    Z = pt.set_subtensor(pt.zeros((n_obs, n))[pt.arange(n_obs), obs_idx], 1)
    
    return Z, obs_idx

class PyMCFilter(BaseFilter):
    @staticmethod
    def predict(dist, c, T, R, Q) -> MvNormal:
        a, P = get_loc_scale_from_distribution(dist)
        a_hat = T.dot(a) + c
        P_hat = matrix_dot(T, P, T.T) + matrix_dot(R, Q, R.T)

        # Force P_hat to be symmetric
        P_hat = 0.5 * (P_hat + P_hat.T)
        d_predicted = pm.MvNormal.dist(mu=a_hat, cov=P_hat)
        
        return d_predicted, collect_default_updates(inputs=[dist, c, T, R, Q], outputs=[d_predicted])

    def update(self, y, a, P, d, H):
        """
        Conjugate update rule for the mean and covariance matrix, with log-likelihood en passant
        TODO: Verify these equations are correct if there are multiple endogenous variables.
        TODO: Is there a more elegant way to handle nans?
        """
        Z, obs_idx = make_Z_matrix(y)
        v = y[obs_idx] - Z.dot(a) - d

        PZT = P.dot(Z.T)
        F = Z.dot(PZT) + H
        F_inv = pt.linalg.solve(F, self.eye_endog, assume_a="pos", check_finite=False)
        
        K = PZT.dot(F_inv)
        I_KZ = self.eye_states - K.dot(Z)

        a_filtered = a + K.dot(v)
        P_filtered = matrix_dot(I_KZ, P, I_KZ.T) + matrix_dot(K, H, K.T)

        inner_term = matrix_dot(v.T, F_inv, v)
        ll = -0.5 * (MVN_CONST + pt.log(pt.linalg.det(F)) + inner_term).ravel()[0]
        d_filtered = pm.MvNormal.dist(a_filtered, P_filtered)
        
        return (d_filtered, ll), collect_default_updates(inputs=[y, a, P, d, H], outputs=[d_filtered, ll])
    
    def kalman_step(self, *args):
        """
        The timing convention follows [1]. a0 and P0 are taken to be predicted states, so we begin
        with an update step rather than a predict step.
        References
        ----------
        .. [1] Durbin, J., and S. J. Koopman. Time Series Analysis by State Space Methods.
               2nd ed, Oxford University Press, 2012.
        """
#         y, predicted_dist, c, d, T, R, H, Q = self.unpack_args(args)
        y, a, P, c, d, T, R, H, Q = args
        (filtered_dist, ll), filter_updates = self.update(y=y, a=a, P=P, d=d, H=H)
        
        new_predicted_dist, pred_updates = self.predict(dist=filtered_dist, c=c, T=T, R=R, Q=Q)
        new_a, new_P = get_loc_scale_from_distribution(new_predicted_dist)
        outputs = new_a, new_P, filtered_dist, new_predicted_dist, ll
        
        updates = collect_default_updates(inputs=args, outputs=outputs)
        updates.update(filter_updates)
        updates.update(pred_updates)
    
        return outputs, updates
```

Here's a simple test with `pm.CustomDist` that throws errors about the rng having multiple clients, that's why I think there's a problem with the updates:

```python
p, m, r, n = 1, 5, 2, 10

kf = PyMCFilter()
kf.eye_endog = pt.eye(p)
kf.eye_states = pt.eye(m)

with pm.Model() as mod:
    c = pm.Normal('c', size=(m))
    d = pm.Normal('d', size=(p))
    T = pm.Normal('T', size=(m, m))
    R = pm.Normal('R', size=(m, r))
    H = pm.Normal('H', size=(p, p))
    H = H @ H.T
    Q = pm.Normal('Q', size=(r, r))
    Q = Q @ Q.T
    
    a0 = pm.Normal('a0', size=(m, ))
    P0_sigma = pm.Exponential('P0_sigma', 1)
    P0 = pt.eye(m) * P0_sigma
    
    def kf_func_factory(data, a0, P0, c, d, T, R, H, Q, size=None):
        (*_, filtered_dist, predicted_dist, ll), updates = pytensor.scan(kf.kalman_step,
                             sequences=[data],
                             non_sequences=[c, d, T, R, H, Q],
                             outputs_info=[a0, P0, None, None, None])
        
        def kf_filtered(data, c, d, T, R, H, Q, size=None):
            return filtered_dist
        
        def kf_predicted(data, c, d, T, R, H, Q, size=None):
            return predicted_dist
        
        def kf_logp(data, c, d, T, R, H, Q, size=None):
            return ll
            
        return kf_filtered, kf_predicted, kf_logp
    
    kf_filtered, kf_predicted, kf_logp = kf_func_factory(data, a0, P0, c, d, T, R, H, Q)
    
    data = pytensor.shared(np.c_[np.arange(n)[:, None], np.full((n, m-p), np.nan)], name='data')
    filtered_dist = pm.CustomDist('kalman_filtered', c, d, T, R, H, Q, dist=kf_filtered, logp=kf_logp, observed=data)
    predicted_dist = pm.CustomDist('kalman_predicted', data, c, d, T, R, H, Q, dist=kf_predicted)
```

This raises errors like `C:\Users\Jesse\miniconda3\envs\econ\lib\site-packages\pymc\pytensorf.py:1056: UserWarning: RNG Variable RandomGeneratorSharedVariable(<Generator(PCG64) at 0x193D8EFF220>) has multiple clients. This is likely an inconsistent random graph.`. It also doesn't sample, but one step at a time. Obviously this would all go into an actual `RandomVariable`, I'm just trying to wrap my head around what needs to be done and how it would look.

ricardoV94: The error says you have a RNG variable that is being used in (at least) two distinct RandomVariables. If you can isolate a MWE I can try to pinpoint the problem more exactly.

Anyway if you put everything inside a single RandomVariable Op you don't have to worry about this sort of stuff. You should have a well defined Op that takes one RNG and does everything in numpy code and returns a well defined set of outputs.

jessegrabowski: You think I should do all this pytensor machinery in numpy? I'm trying not to duplicate code. The pytensor scan could be used to compute the log likelihood, but computing the log lilkelihood always generates the sampling distributions at the same time, so I was trying to think of a way to run the scan, then send the different components (2 distributions + logp graph) to where they need to go.

Basically the problem I'm having is that everything is being done all at once in this one scan, so I am struggling with how to "untangle" it. Another problem is how to untangle the data, which should be passed as observed, from the scanning, which always needs to be done over the data.

pymc_experimental/statespace/core/statespace.py
jessegrabowski: It silences the user warning you get for sampling with a potential. I'm only sampling from the parameter priors then computing the posterior predictive myself, so it's not relevant.  It can be removed after the refactor to a random variable.

pymc_experimental/tests/statespace/test_VARMAX.py
jessegrabowski: I was having trouble with invalid file paths when I ran the tests via github actions. I can try it again using pymc's CI setup.

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -1,121 +1,78 @@
...
+            outputs = (a_next, P_next)
+
+            updates = collect_default_updates(inputs=args, outputs=outputs)
+
+            return outputs, updates
+
+        (statespace, _), updates = pytensor.scan(
+            step_fn, outputs_info=[a0_, P0_], non_sequences=[c_, T_, R_, Q_], n_steps=steps_
+        )
+        (ss_rng,) = tuple(updates.values())
```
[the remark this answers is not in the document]
jessegrabowski: Also I only was able to figure out the order of the updates by grug-brain trial and error. Is it dis-recommended to use `collect_default_updates` when writing an internal distribution so I can control the order in which the random generators are expected? Or is there a way to predict it when I use `collect_default_updates`? Also is there a way to know which generator goes to which variable?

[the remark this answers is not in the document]
ricardoV94: I don't know exactly what's going wrong. In the AR Distribution I just created the SharedVariable manually and passed it as a non-sequence. You can then define your updates manually easily: https://github.com/pymc-devs/pymc/blob/6bfbff508b80e588cc03711798b4e2577ad46449/pymc/distributions/timeseries.py#L631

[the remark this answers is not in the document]
ricardoV94: About your ordering question: `collect_default_updates` gives you a mapping so you shouldn't have to guess what is what. I don't quite get why you need 2 updates (or why PyMC thinks you do) though.

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -0,0 +1,78 @@
...
+from pymc.pytensorf import collect_default_updates
+from pytensor.graph.basic import Node
+
+
+class LinearGaussianStateSpaceRV(SymbolicRandomVariable):
+    default_output = 1
+    _print_name = ("LinearGuassianStateSpace", "\\operatorname{LinearGuassianStateSpace}")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
ricardoV94, line 14: Forgot to remove?

jessegrabowski: I thought you have to do that with python subclasses

ricardoV94: Only if you want to do something custom

pymc_experimental/statespace/filters/distributions.py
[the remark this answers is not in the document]
jessegrabowski: I followed the SharedVariable approach for the discrete markov chain, but I was thinking that using `collect_default_updates` should be the new "correct" way?

Sort of related, if there are two random variables in each step, how should I handle it? For example, something like:
```python
def step(*args):
    level, trend, sigma_level, sigma_trend, old_rng = args
    trend_rng, epsilon_trend = pm.Normal.dist(0, sigma_trend, rng=??).owner.outputs
    level_rng, epsilon_level =pm.Normal.dist(0, sigma_level, rng=??).owner.outputs

    trend_rng, next_trend = trend + epsilon_trend
    level_rng, next_level = level + trend + epsilon_level

    return (next_trend, next_level), ??
``` 

My ideas were I need to either:
1. use 2 rngs, one for each random variable, so the update is `{old_trend_rng:trend_rng, old_level_rng:level_rng}`
2. chain them together, so old_rng goes into the first distribution and `trend_rng` comes out, then `trend_rng` goes into the next distribution and `level_rng` comes out, so the update is `{old_rng:level_rng}`

I tried both and neither worked, but I also didn't spend too long on it...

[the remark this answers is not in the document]
ricardoV94: That is really just a helper for users (who don't know anything about updates and where they come from). If it's not working for you no need to try and force it.

About the two RNGs, I suggest using the separate ones, but both forms are valid.

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -0,0 +1,157 @@
...
+
+class SequenceMvNormalRV(SymbolicRandomVariable):
+    default_output = 1
+    _print_name = ("SequenceMvNormal", "\\operatorname{SequenceMvNormal}")
+
+    def update(self, node: Node):
+        return {node.inputs[-1]: node.outputs[0]}
+
+
+class SequenceMvNormal(Distribution):
```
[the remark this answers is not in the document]
jessegrabowski: The logp graph still had 2 scans. I saw the biggest slowdowns in the dlogp graph, though. I will do some more work benchmarking the two cases, especially with JAX. I was having some trouble with the shapes of the input matrices. I need to make sure they all get (and stay) specified after the model is "built".

Yeah the 2nd scan is just a map, something like `map(lambda x, mu, cov: logp(pm.MvNormal.dist(mu, cov), x), zip([data, mus, covs]))`

jessegrabowski:
Getting close to a usable beta version. Current issue is prior predictive sampling, but the issue might be a result of the convoluted way I've set everything up. I'll try to explain it all as best as I can.

1. User declares some parameters
2. Parameters get shoved into some matrices
3. Matrices go into kalman filter, which outputs a sequence of means, covariances, and logp. The means and covariances include both hidden and observed states.
4. Means and covariances of the **observed** states, as well as the logp terms, go into a distribution
5. Distribution is used in the PyMC model

Steps 4 and 5 are there to get rid of the use of `pm.Potential`. The distribution, `SequenceMvNormal`, is a glorified `pm.CustomDist`.

Now you might ask, why do any of this, why not just take the matrices and directly use them in a distribution? There is another distribution for this in the codebase, `LinearGaussianStateSpace`. The reason why we have to do the kalman filtering is because many (most! even all!) states in the states space model are not observed. The point is the kalman filter is is to get the mean of these hidden states, conditional on the observed data.

So `SequentialMvNormal` lets us sample from the observed states during MCMC, but afterwords we also want the conditional posterior of the hidden states. A naive approach would be to do something like:

```
with pm.Model():
    # make matrices, run filter, get filtered_states, filtered_covariances (inferred hidden states)
    filtered_dist =pm.SequentialMvNormal('filtered', mus=filtered_states, covs=filtered_covariances)
```

But this term doesn't actually contribute anything to the model's logp. It's actually a set of deterministic parameters, so I thought it makes more sense to get it via `pm.sample_posterior_predictive`. So this is done via:

```
conditional_posterior = state_space_model.sample_conditional_posterior(idata, filter_output="filtered")
```

Internally, this creates a dummy model, makes `pm.Flat` dummies for `filtered_states` and `filtered_covariances`, and calls `pm.sample_posterior_predictive`. Similarly, if the user wants an *unconditional* posterior, it creates a dummy model, makes `pm.Flat` dummies for all the statespace matrices, and parameters a `LinearGaussianStateSpace` model that can be sampled with `pm.sample_posterior_predictive`.

The prior wrinkle comes because sometimes we want to know what the hidden states look like as priors, but currently we can only sample from the observed distribution in the prior step (the one made in step 5). In principle, to get prior over hidden states, we would have to:

1. Get prior samples for the state space matrices 
2. Close the model context
3. Rename the "prior" group in the idata to "posterior", to hack `pm.sample_posterior_predictive`
4. Open a new model context
5. Make dummies for all the state space matrices
6. Parameterize a `LinearGaussianStateSpace`
7. Call `pm.sample_posterior_predictive` with the hacked prior idata
8. Rename the "posterior_predictive" group to "prior_predictive"

But step (2) isn't possible, so either I need a genius way to refactor everything, or some kind of work-around.

jessegrabowski:
One other issue I'm hitting is that the kalman filtering can fill in missing values, but the PyMC automatic imputation machinery is getting in the way. Is there any way to turn it off? Or to force a `pm.Data` container to take data with missing values?

ricardoV94:
> Get prior samples for the state space matrices

Why can't this be done in posterior predictive (as new variables)? You may need to have two variables that in your model are conceptually one, but otherwise should be fine?

jessegrabowski:
You're right about priors. It just wouldn't be able to be done inside the model block, because I need to open a new context, prior sample some stuff, close that context, then open a new one. I was too hung up on that detail.

ricardoV94:
> You're right about priors. It just wouldn't be able to be done inside the model block, because I need to open a new context, prior sample some stuff, close that context, then open a new one. I was too hung up on that detail.

Why do you need two steps? Can't you introduce new variables in the posterior_predictive model?

jessegrabowski:
Oh yes you're right. As long as I'm already outside of the initial model definition context, I only need 1

jessegrabowski:
Docs are mostly in -- i'll keep working on those -- and all features that I think are most important are now in. Missing data works, prior sampling works, posterior sampling works, forcasting and IRF works. 3 of the 4 major model classes I want in there are in there.  I'd like to get it merged in so others can play with it and find more bugs/missing functionality.

"Major" missing features:

1. Support for time-varying matrices in the prior/posterior/forecast/irf 
2. Support for exogenous regressions in all models (related to 1)
3. Additive exponential smoothing models (that's the 4th model class)
4. Improved NaN handling. Right now I mask them with -9999 to slip it past the PyMC automatic interpolation, but I wish I could get the missing values back so the idata didn't have -9999 in it.

Mostly I just want to put this to rest for a while and move on to numba backend stuff, so these models can be sampled with nutpie.

jessegrabowski:
Also I have no idea why that prior sampling test is suddenly failing. It passes locally on my mac, windows, and linux machines in float32 mode.

ricardoV94 (commented):
Looks pretty clean!!! Gonna review the notebooks next to grok the API a bit better

conda-envs/environment-test.yml
```diff
@@ -9,6 +9,8 @@ dependencies:
 - pytest>=3.0
 - dask
 - xhistogram
+- statsmodels
+- numba
```
ricardoV94, line 13: Is this still needed anywhere?
```suggestion
```

notebooks/Custom SSM - Daily Seasonality.ipynb
```diff
@@ -0,0 +1,602 @@
+{
```
ricardoV94, line 1: <p>Busy plot, lower alpha again?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-experimental/pull/174/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

pymc_experimental/tests/statespace/test_SARIMAX.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.fixture
+def data():
+    return load_nile_test_data()
+
+
+@pytest.fixture
+def rng():
+    return np.random.default_rng(1337)
```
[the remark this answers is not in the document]
jessegrabowski: It is with a heavy heart that I changed the global test seed to something professional. I feel like Peter Pan leaving wonderland, except I'm in my mid-30s.

notebooks/Custom SSM - Daily Seasonality.ipynb
jessegrabowski: I was actually wanting to delete this notebook, it was an example I did for discourse at one point. There's a better example of how to make your own statespace in the PyMCStateSpace docstring. I'd replace this whole notebook with one based on that.

ricardoV94: Sure

pymc_experimental/tests/statespace/test_SARIMAX.py
```diff
@@ -0,0 +1,81 @@
...
+    assert_allclose(fast_eval(mod.ssm[matrix]), sm_sarimax.ssm[matrix])
+
+
+@pytest.mark.parametrize("order", orders, ids=ids)
+@pytest.mark.parametrize("matrix", ["transition", "selection", "state_cov", "obs_cov", "design"])
+def test_SARIMAX_update_matches_statsmodels(data, order, matrix, rng):
+    p, d, q = order
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
```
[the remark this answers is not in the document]
jessegrabowski: Statsmodels throws a warning if non-invertible/non-stationary parameters are given to the model. I'm just using random numbers for everything to check that the outputs match, so they're going to end up being invalid. I could figure out a set of test values, but I'm lazy.

pymc_experimental/statespace/core/statespace.py
```diff
@@ -0,0 +1,1296 @@
...
+        _verify_group(group)
+        group_idata = getattr(idata, group)
+        dims = None
+        temp_coords = self._fit_coords.copy()
+
+        if not use_data_time_dim and steps is not None:
+            temp_coords.update({TIME_DIM: np.arange(1 + steps, dtype="int")})
+            steps = len(temp_coords[TIME_DIM]) - 1
+        elif steps is not None:
+            n_dimsteps = len(temp_coords[TIME_DIM])
```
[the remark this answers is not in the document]
jessegrabowski: Yeah this is a rehash of that shape vs step conversation we had when I was doing the MarkovChain distribution, which admittedly I totally forgot about. I guess i really have my heart set on the idea that if I ask for 10 steps, I should get back a distribution with shape (10, k). I could change this argument to be called "shape" instead of "steps" to match the `distributions.timeseries` api?

In the statespace framework, the more lags makes the state vector larger, but there's always just one initial state, so no need to account for a variable number of initial states.

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -0,0 +1,271 @@
...
+            inputs=[a0_, P0_, c_, d_, T_, Z_, R_, H_, Q_, steps_],
+            outputs=[ss_rng, statespace_],
+            ndim_supp=1,
+        )
+
+        linear_gaussian_ss = linear_gaussian_ss_op(a0, P0, c, d, T, Z, R, H, Q, steps)
+        return linear_gaussian_ss
+
+
+class LinearGaussianStateSpace:
```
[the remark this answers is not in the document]
jessegrabowski: Yes I think so. I did `Distribution` for the other one, should they both be `Continuous` (since they both are)?

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -0,0 +1,271 @@
...
+import pytensor
+import pytensor.tensor as pt
+from pymc import intX
+from pymc.distributions.dist_math import check_parameters
+from pymc.distributions.distribution import Distribution, SymbolicRandomVariable
+from pymc.distributions.shape_utils import get_support_shape, get_support_shape_1d
+from pymc.gp.util import stabilize
+from pymc.logprob.abstract import _logprob
+
+# from pymc.logprob.basic import logp
```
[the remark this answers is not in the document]
jessegrabowski: Yep!

pymc_experimental/tests/statespace/test_SARIMAX.py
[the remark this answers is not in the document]
ricardoV94: Can you filter the specific warning?

pymc_experimental/tests/statespace/test_distributions.py
```diff
@@ -0,0 +1,129 @@
...
+        theta = ss_mod._gather_required_random_variables()
+        ss_mod.update(theta)
+        matrices = ss_mod.unpack_statespace()
+
+        filter_outputs = ss_mod.kalman_filter.build_graph(pymc_model["data"], *matrices)
+        filter_mus, pred_mus, obs_mu, filter_covs, pred_covs, obs_cov, ll = filter_outputs
+
+    test_ll = fast_eval(ll)
+
+    # TODO: BUG: Why does fast eval end up with a 2d output when filter is "single"?
```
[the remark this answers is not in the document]
jessegrabowski: I'll try to boil it down. Something in the re-writes is changing the shape, evidently in a way that my code depends on. Which is probably bad!

pymc_experimental/tests/statespace/test_SARIMAX.py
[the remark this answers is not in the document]
jessegrabowski: I just found out about `pytest.mark.filterwarnings`, so I'll use that instead.

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -0,0 +1,271 @@
...
+
+class LinearGaussianStateSpaceRV(SymbolicRandomVariable):
+    default_output = 1
+    _print_name = ("LinearGuassianStateSpace", "\\operatorname{LinearGuassianStateSpace}")
+
+    def update(self, node: Node):
+        return {node.inputs[-1]: node.outputs[0]}
+
+
+class _LinearGaussianStateSpace(Distribution):
```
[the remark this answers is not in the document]
jessegrabowski: Yes

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+@_logprob.register(SequenceMvNormalRV)
+def sequence_mvnormal_logp(op, values, mus, covs, logp, steps, rng, **kwargs):
+    # def step_logp(x, mu, cov):
+    #     return logp(pm.MvNormal.dist(mu, cov), x)
+    #
+    # logp_values, updates = pytensor.scan(
+    #     step_logp, sequences=[values[0], mus, covs], strict=True
+    # )
```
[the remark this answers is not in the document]
jessegrabowski: This was an old attempt to compute the logp of the data separately, rather than providing it pre-computed from the filtering. I forgot to take it out.

pymc_experimental/statespace/utils/data_tools.py
```diff
@@ -0,0 +1,145 @@
...
+        data_dims = [TIME_DIM, OBS_STATE_DIM]
+
+    pymc_mod.add_coord(TIME_DIM, index, mutable=True)
+    data = pm.ConstantData("data", values, dims=data_dims)
+
+    return data
+
+
+def mask_missing_values_in_data(values):
+    masked_values = np.ma.masked_invalid(values)
```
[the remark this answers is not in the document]
jessegrabowski: I was worrying about this too. It's a good idea.

pymc_experimental/statespace/utils/pytensor_scipy.py
```diff
@@ -0,0 +1,83 @@
...
+    R: ArrayLike
+        Square matrix of shape N x N
+
+    Returns
+    -------
+    X: at.matrix
+        Square matrix of shape M x M, representing the solution to the DARE
+    """
+
+    return SolveDiscreteARE()(A, B, Q, R)
```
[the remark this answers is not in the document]
jessegrabowski: Yes of course, that's an oversight.

pymc_experimental/statespace/utils/pytensor_scipy.py
```diff
@@ -0,0 +1,83 @@
...
+import pytensor.tensor as at
+import scipy
+from pytensor.tensor import TensorVariable, as_tensor_variable
+from pytensor.tensor.nlinalg import matrix_dot
+from pytensor.tensor.slinalg import solve_discrete_lyapunov
+
+floatX = pytensor.config.floatX
+
+
+class SolveDiscreteARE(at.Op):
```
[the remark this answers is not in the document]
ricardoV94: We can merge here and then port to PyTensor not to slow you down

pymc_experimental/statespace/core/statespace.py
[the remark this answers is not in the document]
ricardoV94: steps is mostly internal like in timeseries no? 

Most users will use shape/dims so I think it's better to stay consistent with timeseries meaning just for developers sake?

I am not a big fan of the step/shape distinction either btw

pymc_experimental/tests/statespace/test_distributions.py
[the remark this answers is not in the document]
ricardoV94: That sounds like a bug indeed

pymc_experimental/statespace/utils/data_tools.py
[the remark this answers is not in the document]
ricardoV94: Also we should emit the same imputation warning we do for vanilla observed with nan

pymc_experimental/statespace/core/statespace.py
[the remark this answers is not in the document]
jessegrabowski: Yeah steps is internal everywhere in the API.

This shows up here because when you do `sample_unconditional`, you can generate data for as long as you like. By default it matches the data's length, but the data isn't used at all in the computation, so I need to put something there for a "simulation length" argument. I guess it could be "simulation length".

notebooks/Custom SSM - Daily Seasonality.ipynb
jessegrabowski: I re-did this notebook, it's now called "Making a Custom Statespace Model". Have a look and see what you think.

pymc_experimental/tests/statespace/test_distributions.py
```diff
@@ -0,0 +1,129 @@
...
+    TIME_DIM,
+)
+from pymc_experimental.tests.statespace.utilities.test_helpers import (
+    load_nile_test_data,
+)
+
+floatX = pytensor.config.floatX
+
+# TODO: This needs to be VERY large for float32 to pass, is there a way to put scipy into float32 computation
+#  to get an apples-to-apples comparison?
```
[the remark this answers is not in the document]
jessegrabowski: This did the trick

pymc_experimental/statespace/utils/pytensor_scipy.py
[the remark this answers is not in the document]
jessegrabowski: Ok, I'm marking this as resolved then, and I'll open the Pytensor PR after this gets merged.

notebooks/Making a Custom Statespace Model.ipynb
```diff
@@ -0,0 +1,1464 @@
+{
```
ricardoV94, line 1: <p>I suggest doing a try/expect and printing the error in the except block. This way the notebook can be run all at once without having to resume after the error</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-experimental/pull/174/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

ricardoV94, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.        x0 = pm.Deterministic(
</pre><p>Do these need to be <code>Deterministic</code> for some of the conditional methods? Otherwise it would be more intuitive to write them as <code>ConstantData</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-experimental/pull/174/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

jessegrabowski:
The only failing test now is related to test_marginalized_change_point_model, which I haven't touched. Was there a breaking change upstream I need to merge into my branch?

ricardoV94:
Could have broken if it started picking up on new PyTensor/PyMC. I'll take a look

notebooks/Making a Custom Statespace Model.ipynb
jessegrabowski: They're not really data though. It's closest to a Dirac at the specified initial value. In the purest sense these should be estimated, but in practice usually they aren't, so I wanted to show both.

jessegrabowski:
<p>I think it was because there were one too many variables being estimated. I fixed that bug and the divergences went away.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-experimental/pull/174/discussion/#comment-1660562592'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jessegrabowski:
<p>Good idea!</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-experimental/pull/174/discussion/#comment-1660562647'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jessegrabowski:
<p>I think it's just nbviewer, it looks fine locally</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-experimental/pull/174/discussion/#comment-1660562712'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jessegrabowski:
<p>This notebook was out of date so I deleted it. But I also updated arviz just in case.</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-experimental/pull/174/discussion/#comment-1660562535'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jessegrabowski:
I think it's ready. The float32 failures are in a different module. Obviously still a lot to do, but it's also very usable as-is.

jessegrabowski:
Big refactor done, maybe have a look at the Making a custom statespace notebook to see how things work now. It's way better i think. First, it's the right way to do it in pytensor -- previously I was just copying a numpy implementation. Second, it gets rid of all the numpy internals in the `structural` module. Third, it adds unlimited flexibility in how parameters can be manipulated before/after going into the matrices. This is relevant if you want to estimate the length of a frequency period, for example.

I think the tests will pass on this next run (the irony of a commit called "all tests pass" failing CI is not lost on me, pride goeth before the fall), and then I would ask for this to be approved for merging, then I'll open PRs for the notebooks to go into pymc-examples, and for SolveDiscreteARE to go to pytensor. I'll also open an issue with a checklist of "to do" things still outstanding that I can work on (and maybe even drum up some contributors on).

pymc_experimental/statespace/models/utilities.py
```diff
@@ -0,0 +1,99 @@
...
+    return coords
+
+
+def variable_by_shape(name, shape, dtype):
+    if len(shape) <= 1:
+        return pt.vector(name, shape=shape, dtype=dtype)
+    elif len(shape) == 2:
+        return pt.matrix(name, shape=shape, dtype=dtype)
+    else:
+        return pt.tensor(name, shape=shape, dtype=dtype)
```
ricardoV94, line 99: `pt.tensor` accepts any shape (and is equivalent to vector/matrix/tensor3... for the same shape).

```suggestion
def variable_by_shape(name, shape, dtype):
  return pt.tensor(name, shape=shape, dtype)
```

But by this point you don't save anything with the helper

ricardoV94 (commented):
The integration with PyMC models looks sweet. 

As is usual for big PRs I left a comment about an unimportant part of the code!

pymc_experimental/statespace/models/utilities.py
jessegrabowski: I'm glad you picked up on this, I wanted to check with you if it was better to just do tensor or try to pick the right type.  I guess your suggestion is my answer!

ricardoV94:
I suggest you can codecov locally (we don't seem to have it in the CI here :( ) just to see if there's some important lines that aren't being tested.

ricardoV94 (approved):
Let me know if there's some part that you want me to inspect more carefully

ricardoV94:
> The PyMC code base mostly seems to use class constructors (is that the right name for the classmethod(__new__) design pattern? It's still a bit magic to me) for shared functionality.

That's an artifact that our classes are just functions in disguise. Not a pattern that we should try to stick with unless necessary. https://github.com/pymc-devs/pymc/issues/5308

ricardoV94:
> Also maybe have a look at how I handle missing values/data registration in data_tools.py and PyMCStateSpace.build_statespace_graph. I'm curious if I can put the nan values back in before it goes to sampling? I'm a bit disappointed that the -9999 values ends up in the idata.constant_data group, rather than the NaNs.

As I mentioned above, I think the clean solution is to store a boolean mask of the nan/not-nan positions, and then the unraveled non-nan entries. Not sure how helpful but you can have a look at the more recent `PartialObservedRV`: https://github.com/pymc-devs/pymc/blob/d59a960f89873667d6190489ff0e975091e57d10/pymc/distributions/distribution.py#L1166-L1292

pymc_experimental/statespace/filters/kalman_filter.py
```diff
@@ -0,0 +1,929 @@
...
+            An identity matrix of shape (k_states, k_states), stored for computational efficiency
+
+        eye_posdef : TensorVariable
+            An identity matrix of shape (k_posdef, k_posdef), stored for computational efficiency
+
+        eye_endog : TensorVariable
+            An identity matrix of shape (k_endog, k_endog), stored for computational efficiency
+        """
+
+        self.mode: str = mode
```
ricardoV94, line 71: This is used in 1 out of 3 scans in this module. Is that expected?

ricardoV94:
> kalman_filter.py is the beating heart, everything else is fancy bibs and bobs. If you can spot any optimizations there, it would offer a lot of improvement across the module. It's the most mathy, but there's also a lot of dumb logic about sorting variables by shape, adding checks, etc

I could only have a superficial look, not being familiar enough with this kind of models and math :(
The Asserts/SpecifyShapes should be pretty cheap / removable during compilation. The transpositions/concatenation are a tax we pay for every timeseries at the moment. Hopefully they aren't crazy expensive. Transpositions are as cheap as they get, what matters is how data layout plays into subsequent loops, but that's a bit too low-level optimization to look at before we see unreasonable bottlenecks. Concatenations may or not require copying arrays, I don't remember now.

I don't see anything obvious from my myopic perspective :D

pymc_experimental/statespace/models/SARIMAX.py
```diff
@@ -0,0 +1,409 @@
...
+        c = self.ssm["state_intercept"]
+
+        x0 = pt.linalg.solve(pt.identity_like(T) - T, c, assume_a="gen", check_finite=True)
+
+        method = "direct" if (self.k_states < 5) or (mode == "JAX") else "bilinear"
+        P0 = solve_discrete_lyapunov(T, pt.linalg.matrix_dot(R, Q, R.T), method=method)
+
+        return x0, P0
+
+    def make_symbolic_graph(self) -> None:
```
junpenglao, line 311: Are you using the Hamilton representation or Harvey here? Seeing that you are testing against the implementation in statsmodel I am assuming Harvey? Would be great to add a reference here.

[ref](http://www-stat.wharton.upenn.edu/~stine/stat910/lectures/14_state_space.pdf)

jessegrabowski: Yeah it's Harvey, but you can get Hamilton by using `state_structure = "interpretable"`. I could use the Hamilton/Harvey names (statsmodels does), but I was thinking about something that would help a new user know why to pick one or the other

junpenglao: Adding a comment should do, maybe something like "this is equivalent to the parameterization used in `statsmodels.tsa.statespace.sarimax.SARIMAX(..., hamilton_representation=False)` which use the Harvey representation (ref ...)"

jessegrabowski: I updated the docstring to be a bit more clear, and added a reference to Harvey (1989). Turns out I was wrong, I don't have Hamilton here, it's more verbose than that. My idea was to have the maximally performant (Harvey) and maximally interpretable (all states are pure AR lags or MA lags with no mixing). 

I'm not opposed to adding Hamilton and even Akaike, if you think people would be interested in having it.

junpenglao: Might be interesting to compare performance, but other than that I see no particular reason to add additional parameterization.

jessegrabowski: I was hoping you would say that : )

PR #182, opened 2023-05-31
Implement limit case of R2D2M2CP for P

### LINKED ISSUE #170: Implement strict sign for r2d2m2cp
There are cases where variable sign may cause symmetries in the posterior distribution, but the prior is still usefull in the context

So far 

```python
sigma, beta = pmx.distributions.R2D2M2CP(
            "beta",
            1,
            inputs.std(),
            dims="vars",
            r2=0.6,
            r2_std=0.1,
            positive_probs=[1, 0.7],
            variables_importance=[20, 10],
            centered=True,
        )
```

Will fail because `nan` will be resulted due to `positive_probs[0] = 1`

In this case I think it does make sense to use another distribution for the prior. The immediate choice is the Truncated normal distibuton with mean ans sigma calculated at `p=0.99`.

Additionally, in case positive_probs_std is provided, `std=0` would be required for the entry

Any other thoughts on this?

ferrine:
close #170 to enable strict coefficient sign in the model

```python
with pm.Model(coords=dict(a=range(2))) as model:
    std, beta = pmx.distributions.R2D2M2CP(
        "beta1",
        1,
        [1, 1],
        dims="a",
        r2=0.8,
        r2_std=0.1,
        positive_probs=[0.5, 0],
        positive_probs_std=[0.3, 0],
        centered=False,
    )
```
image

### DIFF
--- a/pymc_experimental/distributions/multivariate/r2d2m2cp.py
+++ b/pymc_experimental/distributions/multivariate/r2d2m2cp.py
@@ -394,7 +394,9 @@ def R2D2M2CP(
     *broadcast_dims, dim = dims
     input_sigma = pt.as_tensor(input_sigma)
     output_sigma = pt.as_tensor(output_sigma)
-    with pm.Model(name):
+    with pm.Model(name) as model:
+        if not all(isinstance(model.dim_lengths[d], pt.TensorConstant) for d in dims):
+            raise ValueError(f"{dims!r} should be constant length imutable dims")
         if r2_std is not None:
             r2 = pm.Beta("r2", mu=r2, sigma=r2_std, dims=broadcast_dims)
         phi = _phi(
--- a/pymc_experimental/distributions/multivariate/r2d2m2cp.py
+++ b/pymc_experimental/distributions/multivariate/r2d2m2cp.py
@@ -135,7 +135,7 @@ def _psi(positive_probs, positive_probs_std, *, dims):
 def _phi(
     variables_importance,
     variance_explained,
-    variance_explained_concentration,
+    importance_concentration,
     *,
     dims,
 ):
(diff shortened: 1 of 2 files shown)


ricardoV94:
Failing test should be fixed by #183

ferrine:
Just rebased

ferrine:
I found an issue if all the positive_probs are at the limit case

pymc_experimental/distributions/multivariate/r2d2m2cp.py
```diff
@@ -59,16 +67,141 @@ def _R2D2M2CP_beta(
...
 
 
+def _broadcast_as_dims(*values, dims):
+    model = pm.modelcontext(None)
+    shape = [len(model.coords[d]) for d in dims]
+    ret = tuple(np.broadcast_to(v, shape) for v in values)
+    # strip output
+    if len(values) == 1:
+        ret = ret[0]
+    return ret
```
ricardoV94, line 118: What if you have mutable dims?

ferrine: this dims can't be mutable since they are used to create RVs, no?

ricardoV94: Yes they can. RVs can have dynamic shapes (just not number of dimensions)

ferrine: these particular ones are supposed to be static shape parameters anyway..., like parameters in a linear regression.

ferrine: What would you suggest to do?

ferrine: @ricardoV94 it seems to be impossible to implement in a clean way with mutable dims, but I bet in future it can be refactored into some more generic implementation. The main challenge is to take in account the limit case where random variable is partial

ferrine: I doubt this particular use-
case is frequent, we can just create a new issue for that

ferrine: @ricardoV94 this is ready for review, I've added checks for that and opened an issue

ricardoV94 (approved):

PR #189, opened 2023-06-06
model_builder adaptations for full sklearn compat

michaelraczycki:
Adaptations for full sklearn compatibility:
All functions expect data to be passed around in X,y format known from sklearn
output_var now is a property, which allows to use it in loading functions (still save saves entire original dataset - concatenated X and y, so this was needed to be able to always tell which column is a target column)
fit now requires X to be pd.DataFrame and y to be pd.Series


michaelraczycki:
also in order to simplify the code, now the model_config and sampler_config can be passed only at the class initialization, other approaches where they could be passed to build model required constant checks and was polluting the code, without clear benefits.

pymc_experimental/preprocessing/standard_scaler.py
```diff
@@ -0,0 +1,16 @@
+import pandas as pd
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.preprocessing import StandardScaler
+
+
+class StandardScalerDF(StandardScaler, TransformerMixin, BaseEstimator):
```
twiecki, line 6: Is this only for dataframe support? Because I think sklearn added that: https://blog.scikit-learn.org/technical/pandas-dataframe-output-for-sklearn-transformer/

twiecki: Also https://ponder.io/scikit-learns-transformers-now-output-pandas-dataframes/

michaelraczycki: thanks! Indeed I've added that as I've noticed that It was always returning ndarray, which was causing us to loose info about column names, and that was not really acceptable. I'll include the configurable sklearn solution instead of that

pymc_experimental/model_builder.py
```diff
@@ -400,8 +423,8 @@ def load(cls, fname: str):
 
     def fit(
         self,
-        X: Union[np.ndarray, pd.DataFrame, pd.Series],
-        y: Union[np.ndarray, pd.Series],
+        X: pd.DataFrame,
+        y: pd.Series,
```
twiecki, line 427: Can we make `y` be optional.

pymc_experimental/preprocessing/standard_scaler.py
twiecki: But then you can just remove this code here, no?

pymc_experimental/model_builder.py
twiecki: The issue is that if we don't do that it will limit the models we can fit this on, as not all actually have an X and a y, for example, the change-point model only has a single time-series.

PR #210, opened 2023-07-05
ModelBuilder.load versatility improvements

michaelraczycki:
fixing dims format, enabling input param preservation.
The fixes needed to be applied for models that take input parameters that aren't included in the dataset. Furthermore the json conversion was changing dims from tuple to a list, which was in the end changing model id and failing the load function

### DIFF
--- a/pymc_experimental/tests/test_model_builder.py
+++ b/pymc_experimental/tests/test_model_builder.py
@@ -13,6 +13,7 @@
 #   limitations under the License.
 
 import hashlib
+import json
 import sys
 import tempfile
 from typing import Dict
@@ -43,8 +44,8 @@ def toy_y(toy_X):
 @pytest.fixture(scope="module")
 def fitted_model_instance(toy_X, toy_y):
     sampler_config = {
-        "draws": 500,
-        "tune": 300,
+        "draws": 100,
+        "tune": 100,
         "chains": 2,
         "target_accept": 0.95,
     }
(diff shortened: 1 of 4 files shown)


ricardoV94 (commented):
Can you add a test?

pymc_experimental/tests/test_model_builder.py
```diff
@@ -122,6 +131,38 @@ def default_sampler_config(self) -> Dict:
...
+        "a": {
+            "loc": 0,
+            "scale": 10,
+            "dims": [
+                "x",
+            ],
+        },
+    }
+    converted_model_config = fitted_model_instance._convert_dims_to_tuple(model_config)
+    assert converted_model_config["a"]["dims"] == ("x",)
```
ricardoV94, line 163: Would be more useful to have a test for the "global functionality" and not the specific function used internally. My understanding is that models with dims failed to save and load, and now they work. Can you test along those lines?

Then the test will stay relevant even if the internal way that's achieved changes.

michaelraczycki: Each save/load test in the child classes test this fix, as load itself checks the model IDs which are dependent on model config content and some other things

ricardoV94: I don't see why any of the other tests would fail if we reverted the fix. Only this test seems to actually check for dims?

michaelraczycki: Good catch, for some reason I omitted the id comparison in load while implementing load function in clv models. That's why it didn't complain about it the way DelayedSaturatedMMM was. Now it's fixed, also addressing the removal of temporary file that was mentioned in marketing

michaelraczycki: you can find the adapted tests here:
[https://github.com/pymc-labs/pymc-marketing/pull/317/commits/99568344d7ba68a54aad7119349f924b10ad4778](modified load with save/load_tests)

ricardoV94: I made the `_convert_dims_to_tuple` be a No-op, and none of the pre-existing tests failed: 
image

We should include a direct test that dims have the correct type after save/load, in this library :)

ricardoV94: Made that method simply return the `config_dict` without the logic to convert to tuple

michaelraczycki: @ricardoV94 I've introduced coords into the test model, now if you'd remove the _convert_dims_to_tuple, the load would fail as change from tuple to array in dims would result in different sha256, and in result fail the load id assertion

ricardoV94 (approved):

PR #216, opened 2023-07-17
Implement utility to change value variable transforms

ricardoV94:
Provides a cleaner solution to use cases like https://github.com/pymc-devs/pymc/issues/5443

### DIFF
--- a/pymc_experimental/model_transform/conditioning.py
+++ b/pymc_experimental/model_transform/conditioning.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional, Sequence, Union
+from typing import Any, List, Mapping, Optional, Sequence, Union
 
 from pymc import Model
 from pymc.logprob.transforms import RVTransform
@@ -26,7 +26,9 @@ from pymc_experimental.utils.model_fgraph import (
 from pymc_experimental.utils.pytensorf import rvs_in_graph
 
 
-def observe(model: Model, vars_to_observations: Dict[Union["str", TensorVariable], Any]) -> Model:
+def observe(
+    model: Model, vars_to_observations: Mapping[Union["str", TensorVariable], Any]
+) -> Model:
     """Convert free RVs or Deterministics to observed RVs.
 
     Parameters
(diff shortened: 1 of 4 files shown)


pymc_experimental/model_transform/conditioning.py
```diff
@@ -206,3 +213,132 @@ def do(
...
+    model: Model,
+    vars_to_transforms: Dict[ModelVariable, Union[RVTransform, None]],
+) -> Model:
+    """Change the value variables transforms in the model
+
+    Parameters
+    ----------
+    model: Model
+    vars_to_transforms: Dict
+        Mapping between RVs and new transforms to be applied to the respective value variables
```
michaelosthege, line 228:
```suggestion
    vars_to_transforms: Mapping[ModelVariable, Union[RVTransform, None]],
) -> Model:
    """Change the value variables transforms in the model

    Parameters
    ----------
    model : Model
    vars_to_transforms : dict-like
        Mapping from RVs to new transforms to be applied to the respective value variables
```

pymc_experimental/model_transform/conditioning.py
```diff
@@ -206,3 +213,132 @@ def do(
...
+        #   Mean, Standard deviation
+        # p 0.67, 0.16
+
+    """
+    vars_to_transforms = {
+        parse_vars(model, var)[0]: transform for var, transform in vars_to_transforms.items()
+    }
+
+    if set(vars_to_transforms.keys()) - set(model.free_RVs):
+        raise ValueError(f"All keys must be free variables in the model: {vars_to_transforms}")
```
michaelosthege, line 266:
```suggestion
        raise ValueError(f"All keys must be free variables in the model: {set(vars_to_transforms)}")
```

michaelosthege (approved):
interesting

pymc_experimental/model_transform/conditioning.py
ricardoV94: Took some suggestions, not sure about changing `Dict` to `Mapping`

pymc_experimental/model_transform/conditioning.py
ricardoV94: printing `model.free_RVs`. I don't want to make it a set so that the error message doesn't confusingly change when the same snippet is rerun

pymc_experimental/model_transform/conditioning.py
michaelosthege: Usually mypy complains about using `Dict` for a parameter

ricardoV94: > Usually mypy complains about using `Dict` for a parameter

Do you have some link? I never faced that

michaelosthege: https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance

ricardoV94: Changed

michaelosthege (approved):

PR #218, opened 2023-07-24
Copy model-related shared variables in `model_fgraph`

ricardoV94:
In earlier iterations of the do-blogpost it became clear that it's more intuitive for model-related shared variables to be copied, instead of shared across different models. Some of these variables are created by PyMC (dim_lengths) and users would have to know the internals not to be surprised about the behavior.

User defined `MutableData` is also copied, because unlike the underlying `SharedVariable`s, nothing about the name and documentation suggests these variables are supposed to be "shared" across models (just that they can be mutated).

I have only allowed user-defined shared variables that are *not* model variables to be actually shared across model cloning.

This now works as (imo) expected:

```python
import pymc as pm

with pm.Model(coords_mutable={"i": [0]}) as m:
    x = pm.Normal("x")
    y = pm.Normal("y", x, dims=("i",))
    
new_m = do(m, {x: 0})
new_m.set_dim("i", new_length=5, coord_values=list(range(5)))

assert pm.draw(m["y"]).shape == (1,)  # Before this would also be (5,)
assert pm.draw(new_m["y"].shape) == (5,)
```

This PR also fixes a bug, as RNG shared variables weren't actually being copied and rendered independent with the old `clone` approach. The relevant test was missing an `assert` :( (see first commit of this PR)

### DIFF
--- a/pymc_experimental/utils/model_fgraph.py
+++ b/pymc_experimental/utils/model_fgraph.py
@@ -4,7 +4,7 @@ import pytensor
 from pymc.logprob.transforms import RVTransform
 from pymc.model import Model
 from pymc.pytensorf import find_rng_nodes
-from pytensor import Variable
+from pytensor import Variable, shared
 from pytensor.graph import Apply, FunctionGraph, Op, node_rewriter
 from pytensor.graph.rewriting.basic import out2in
 from pytensor.scalar import Identity
@@ -184,8 +184,7 @@ def fgraph_from_model(
 
     # Replace RNG nodes so that seeding does not interfere with old model
     for rng in find_rng_nodes(model_vars):
-        new_rng = rng.clone()
-        new_rng.set_value(rng.get_value(borrow=False))
+        new_rng = shared(rng.get_value(borrow=False))
         memo[rng] = new_rng
 
     fgraph = FunctionGraph(
--- a/pymc_experimental/utils/model_fgraph.py
+++ b/pymc_experimental/utils/model_fgraph.py
@@ -1,3 +1,4 @@
+from copy import copy
 from typing import Dict, Optional, Sequence, Tuple
 
 import pytensor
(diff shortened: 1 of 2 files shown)


twiecki (approved):

pymc_experimental/utils/model_fgraph.py
```diff
@@ -184,8 +184,7 @@ def fgraph_from_model(
 
     # Replace RNG nodes so that seeding does not interfere with old model
     for rng in find_rng_nodes(model_vars):
-        new_rng = rng.clone()
-        new_rng.set_value(rng.get_value(borrow=False))
+        new_rng = shared(rng.get_value(borrow=False))
```
lucianopaz: I don't understand why the old approach didn't work. You were creating a clone, then setting its value to a new generator. Why would wrapping a new shared around a generator work and the old approach not.

pymc_experimental/utils/model_fgraph.py
```diff
@@ -182,10 +185,28 @@ def fgraph_from_model(
 
     memo = {}
 
-    # Replace RNG nodes so that seeding does not interfere with old model
-    for rng in find_rng_nodes(model_vars):
-        new_rng = shared(rng.get_value(borrow=False))
-        memo[rng] = new_rng
+    # Replace the following shared variables in the model:
+    # 1. RNGs
+    # 2. MutableData (could increase memory usage significantly)
```
lucianopaz, line 190: We'll have to wait and get some feedback. If people complain about this, we can review the memory sharing.

lucianopaz (approved):
LGTM. Great work @ricardoV94

pymc_experimental/utils/model_fgraph.py
ricardoV94: A sharedvariable.clone reuses the same container. The only thing that's cloned is the `Variable` object: https://github.com/pymc-devs/pytensor/blob/673c1accf98659ba0457759431cb36eef4659f63/pytensor/compile/sharedvalue.py#L139-L149

PR #226, opened 2023-08-03
Ignore constant shape dependencies in MarginalModel

ricardoV94:
CI started to fail in latest version of PyTensor due to https://github.com/pymc-devs/pytensor/pull/280/commits/0cb5fdbd6c3ec58cf0172db91041098f6e175ab7

Fix by ignoring dependecies when Shape ought to be constant (all static type shape of entries of input are fixed)

### DIFF
--- a/pymc_experimental/marginal_model.py
+++ b/pymc_experimental/marginal_model.py
@@ -21,6 +21,8 @@ from pytensor.tensor.elemwise import Elemwise
 
 __all__ = ["MarginalModel"]
 
+from pytensor.tensor.shape import Shape
+
 
 class MarginalModel(Model):
     """Subclass of PyMC Model that implements functionality for automatic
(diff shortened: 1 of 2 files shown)


pymc_experimental/marginal_model.py
```diff
@@ -251,9 +253,24 @@ class FiniteDiscreteMarginalRV(MarginalRV):
     """Base class for Finite Discrete Marginalized RVs"""
 
 
+def static_shape_ancestors(vars):
```
jessegrabowski, line 256: So before https://github.com/pymc-devs/pytensor/commit/0cb5fdbd6c3ec58cf0172db91041098f6e175ab7, these Shape Ops would have been caught by a the simple `isinstance(var, Constant)` check on line 276?

Now a "constant" shape is one which 1) has no `None`s, and 2) has only `Shape` `Op` ancestors?

pymc_experimental/tests/test_marginal_model.py
```diff
@@ -411,3 +415,14 @@ def test_marginalized_transforms(transform, expected_warning):
             transform_name = transform.name
         assert f"sigma_{transform_name}__" in ip
     np.testing.assert_allclose(m.compile_logp()(ip), m_ref.compile_logp()(ip))
+
+
+def test_is_conditional_dependent_static_shape():
+    """Test that we don't consider dependencies through "constant" shape Ops"""
+    x1 = pt.matrix("x1", shape=(None, 5))
+    y1 = pt.random.normal(size=pt.shape(x1))
+    assert is_conditional_dependent(y1, x1, [x1, y1])
```
jessegrabowski, line 424: I would have guessed this case would fail, because there is a `None` in the shape (line 256)?

jessegrabowski (approved):
looks good, just asked a couple questions to check that I understand how it works.

pymc_experimental/marginal_model.py
ricardoV94: Before there wouldn't be any dependency between x -> shape(x) -> y(shape(x)), because `shape(x)` would just be a constant (so no further edges to x)

PR #233, opened 2023-08-22
version bump

michaelraczycki:
0.0.10 -> 0.0.11


pymc_experimental/version.txt
```diff
@@ -1 +1,2 @@
-0.0.10
+
```
twiecki, line 1:
```suggestion
```

ricardoV94:
I don't remember if we need to bump twice in case we already did a release and added the 0.11 tag like it happened now

michaelraczycki:
I think I can just manually re-run the workflows after merging, and it should release to pypi

michaelraczycki:
if that will turn out to be blocking for the release, I'd just delete the 0.0.11 (since it has only my bug fix) and release 0.0.11 again

ricardoV94:
You can do that. You still have to delete the tag in the main branch: https://devconnected.com/how-to-delete-local-and-remote-tags-on-git/

PR #235, opened 2023-08-25
Change default_*_config from @property to @staticmethod

### LINKED ISSUE #225: change default_model_config form @property to @staticmethod
Current design is shorter for usage inside the classes themselves, but when it comes to end user experience it forces initialization of the model class to take a look at customizable variables

pdb5627:
Fixes #225. Since the method changed from a property to a method, it was renamed from a noun to a verb. Both `default_model_config` and `default_sampler_config` were changed.

### DIFF
--- a/pymc_experimental/linearmodel.py
+++ b/pymc_experimental/linearmodel.py
@@ -20,16 +20,16 @@ class LinearModel(ModelBuilder):
     _model_type = "LinearModel"
     version = "0.1"
 
-    @property
-    def default_model_config(self):
+    @classmethod
+    def get_default_model_config(self):
         return {
             "intercept": {"loc": 0, "scale": 10},
             "slope": {"loc": 0, "scale": 10},
             "obs_error": 2,
         }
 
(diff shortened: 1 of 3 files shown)


michaelraczycki:
@pdb5627 why did you decide to go with class method over static?

pdb5627:
> @pdb5627 why did you decide to go with class method over static?

@michaelraczycki By mistake! I'll edit the PR to change it from `classmethod` to `staticmethod`.

michaelraczycki:
The main reason why I wanted to make it @staticfunction over @classfunction, is what you said about the @property support in 3.11. In case of using static, we don't need to rename it to verb after giving up the property decorator, and in this way don't impact the main api. In general I'll always lean towards not changing api unless there's some direct and significant benefit for doing otherwise

michaelraczycki (commented):
lgtm, thank you @pdb5627 for your contributions!

michaelraczycki (approved):

Issue #97, opened 2022-12-12
Kernel crashes during pathfinder fitting


fonnesbeck:
For several models (not sure what characteristics yet cause if to fail), fitting with pathfinder causes the Python kernel to crash. I've tried this both on CPU and GPU. The log is below:

```
info 12:18:18.397: Disposing kernel .jvsc74a57bd0a0954a124d2bf69a9421ff2dbb2519042f2244cade5040e3d13151bd4c002865./Users/cfonnesbeck/mambaforge/envs/pie/python./Users/cfonnesbeck/mambaforge/envs/pie/python.-m#ipykernel_launcher for notebook Interactive-1.interactive due to selection of another kernel or closing of the notebook
info 12:18:18.397: Dispose Kernel 'Interactive-1.interactive' associated with '/Users/cfonnesbeck/phillies/pie/research/projections/pitchers/pitcher_proj.py'
info 12:18:18.406: Dispose Kernel 'Interactive-1.interactive' associated with '/Users/cfonnesbeck/phillies/pie/research/projections/pitchers/pitcher_proj.py'
[I 12:18:18.433 NotebookApp] Kernel shutdown: db435396-0e9c-48cf-8b9d-7ac31bccd3da
[I 12:18:19.402 NotebookApp] Starting buffering for db435396-0e9c-48cf-8b9d-7ac31bccd3da:05dcd16b-b821-4492-bbbf-5224df1874c6
info 12:18:23.589: Starting interactive window for resource '/Users/cfonnesbeck/phillies/pie/research/projections/pitchers/pitcher_proj.py' with controller '.jvsc74a57bd0a0954a124d2bf69a9421ff2dbb2519042f2244cade5040e3d13151bd4c002865./Users/cfonnesbeck/mambaforge/envs/pie/python./Users/cfonnesbeck/mambaforge/envs/pie/python.-m#ipykernel_launcher (Interactive)'
info 12:18:23.636: Attempting to start a server because of preload conditions ...
info 12:18:23.661: Process Execution: > ~/mambaforge/envs/pie/bin/python -m pip list
> ~/mambaforge/envs/pie/bin/python -m pip list
info 12:18:24.433: Got env vars with python /Users/cfonnesbeck/mambaforge/envs/pie/bin/python, with env var count 104 and custom env var count 93 in 780ms
[...]

PR #238, opened 2023-09-11
Fix issues in pathfinder wrapper

ricardoV94:
Related to #97 

We were doing senseless things, like drawing from an isometric MvNormal with mean two as the initial point of the VI fit and then discarding such fit? This killed my poor RAM :)

Also I am not sure what the `inference_loop` was doing. I am not familiar with the internals of blackjax (and the documentation references a version that hasn't yet been released), but I think we just want to do `init -> sample`? Can someone confirm this (@junpenglao @zaxtax?)

Pathfinder doesn't seem to work particularly well with the 8 school example. It's very sensitive to the initial value of `mu`, and the default converges to something that looks wrong, see [gist](https://gist.github.com/ricardoV94/eafd20ac47d63525253b0a8adf5e5d76). Because of this I don't know what better thing to assert in the test.

Also the pymc-examples is broken (it is not using mu or tau :P): https://www.pymc.io/projects/examples/en/latest/variational_inference/pathfinder.html


twiecki:
Ouch, well these all sounds like fixes to me.

junpenglao:
I think it is better to depends on blackjax-nightly (which would soon be release as blackjax v1), and use
```
pathfinder = blackjax.kernels.pathfinder(logdensity_fn)
state, _ = pathfinder.approximate(rng_key, ...)
_, rng_key = random.split(rng_key)
samples, _ = pathfinder.sample(rng_key, state, 5_000)
```
per https://blackjax-devs.github.io/sampling-book/algorithms/pathfinder.html

ricardoV94:
> I think it is better to depends on blackjax-nightly (which would soon be release as blackjax v1), and use

How soon is that soon? Would it be too much of a PITA to get another pre v1 release if that's faster?

junpenglao:
We just release Blackjax v1

twiecki (approved):

twiecki (approved):

twiecki:
Can we merge?

ricardoV94:
I would like to have a model where this actually works with as a test case, but since we are in experimental maybe that's fine.

ricardoV94:
Needs to be rebased for tests to pass

PR #239, opened 2023-09-13
Implement Chi distribution helper

tvwenger:
Following the discussion [here](https://github.com/pymc-devs/pymc/discussions/6905) and obsolete PRs [here](https://github.com/pymc-devs/pytensor/pull/440) and [here](https://github.com/pymc-devs/pymc/pull/6907), I have attempted to implement a Chi distribution helper class using `CustomDist`.

The distribution is functionally correct in that it is able to reproduce the `scipy` chi PDF.

There is an issue with adding this distribution to a `pymc` model, however. For example:

```python
import pymc as pm
from pymc_experimental.distributions import Chi
with pm.Model() as model:
    pm.Chi("x", df=1)
print(model.initial_point())
```
results in
```
TypeError: Cannot convert Type Scalar(float64, shape=()) (of Variable Exp.0) into Type Vector(float64, shape=(1,)). You can try to manually convert Exp.0 into a Vector(float64, shape=(1,)).
```
I can't quite narrow down this issue, so any help would be appreciated!

### DIFF
--- a/pymc_experimental/tests/distributions/test_continuous.py
+++ b/pymc_experimental/tests/distributions/test_continuous.py
@@ -26,6 +26,7 @@ from pymc.testing import (
     BaseTestDistributionRandom,
     Domain,
     R,
+    Rplus,
     Rplusbig,
     assert_moment_is_expected,
     check_logcdf,
@@ -35,7 +36,7 @@ from pymc.testing import (
 )
 
 # the distributions to be tested
-from pymc_experimental.distributions import GenExtreme
+from pymc_experimental.distributions import Chi, GenExtreme
 
 
 class TestGenExtremeClass:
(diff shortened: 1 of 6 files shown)


pymc_experimental/tests/distributions/test_continuous.py
```diff
@@ -149,3 +150,54 @@ class TestGenExtreme(BaseTestDistributionRandom):
...
+        [
+            (1, None, 1),
+            (1, 5, np.full(5, 1)),
+            (np.arange(1, 6), None, np.arange(1, 6)),
+        ],
+    )
+    def test_chi_moment(self, nu, size, expected):
+        with pm.Model() as model:
+            Chi("x", nu=nu, size=size)
+        assert_moment_is_expected(model, expected)
```
ricardoV94, line 189: No need for this. We are adding default moments to CustomDist in https://github.com/pymc-devs/pymc/pull/6873

tvwenger:
Alright, I fixed one problem and found another. I was passing the `nu` parameter to `CustomDist` as a tuple instead of as parameter. That fixes that error, but introduces a new one for the provided code snippet. The initial logp for the distribution is `-inf`:
```python
import pymc as pm
from pymc_experimental.distributions import Chi
with pm.Model() as model:
    Chi("x", nu=1)
print(model.initial_point())
# {'x_log__': array(-inf)}
```

ricardoV94:
That will be solved by the linked PR which implements initial values for CustomDist.

For now we can provide a maxwell_moment function to the moment kwarg of customDist. Probably it's good enough to set the moment to ones

ricardoV94:
Do you also want to add a helper for the Maxwell distribution?

ricardoV94:
The failing tests should be fixed by #240

ricardoV94:
Rebased from main, tests should now pass and we can merge

ricardoV94 (approved):

PR #240, opened 2023-09-14
Bump minimum PyMC dependency

ricardoV94:
Closes #237

PR #243, opened 2023-09-16
Fix imports from `gp.util`

jessegrabowski:
pymc-devs/pymc@ec64d40 breaks imports in the `latent-approx.py` module, this PR fixes it. Also does the same refactoring as that PR here, removing custom SolveTriangular Ops definitions with `pt.linalg.solve_triangular`


ricardoV94:
Duplicate of #240? Can you push there the missing fixes? That one also updates the minimum PyMC dependency which I think we have to

jessegrabowski:
OK, I will close this and put it over there instead.

PR #249, opened 2023-10-07
Set data when building Linearmodel

### LINKED ISSUE #248: LinearModel not setting data
When I originally worked on `LinearModel`, I assumed that `fit()` would set data after the model was built, so `build_model` just put in some dummy data. In later refactoring, this assumption turned out not to be true, so `LinearModel` never set the model data in the fitting process. I will prepare a PR for a test to check the fit of parameters and add setting data to `LinearModel.build_model`.

pdb5627:
Fixes #248 

I also have a commit in here with some clean-up to `ModelBuilder.sample_prior_predictive` that I saw while working on this. It's unrelated but relatively minor.

### DIFF
--- a/pymc_experimental/model_builder.py
+++ b/pymc_experimental/model_builder.py
@@ -608,7 +608,7 @@ class ModelBuilder:
             self.set_idata_attrs(prior_pred)
             if extend_idata:
                 if self.idata is not None:
-                    self.idata.extend(prior_pred)
+                    self.idata.extend(prior_pred, join="right")
                 else:
                     self.idata = prior_pred
 
--- a/pymc_experimental/model_builder.py
+++ b/pymc_experimental/model_builder.py
@@ -299,8 +299,8 @@ class ModelBuilder:
         with self.model:
             sampler_args = {**self.sampler_config, **kwargs}
             idata = pm.sample(**sampler_args)
-            idata.extend(pm.sample_prior_predictive())
-            idata.extend(pm.sample_posterior_predictive(idata))
+            idata.extend(pm.sample_prior_predictive(), join="right")
+            idata.extend(pm.sample_posterior_predictive(idata), join="right")
 
         idata = self.set_idata_attrs(idata)
         return idata
@@ -640,7 +640,7 @@ class ModelBuilder:
         with self.model:  # sample with new input data
             post_pred = pm.sample_posterior_predictive(self.idata, **kwargs)
             if extend_idata:
-                self.idata.extend(post_pred)
+                self.idata.extend(post_pred, join="right")
 
         posterior_predictive_samples = az.extract(
             post_pred, "posterior_predictive", combined=combined
--- a/pymc_experimental/model_builder.py
+++ b/pymc_experimental/model_builder.py
@@ -601,17 +601,16 @@ class ModelBuilder:
 
         if self.model is None:
             self.build_model(X_pred, y_pred)
-
-        self._data_setter(X_pred, y_pred)
-        if self.model is not None:
-            with self.model:  # sample with new input data
-                prior_pred: az.InferenceData = pm.sample_prior_predictive(samples, **kwargs)
-                self.set_idata_attrs(prior_pred)
-                if extend_idata:
-                    if self.idata is not None:
-                        self.idata.extend(prior_pred)
-                    else:
-                        self.idata = prior_pred
+        else:
+            self._data_setter(X_pred, y_pred)
+        with self.model:  # sample with new input data
+            prior_pred: az.InferenceData = pm.sample_prior_predictive(samples, **kwargs)
+            self.set_idata_attrs(prior_pred)
+            if extend_idata:
+                if self.idata is not None:
+                    self.idata.extend(prior_pred)
+                else:
+                    self.idata = prior_pred
 
         prior_predictive_samples = az.extract(prior_pred, "prior_predictive", combined=combined)
 
--- a/pymc_experimental/model_builder.py
+++ b/pymc_experimental/model_builder.py
@@ -595,7 +595,7 @@ class ModelBuilder:
             Prior predictive samples for each input X_pred
         """
         if y_pred is None:
-            y_pred = np.zeros(len(X_pred))
+            y_pred = pd.Series(np.zeros(len(X_pred)), name=self.output_var)
         if samples is None:
             samples = self.sampler_config.get("draws", 500)
 
--- a/pymc_experimental/tests/test_linearmodel.py
+++ b/pymc_experimental/tests/test_linearmodel.py
@@ -21,6 +21,7 @@ import numpy as np
 import pandas as pd
 import pytest
 import xarray as xr
+from pytest import approx
 
 from pymc_experimental.linearmodel import LinearModel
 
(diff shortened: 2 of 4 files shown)


twiecki:
Thanks @pdb5627! Not sure what the failing test is about, I restarted for now.

pymc_experimental/model_builder.py
```diff
@@ -601,17 +601,16 @@ def sample_prior_predictive(
...
-                    else:
-                        self.idata = prior_pred
+        else:
+            self._data_setter(X_pred, y_pred)
+        with self.model:  # sample with new input data
+            prior_pred: az.InferenceData = pm.sample_prior_predictive(samples, **kwargs)
+            self.set_idata_attrs(prior_pred)
+            if extend_idata:
+                if self.idata is not None:
+                    self.idata.extend(prior_pred)
```
ricardoV94, line 611: I think we want this so that it overrides pre-existing prior_predictive groups. Can you confirm in a test?

```suggestion
                    self.idata.extend(prior_pred, join="right")
```

pymc_experimental/tests/test_linearmodel.py
```diff
@@ -44,9 +54,10 @@ def toy_X():
 
 
 @pytest.fixture(scope="module")
-def toy_y(toy_X):
-    y = 5 * toy_X["input"] + 3
-    y = y + np.random.normal(0, 1, size=len(toy_X))
+def toy_y(toy_X, toy_actual_params):
+    y = toy_actual_params["slope"] * toy_X["input"] + toy_actual_params["intercept"]
+    np.random.seed(427)
+    y = y + np.random.normal(0, toy_actual_params["obs_error"], size=len(toy_X))
```
ricardoV94, line 60: Better not to mess with global seeding

```suggestion
    rng = np.random.default_rng(427)
    y = y + rng.normal(0, toy_actual_params["obs_error"], size=len(toy_X))
```

pymc_experimental/tests/test_linearmodel.py
```diff
@@ -101,6 +112,15 @@ def test_fit(fitted_linear_model_instance):
     assert isinstance(post_pred, xr.DataArray)
 
 
+def test_parameter_fit(fitted_linear_model_instance, toy_actual_params):
```
ricardoV94, line 115: Nitpick, we should actually fit in the test if we want to check it worked correctly.

The `fitted_model` used as a fixture should have a dummy fit that is fast enough, for tests that are not really testing the quality of fit.

pymc_experimental/tests/test_linearmodel.py
```diff
@@ -21,6 +21,7 @@
 import pandas as pd
 import pytest
 import xarray as xr
+from pytest import approx
```
ricardoV94, line 24: We usually use `np.testing.assert_allclose`, so it may be good to use it here also for consistency

pymc_experimental/tests/test_linearmodel.py
pdb5627: That's a good thought. I can work on that.

pymc_experimental/model_builder.py
pdb5627: There are might be some other calls to `extend` in other methods that should have `join="right"` added. I'll check, and if there are, I'll add a commit to add that.

pymc_experimental/tests/test_linearmodel.py
pdb5627: OK, good to know. I'll make that change.

twiecki:
Ping @ricardoV94

pymc_experimental/model_builder.py
```diff
@@ -601,17 +601,16 @@ def sample_prior_predictive(
...
-                    else:
-                        self.idata = prior_pred
+        else:
+            self._data_setter(X_pred, y_pred)
+        with self.model:  # sample with new input data
+            prior_pred: az.InferenceData = pm.sample_prior_predictive(samples, **kwargs)
+            self.set_idata_attrs(prior_pred)
+            if extend_idata:
+                if self.idata is not None:
+                    self.idata.extend(prior_pred, join="right")
```
ricardoV94, line 611: Can we test that all these `join="right"` are doing the right thing (i.e., discarding the old value and replacing the new one), and that `extend_idata=False` is being respected?

pdb5627: It should be possible. I'll work on it and update the PR accordingly.

pymc_experimental/tests/test_model_builder.py
```diff
@@ -220,6 +220,68 @@ def test_sample_posterior_predictive(fitted_model_instance, combined):
     assert np.issubdtype(pred[fitted_model_instance.output_var].dtype, np.floating)
 
 
+@pytest.mark.parametrize("extend_idata", [True, False])
+def test_sample_prior_extend_idata_param(fitted_model_instance, extend_idata):
```
ricardoV94, line 224: I'm concerned about mutating the `fitted_model_instance` in these tests, as it may be used in other tests. 
In general I am not a fan of fixtures that return mutable objects that we probably want to mutate further to investigate the behavior. I suggest creating a dummy model in this test and pass a copy of the fitted idata at the beginning.

Also can we parametrize the sample method to merge this and the posterior predictive tests. They seem to share most of the logic anyway.

pdb5627: I see what you're saying about modifying the `fitted_model_instance` object. My thought is to make the `fitted_model_instance` fixture return a copy of the model instance so individual tests can do what they like with it without any possibility of affecting other tests. The overhead for making a copy doesn't seem to add much to the test run time.

I originally wrote the prior and posterior predictive tests in the same test, but there were so many `if` `else` branches that I decided to split the test. But then I ended up making intermediate variables to clean up the code, so it turned out the same after all, so I'm combining them again. Thanks for the suggestion.

pymc_experimental/tests/test_model_builder.py
```diff
@@ -61,6 +65,14 @@ def fitted_model_instance(toy_X, toy_y):
     return model
 
 
+@pytest.fixture
+def fitted_model_instance(fitted_model_instance_base):
+    """Get a fitted model instance. The instance is copied after being fit,
+    so tests using this fixture can modify the model object without affecting
+    other tests."""
+    return copy.deepcopy(fitted_model_instance_base)
```
ricardoV94, line 73: copy doesn't really work for objects that have PyMC models: see https://github.com/pymc-devs/pymc/issues/6985

The approach is not too bad though. What I suggest is to create the idata once and then in this fixture recreate the model and glue-in a copy of the idata. I did something like that with a helper method in this PR: https://github.com/pymc-labs/pymc-marketing/pull/393/commits/44985a82d483d9defe7f80f6600818b64598b6c7

Check the `_build_with_idata` method and how that's used by `thin_fit_result`. Something similar could be used for a `ModelBuilder.copy()`

pdb5627: I came up with a workaround for copying the model without using `copy.deepcopy`.

I also noticed that there's a test marked for skipping on win32 due to lack of permissions for temp files, but the marked test doesn't use a temp file. There is a different test that does use a temp file. I thought maybe the annotation got onto the wrong test, so I made a commit to fix that possible issue. If that's wrong or you want to handle it as it's own issue, no problem, I'll take that commit back out.

ricardoV94 (approved):
Looks great!

twiecki:
Thanks for the contribution @pdb5627, does this unblock you for the ModelBuilder refactor?

PR #251, opened 2023-10-07
Fix CI errors in statspace test suite

jessegrabowski:
The statsmodels import in the statespace tests are causing everyone's CI to break. This PR fixes that.


jessegrabowski:
@ricardoV94 can you look at how I wrote the ignore statement in `setup.cfg` and see if you understand why it's not working?

jessegrabowski:
Down to one error, but it's in `pymc.sampling.jax`. [This line](https://github.com/pymc-devs/pymc/blob/602234b98d7c2c1e4382a56c16e7f71c1241bc8a/pymc/sampling/jax.py#L185):

```python
    postprocessing_backend: Literal["cpu", "gpu"] | None = None,
```

Is not 3.9 compatible (the pipe union operator was added in 3.10).

We can either skip the test here, or patch it on the pymc side. Thoughts @twiecki @ricardoV94 ?

twiecki:
@jessegrabowski Let's xfail it here, merge, and fix it on pymc.

jessegrabowski:
Looks like there's already a patch in the works here: https://github.com/pymc-devs/pymc/pull/6944

jessegrabowski:
Make sure those config files I removed weren't there for some other reason (related to building for instance). They seemed to just have pytest and pre-commit settings that were being ignored anyway

twiecki (approved):

twiecki:
You can test that by running `python setup.py sdist`

jessegrabowski:
Worked fine, I'm merging this.

PR #252, opened 2023-10-12
PC prior distribution for Student T dof

bwengals:
Moving this over from [here](https://github.com/pymc-devs/pymc/pull/6827).  

## What is this PR about?

Adding penalized complexity prior for the Student-T degrees of freedom parameter. Useful in models where the likelihood was normal, but you'd need some robustness so you switch to a Student T likelihood. It's implemented already in INLA.

The reason this is useful for modeling is you can "robustify" your Gaussian likelihood by making it a student t in a more principled way.  You can use this prior to express in a meaningful way "I think there is a 50% or 20% or whatever chance that the degrees of freedom is over 30 (~normal likelihood)", whereas if you use a Gamma(2, 0.1) or worse fix the degrees of freedom to some value, you risk watering down the information coming from the data via the likelihood.  

## Outstanding issues

- Still need to add tests but I think it'll be good enough for now to compare the logp to INLA.
- Might be better to use `BoundedContiuous` instead of `PositiveContinuous`?  Went down that rabbit hole some already, might need some help here.  Seems to work fine now for normal use, but I haven't tested too many weird edge cases.

### DIFF
--- a/pymc_experimental/distributions/continuous.py
+++ b/pymc_experimental/distributions/continuous.py
@@ -19,17 +19,26 @@ Experimental probability distributions for stochastic nodes in PyMC.
 The imports from pymc are not fully replicated here: add imports as necessary.
 """
 
-from typing import List, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 import numpy as np
 import pytensor.tensor as pt
 from pymc.distributions.dist_math import check_parameters
 from pymc.distributions.distribution import Continuous
 from pymc.distributions.shape_utils import rv_size_is_none
+from pymc.distributions.continuous import (
+    check_parameters, DIST_PARAMETER_TYPES, PositiveContinuous
+)
 from pymc.pytensorf import floatX
 from pytensor.tensor.random.op import RandomVariable
-from pytensor.tensor.variable import TensorVariable
+from pytensor.tensor import TensorVariable
 from scipy import stats
+
+from pymc_experimental.distributions.dist_math import (
+    studentt_kld_distance,
+    pc_prior_studentt_logp,
+    pc_prior_studentt_kld_dist_inv_op,   
+)
 
 
 class GenExtremeRV(RandomVariable):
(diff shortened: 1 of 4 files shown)


ricardoV94 (approved):

pymc_experimental/distributions/dist_math.py
```diff
@@ -0,0 +1,90 @@
...
+    return pt.sqrt(
+        1
+        + pt.log(2 * pt.reciprocal(nu - 2))
+        + 2 * pt.gammaln((nu + 1) / 2)
+        - 2 * pt.gammaln(nu / 2)
+        - (nu + 1) * (pt.digamma((nu + 1) / 2) - pt.digamma(nu / 2))
+    )
+
+
+def tri_gamma_approx(x):
```
ferrine, line 37: it is already implemented

pymc_experimental/distributions/continuous.py
```diff
@@ -216,3 +226,62 @@ def moment(rv, size, mu, sigma, xi):
         if not rv_size_is_none(size):
             mode = pt.full(size, mode)
         return mode
+
+
+class PCPriorStudentT_dof_RV(RandomVariable):
```
ferrine, line 231: needs a docstring

ricardoV94: Usually we don't document the RV, but the Distribution class, which doesn't have a docstring either

pymc_experimental/distributions/continuous.py
```diff
@@ -216,3 +226,62 @@ def moment(rv, size, mu, sigma, xi):
...
+    def moment(rv, size, lam):
+        mean = pc_prior_studentt_kld_dist_inv_op(1.0 / lam)
+        if not rv_size_is_none(size):
+            mean = pt.full(size, mean)
+        return mean
+
+    @classmethod
+    def get_lam(cls, alpha=None, U=None, lam=None):
+        if (alpha is not None) and (U is not None):
+            return -np.log(alpha) / studentt_kld_distance(U)
```
ricardoV94, line 271: Will this work with symbolic inputs?

pymc_experimental/distributions/dist_math.py
ricardoV94: This approximation will be much more performant

pymc_experimental/distributions/dist_math.py
```diff
@@ -0,0 +1,90 @@
...
+    """This function constructs a numerical approximation to the inverse of the KLD distance
+    function, `studentt_kld_distance`.  It does a spline fit for degrees of freedom values
+    from 2 + 1e-6 to 4000.  2 is the smallest valid value for the student t degrees of freedom, and
+    values above 4000 don't seem to change much (nearly Gaussian past 30).  It's then wrapped by
+    `SplineWrapper` so it can be used as a PyTensor op.
+    """
+    NU_MIN = 2.0 + 1e-6
+    nu = np.concatenate((np.linspace(NU_MIN, 2.4, 2000), np.linspace(2.4 + 1e-4, 4000, 10000)))
+    return UnivariateSpline(
+        studentt_kld_distance(nu).eval()[::-1],
```
ricardoV94, line 82: Having an eval is a bit dangerous. If it comes up from an RV you're going to get a random value. The safe thing to do is to `constant_fold` and raise if it can't be done.

Or create a PyTensor Op that wraps UnivariateSpline

pymc_experimental/distributions/continuous.py
ricardoV94, line 271:
```suggestion
            return -pt.log(alpha) / studentt_kld_distance(U)
```

pymc_experimental/distributions/dist_math.py
bwengals: I saw you added trigamma recently, I'll give that a try.  I used this approx because at the time the gradient wasn't implement yet, where the gradient for the approx is easy.  Wasn't concerned with performance at the time, but will take another look

pymc_experimental/distributions/dist_math.py
bwengals: @ricardoV94 It only comes from `nu`, which is passed in above as a fixed numpy array, so I think `eval` is safe here  (unless I'm missing your point).  

Thanks @ferrine, will look into that.  I remember needing to use `UnivariateSpline` this way because I needed this particular behavior

> if ext=3 of ‘const’, return the boundary value.

as `nu` goes to infinity.

bwengals:
To update where this is at, tests were passing except for windows and I couldn't figure out why.  Not sure what here would depend on windows so maybe need to dig into that in a VM or something to figure that out.  Definitely some room to improve how some of this is structured, but was working well enough I was comfortable using it.

pymc_experimental/distributions/dist_math.py
ricardoV94: Ah I missed the inputs were constant, nvm on my end

ricardoV94:
@bwengals the biggest difference with Windows generally is that they sometimes default to int32 dtypes where linux defaults to int64 IIRC. Maybe something in the np.linspace overflows in windows? (I didn't look at which test was even failing, so apologies if this is completely of the target)

bwengals:
> overflows in windows? 

Ah I bet that's it!  The test produces inf and it shouldn't.  Well thanks for giving me a starting place

pymc_experimental/distributions/dist_math.py
jessegrabowski: If it's always a know constant could you use `.data` instead of `.eval()`?

ricardoV94:
@bwengals interested in updating this one?

PR #255, opened 2023-10-27
Fix DiscreteMarkovChain logp

### LINKED ISSUE #254: Discrete_markov_chain.ipynb
Hi everyone! 

I am trying to run the discrete_markov_chain.ipynb example but I get two errors in two different parts of the code, the first when it does a demonstration of the API, in the seventh part ie: 

with pm.Model() as model:
    x0 = pm.Categorical.dist(np.ones(3) / 3, size=(100,))
    P = pm.Dirichlet("P", a=[1, 1, 1], size=(3,))
    discrete_mc = DiscreteMarkovChain("MarkovChain", P=P, init_dist=x0, observed=chains)
    idata = pm.sample()

With the following error: 
ValueError: TensorType broadcastable/shape must be a boolean, integer or None, got <class 'numpy.int64'> 1

In part thirteen, the part where you create the actual MS-AR model, ie: 

with hmm:
    # You have to assign BinaryMetropolis by hand, the default is Metropolis and it breaks.
    # BinaryGibbsMetropois seems to do much better than BinaryMetropolis, but also much slower.
    idata = pm.sample(
        step=[
            pm.BinaryGibbsMetropolis([hidden_states]),
            pm.NUTS([P, state_mus, ar_coefs, sigma], target_accept=0.99),
        ],
        drawings=5_000,
    )

With the following error: 
ValueError: Random variables detected in logp graph: {bernoulli_rv{0, (0,), int64, False}.out}.
This can happen when the DensityDist logp or Interval transform functions refer to nonlocal variables,
or when not all rvs have a corresponding value variable.

ricardoV94:
Closes #254


jessegrabowski (approved):
Looks great, thanks for fixing this up!

ricardoV94:
Regarding divergences yeah probably just a seed thing

Issue #95, opened 2022-11-22
Extend marginalization support to Truncated discrete RVs


ricardoV94:
#91 Implements automatic logp marginalization for bernoulli, categorical and discrete uniform, as those are the only default discrete  univariate RVs with finite support. We could extend it easily to right-truncated discrete RVs so that people can use it for Truncated poisson, binomial, mixtures, etc...)

We could also extend it to clipped RVs, but I am not sure how useful that is

PR #257, opened 2023-11-02
Support HMM via marginalization of DiscreteMarkovChain

### LINKED ISSUE #167: Cover `DiscreteMarkovChain` distributions with the marginal models

ricardoV94:
The following example defines a 2-state HMM, with a 0.9 transition probability of staying in the same state, and a Normal emission centered around -1 for state 0 and 1 for state 1.

```python
import arviz as az
import numpy as np
import matplotlib.pyplot as plt
import pymc as pm
from pymc_experimental import MarginalModel
from pymc_experimental.distributions import DiscreteMarkovChain

with MarginalModel() as m:
    P = [[0.9, 0.1], [0.1, 0.9]]
    init_dist = pm.Categorical.dist(p=[1, 0])
    chain = DiscreteMarkovChain("chain", P=P, init_dist=init_dist, steps=10)
    emission = pm.Normal("emission", mu=chain * 2 - 1, sigma=0.5)

    m.marginalize([chain])
    
    with m:
        idata = pm.sample(100)

plt.plot(az.extract(idata)["emission"].values, color="k", alpha=0.03)
plt.yticks([-1, 1])
plt.ylabel("Emission")
plt.xlabel("Step");
```
image

## Not implemented

Higher order lags and batch P matrices not supported due to complexity (and me not groking the exact API)

Closes #167

### DIFF
--- a/pymc_experimental/model/marginal_model.py
+++ b/pymc_experimental/model/marginal_model.py
@@ -10,21 +10,16 @@ from pymc.backends.arviz import coords_and_dims_for_inferencedata
 from pymc.distributions.discrete import Bernoulli, Categorical, DiscreteUniform
 from pymc.distributions.transforms import Chain
 from pymc.logprob.abstract import _logprob
-from pymc.logprob.basic import conditional_logp
+from pymc.logprob.basic import conditional_logp, logp
 from pymc.logprob.transforms import IntervalTransform
 from pymc.model import Model
 from pymc.pytensorf import compile_pymc, constant_fold, inputvars
 from pymc.util import _get_seeds_per_chain, dataset_to_point_list, treedict
-from pytensor import Mode
+from pytensor import Mode, scan
 from pytensor.compile import SharedVariable
 from pytensor.compile.builders import OpFromGraph
-from pytensor.graph import (
-    Constant,
-    FunctionGraph,
-    ancestors,
-    clone_replace,
-    vectorize_graph,
-)
+from pytensor.graph import Constant, FunctionGraph, ancestors, clone_replace
+from pytensor.graph.replace import vectorize_graph
 from pytensor.scan import map as scan_map
 from pytensor.tensor import TensorType, TensorVariable
 from pytensor.tensor.elemwise import Elemwise
@@ -32,6 +27,8 @@ from pytensor.tensor.shape import Shape
 from pytensor.tensor.special import log_softmax
 
 __all__ = ["MarginalModel"]
+
+from pymc_experimental.distributions import DiscreteMarkovChain
 
 
 class MarginalModel(Model):
@@ -247,16 +244,25 @@ class MarginalModel(Model):
             self[var] if isinstance(var, str) else var for var in rvs_to_marginalize
         ]
 
-        supported_dists = (Bernoulli, Categorical, DiscreteUniform)
         for rv_to_marginalize in rvs_to_marginalize:
             if rv_to_marginalize not in self.free_RVs:
                 raise ValueError(
                     f"Marginalized RV {rv_to_marginalize} is not a free RV in the model"
                 )
-            if not isinstance(rv_to_marginalize.owner.op, supported_dists):
+
+            rv_op = rv_to_marginalize.owner.op
+            if isinstance(rv_op, DiscreteMarkovChain):
+                if rv_op.n_lags > 1:
+                    raise NotImplementedError(
+                        "Marginalization for DiscreteMarkovChain with n_lags > 1 is not supported"
+                    )
+                if rv_to_marginalize.owner.inputs[0].type.ndim > 2:
+                    raise NotImplementedError(
+                        "Marginalization for DiscreteMarkovChain with non-matrix transition probability is not supported"
+                    )
+            elif not isinstance(rv_op, (Bernoulli, Categorical, DiscreteUniform)):
                 raise NotImplementedError(
-                    f"RV with distribution {rv_to_marginalize.owner.op} cannot be marginalized. "
-                    f"Supported distribution include {supported_dists}"
+                    f"Marginalization of RV with distribution {rv_to_marginalize.owner.op} is not supported"
                 )
 
             if rv_to_marginalize.name in self.named_vars_to_dims:
@@ -492,6 +498,10 @@ class FiniteDiscreteMarginalRV(MarginalRV):
     """Base class for Finite Discrete Marginalized RVs"""
 
 
+class DiscreteMarginalMarkovChainRV(MarginalRV):
+    """Base class for Discrete Marginal Markov Chain RVs"""
+
+
 def static_shape_ancestors(vars):
     """Identify ancestors Shape Ops of static shapes (therefore constant in a valid graph)."""
     return [
@@ -620,11 +630,17 @@ def replace_finite_discrete_marginal_subgraph(fgraph, rv_to_marginalize, all_rvs
     replace_inputs.update({input_rv: input_rv.type() for input_rv in input_rvs})
     cloned_outputs = clone_replace(outputs, replace=replace_inputs)
 
-    marginalization_op = FiniteDiscreteMarginalRV(
+    if isinstance(rv_to_marginalize.owner.op, DiscreteMarkovChain):
+        marginalize_constructor = DiscreteMarginalMarkovChainRV
+    else:
+        marginalize_constructor = FiniteDiscreteMarginalRV
+
+    marginalization_op = marginalize_constructor(
         inputs=list(replace_inputs.values()),
         outputs=cloned_outputs,
         ndim_supp=ndim_supp,
     )
+
     marginalized_rvs = marginalization_op(*replace_inputs.keys())
     fgraph.replace_all(tuple(zip(rvs_to_marginalize, marginalized_rvs)))
     return rvs_to_marginalize, marginalized_rvs
@@ -640,6 +656,9 @@ def get_domain_of_finite_discrete_rv(rv: TensorVariable) -> Tuple[int, ...]:
     elif isinstance(op, DiscreteUniform):
         lower, upper = constant_fold(rv.owner.inputs[3:])
         return tuple(range(lower, upper + 1))
+    elif isinstance(op, DiscreteMarkovChain):
+        P = rv.owner.inputs[0]
+        return tuple(range(pt.get_vector_length(P[-1])))
 
     raise NotImplementedError(f"Cannot compute domain for op {op}")
 
@@ -647,7 +666,7 @@ def get_domain_of_finite_discrete_rv(rv: TensorVariable) -> Tuple[int, ...]:
 def _add_reduce_batch_dependent_logps(
     marginalized_type: TensorType, dependent_logps: Sequence[TensorVariable]
 ):
-    """Add the logps of dependent RVs while reducing extra batch dims as assessed from the `marginalized_type`."""
+    """Add the logps of dependent RVs while reducing extra batch dims relative to `marginalized_type`."""
 
     mbcast = marginalized_type.broadcastable
     reduced_logps = []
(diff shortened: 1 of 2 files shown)


junpenglao:
Is it using the Viterbi algorithm?

jessegrabowski:
Currently just the forward algorithm to compute the logp

junpenglao:
> If I understand well, viterbi just gives the most probable sequence of hidden states in a maximum likelihood setting? We should be able to back that out of the posterior pretty easily. You'll need to school me if I'm over simplifying.

yes Viterbi gives the posterior mode - but you are marginalizing the state to compute the likelihood here right?

ricardoV94:
> but you are marginalizing the state to compute the likelihood here right?

Yes, but to be precise: to compute the logp of any dependent variables, which may be observed/unobserved or a mix.

pymc_experimental/marginal_model.py
```diff
@@ -478,3 +498,64 @@ def logp_fn(marginalized_rv_const, *non_sequences):
...
+    log_alpha_seq = pt.moveaxis(pt.concatenate([logp_init[None], log_alpha_seq], axis=0), -1, 0)
+
+    # Final logp is the sum of the sum of the emission probs and the transition probabilities
+    # pt.add is used in case there are multiple emissions that depend on the same markov chain; in this case, we compute
+    # the joint probability of seeing everything together.
+    joint_log_obs_given_states = pt.logsumexp(pt.add(*vec_logp_emission) + log_alpha_seq, axis=0)
+
+    # If there are multple emisson streams, we have to add dummy logps for the remaining value variables. The first
+    # return is the joint probability of everything together, but PyMC still expects one logp for each one.
+    dummy_logps = (pt.constant(np.zeros((4,))),) * (len(values) - 1)
```
ricardoV94, line 560: Does it work with empty arrays? That would be "mathematically cleaner". Same for the pre-existing marginal discrete logp
```suggestion
    dummy_logps = (pt.constant(np.zeros((0,))),) * (len(values) - 1)
```

jessegrabowski: This is debug code that I left in my accident. I'll make a test case for empty arrays

ricardoV94:
Seems like our "clever" approach is not correct. We need to combine the emission probabilities as we compute the state probabilities. I thought we could factor them out but it doesn't seem to be the case.

jessegrabowski:
> Seems like our "clever" approach is not correct. We need to combine the emission probabilities as we compute the state probabilities iteratively. I thought we could factor them out but it doesn't seem to be the case.

I added the example from this youtube vid as a test case, so we can get to a solution.

I'm in the process of refactoring the logp function to compute alpha correctly, but it's typically a nested loop. Here's numpy code:

```
transition_probs = np.array([[0.5, 0.5], 
                             [0.3, 0.7]])
initial_probs = np.array([0.375, 0.625])

T = 3
data = [0, 0, 1]
log_alpha = np.zeros((T, 2))
x_dists = [stats.bernoulli(p=0.2), stats.bernoulli(p=0.6)]

def eval_logp(x, dists):
    return np.array([d.logpmf(x) for d in dists])

log_alpha[0, :] = np.log(initial_probs) + eval_logp(data[0], x_dists)
for t in range(1, T):
    obs = data[t]
    for s in range(transition_probs.shape[0]):
        step_log_prob = x_dists[s].logpmf(obs) + np.log(transition_probs[:, s]) + log_alpha[t-1, :]
        log_alpha[t, s] = logsumexp(step_log_prob)
```
I'm trying to think how we can vectorize the inner loop, open to suggestions.

pymc_experimental/marginal_model.py
```diff
@@ -378,11 +388,17 @@ def replace_finite_discrete_marginal_subgraph(fgraph, rv_to_marginalize, all_rvs
...
+    if isinstance(rv_to_marginalize.owner.op, DiscreteMarkovChain):
+        marginalize_constructor = DiscreteMarginalMarkovChainRV
+    else:
+        marginalize_constructor = FiniteDiscreteMarginalRV
+
+    marginalization_op = marginalize_constructor(
         inputs=list(replace_inputs.values()),
         outputs=cloned_outputs,
-        ndim_supp=0,
+        ndim_supp=ndim_supp,
```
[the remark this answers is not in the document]
jessegrabowski: Right, I was misunderstanding what was being checked by ndim_supp. I restored the check that there are no up-stream MvRVs, and hard-coded the expected ndim_supp for each "class" of variable we support

pymc_experimental/marginal_model.py
```diff
@@ -478,3 +498,64 @@ def logp_fn(marginalized_rv_const, *non_sequences):
...
+        replace={u: v for u, v in zip(op.inner_inputs, marginalized_rvs_node.inputs)},
+    )
+
+    chain_rv, *dependent_rvs = inner_rvs
+    P_, n_steps_, init_dist_, rng = chain_rv.owner.inputs
+
+    domain = pt.arange(P_.shape[0], dtype="int32")
+
+    vec_eval_logp = pt.vectorize(eval_logp, "()->()")
+    logp_init = vec_eval_logp(domain)
```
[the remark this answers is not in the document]
jessegrabowski: What is the sub_dict for the `graph_replace` approach? Can we just directly compute `logp(init_dist, domain)`, or am I missing something?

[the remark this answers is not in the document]
ricardoV94: Not all logps allow you to pass batches values (which you are doing here with the domain)

pymc_experimental/marginal_model.py
```diff
@@ -478,3 +498,64 @@ def logp_fn(marginalized_rv_const, *non_sequences):
...
+        step_alpha, non_sequences=[pt.log(P_)], outputs_info=[logp_init], n_steps=n_steps_
+    )
+
+    # Scan works over the T dimension, so output is (T, k). We need to swap to (k, T)
+    log_alpha_seq = pt.moveaxis(pt.concatenate([logp_init[None], log_alpha_seq], axis=0), -1, 0)
+
+    # Final logp is the sum of the sum of the emission probs and the transition probabilities
+    # pt.add is used in case there are multiple emissions that depend on the same markov chain; in this case, we compute
+    # the joint probability of seeing everything together.
+    joint_log_obs_given_states = pt.logsumexp(pt.add(*vec_logp_emission) + log_alpha_seq, axis=0)
```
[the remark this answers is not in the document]
jessegrabowski: I think we can do this add way sooner, check the updated code

pymc_experimental/model/marginal_model.py
```diff
@@ -245,16 +242,25 @@ def marginalize(
...
+
+            rv_op = rv_to_marginalize.owner.op
+            if isinstance(rv_op, DiscreteMarkovChain):
+                if rv_op.n_lags > 1:
+                    raise NotImplementedError(
+                        "Marginalization for DiscreteMarkovChain with n_lags > 1 is not supported"
+                    )
+                if rv_to_marginalize.owner.inputs[0].type.ndim > 2:
+                    raise NotImplementedError(
+                        "Marginalization for DiscreteMarkovChain with non-matrix transition probability is not supported"
```
jessegrabowski, line 259: Can a markov chain have a non-matrix transition probability?

pymc_experimental/model/marginal_model.py
```diff
@@ -245,16 +242,25 @@ def marginalize(
...
                 raise ValueError(
                     f"Marginalized RV {rv_to_marginalize} is not a free RV in the model"
                 )
-            if not isinstance(rv_to_marginalize.owner.op, supported_dists):
+
+            rv_op = rv_to_marginalize.owner.op
+            if isinstance(rv_op, DiscreteMarkovChain):
+                if rv_op.n_lags > 1:
+                    raise NotImplementedError(
+                        "Marginalization for DiscreteMarkovChain with n_lags > 1 is not supported"
```
jessegrabowski, line 255: Not a blocker, but we should be able to support this. Here's a paper with equations for a 2nd order HMM, we should be able to work out the general formula?

https://journals.sagepub.com/doi/pdf/10.1177/1550147718772541

pymc_experimental/model/marginal_model.py
```diff
@@ -454,18 +455,20 @@ def transform_input(inputs):
...
                 tuple(len(coord) for coord in stacked_dims.values()) + logps.shape[1:],
             )
-            if rv.name in m.named_vars_to_dims:
-                rv_dims[rv.name] = list(m.named_vars_to_dims[rv.name])
-                rv_dims["lp_" + rv.name] = rv_dims[rv.name] + ["lp_" + rv.name + "_dim"]
+            if marginalized_rv.name in m.named_vars_to_dims:
+                rv_dims[marginalized_rv.name] = list(m.named_vars_to_dims[marginalized_rv.name])
+                rv_dims["lp_" + marginalized_rv.name] = rv_dims[marginalized_rv.name] + [
+                    "lp_" + marginalized_rv.name + "_dim"
+                ]
```
jessegrabowski, line 471: Changes on lines 397-471 are unrelated to HMM yes?

pymc_experimental/model/marginal_model.py
```diff
@@ -580,10 +587,13 @@ def replace_finite_discrete_marginal_subgraph(fgraph, rv_to_marginalize, all_rvs
         raise ValueError(f"No RVs depend on marginalized RV {rv_to_marginalize}")
 
     ndim_supp = {rv.owner.op.ndim_supp for rv in dependent_rvs}
-    if max(ndim_supp) > 0:
+    if len(ndim_supp) != 1:
         raise NotImplementedError(
-            "Marginalization of withe dependent Multivariate RVs not implemented"
+            "Marginalization with dependent variables of different support not implemented"
```
jessegrabowski, line 592: What does "variables of different support" mean?

pymc_experimental/model/marginal_model.py
```diff
@@ -638,10 +654,29 @@ def get_domain_of_finite_discrete_rv(rv: TensorVariable) -> Tuple[int, ...]:
     elif isinstance(op, DiscreteUniform):
         lower, upper = constant_fold(rv.owner.inputs[3:])
         return tuple(range(lower, upper + 1))
+    elif isinstance(op, DiscreteMarkovChain):
+        p = rv.owner.inputs[0]
```
jessegrabowski, line 658: nitpick: capital `P` since it's a matrix

pymc_experimental/model/marginal_model.py
```diff
@@ -638,10 +654,29 @@ def get_domain_of_finite_discrete_rv(rv: TensorVariable) -> Tuple[int, ...]:
...
+        return tuple(range(pt.get_vector_length(p[-1])))
 
     raise NotImplementedError(f"Cannot compute domain for op {op}")
 
 
+def _add_reduce_batch_dependent_logps(
+    marginalized_type: TensorType, dependent_logps: Sequence[TensorVariable]
+):
+    """Add the logps of dependent RVs while reducing extra batch dims as assessed from the `marginalized_type`."""
```
jessegrabowski, line 667: "...while reducing extra batch dims inferred from the `marginalized_type`"

pymc_experimental/model/marginal_model.py
```diff
@@ -716,3 +747,69 @@ def logp_fn(marginalized_rv_const, *non_sequences):
...
+    batch_chain_value = pt.moveaxis(pt.full((*chain_shape, domain.size), domain), -1, 0)
+    batch_logp_emissions = vectorize_graph(reduced_logp_emissions, {chain_value: batch_chain_value})
+
+    # Step 2: Compute the transition probabilities
+    # This is the "forward algorithm", alpha_t = p(y | s_t) * sum_{s_{t-1}}(p(s_t | s_{t-1}) * alpha_{t-1})
+    # We do it entirely in logs, though.
+
+    # To compute the prior probabilities of each state, we evaluate the logp of the domain (all possible states) under
+    # the initial distribution. This is robust to everything the user can throw at it.
+    batch_logp_init_dist = pt.vectorize(lambda x: logp(init_dist_, x), "()->()")(
```
jessegrabowski, line 791: No way to avoid this lambda here with `vectorize_graph`? I recall this used to be a little function.

pymc_experimental/tests/model/test_marginal_model.py
```diff
@@ -598,3 +599,131 @@ def test_is_conditional_dependent_static_shape():
...
+        [10, 10],
+        [-10, 10],
+        [-10, 10],
+    ):
+        pt = {"norm": test_value}
+        np.testing.assert_allclose(logp_fn(pt), ref_logp_fn(pt))
+
+
+@pytest.mark.parametrize("batch_chain", (True,), ids=lambda x: f"batch_chain={x}")
+@pytest.mark.parametrize("batch_emission", (True,), ids=lambda x: f"batch_emission={x}")
```
jessegrabowski, line 649: These parameterizations have only one value -- are they needed?

jessegrabowski (approved):
I'm happy where it's landed. I wish we solved the two major problems (categorical emission and lags), but I'd rather have it merged and getting used than keep it in limbo while we wait for free time to make it perfect. I'll open an issue about those points after it's merged.

pymc_experimental/model/marginal_model.py
ricardoV94: Should be valid for batch dims

ricardoV94:
Categorical is one of the goals I have with #300

I think it's already working there, but I need to rebase and check one we merge this

ricardoV94:
The lags is a nice follow up. There current distribution doesn't have a clear API for lags and batch dims.

We just need to agree on this and then it should be straightforward to support both.

The design question is: how do you specify a markov chain with 2 lags and an extra batch dimension? Say something with shape (5, 100) with two lags but different transition matrixes for each of the five batched chains

jessegrabowski:
Yeah good questions. You're right it's not clear. I guess the distribution has to store the `n_lags` variable and `marginalize` will have to ask it? Not sure. In general, the way lags are handled are not good -- at higher orders the transition matrix is almost certainly going to be sparse, so it makes more sense to make one huge `k**n, k**n` sparse matrix and store a hash table to index into it for specific lag tuples.

pymc_experimental/model/marginal_model.py
ricardoV94: Can you open an issue for it?

pymc_experimental/tests/model/test_marginal_model.py
ricardoV94: Yeah, should remove. This was when I started testing batch dims and got confused about the API

pymc_experimental/model/marginal_model.py
```diff
@@ -245,16 +242,25 @@ def marginalize(
...
+                if rv_to_marginalize.owner.inputs[0].type.ndim > 2:
+                    raise NotImplementedError(
+                        "Marginalization for DiscreteMarkovChain with non-matrix transition probability is not supported"
+                    )
+            elif not isinstance(rv_op, (Bernoulli, Categorical, DiscreteUniform)):
                 raise NotImplementedError(
-                    f"RV with distribution {rv_to_marginalize.owner.op} cannot be marginalized. "
-                    f"Supported distribution include {supported_dists}"
+                    f"Marginalization of RV with distribution {rv_to_marginalize.owner.op} is not supported"
                 )
```
zaxtax, line 264: I thought the old error message was more helpful

zaxtax (approved):
I think issues should be open for some the marginalizations that are not implemented now as well. But otherwise, looks good to merge.

pymc_experimental/model/marginal_model.py
ricardoV94: It was but it not gonna scale

jessegrabowski: maybe link to the docs where it lists all the supported distributions then?

jessegrabowski:
What's left to do to get this merged?

pymc_experimental/model/marginal_model.py
ricardoV94: It's in a separate refactoring commit, you don't get to complain

pymc_experimental/model/marginal_model.py
ricardoV94: The way to avoid it is to be a little function, but seems like a fine use for lambda?

pymc_experimental/tests/model/test_marginal_model.py
ricardoV94: It was during debug I removed False. Reintroducing

pymc_experimental/model/marginal_model.py
ricardoV94: The notes state that functionality is restricted, only finite discrete RVs are supported which is kind of true. Although we don't yet support Truncated/Censored of infinite discrete RVs which thus become finite: #95 

We also don't support Multinomial which in theory is finite... So I think we the disclaimer functionality is restricted and this error message indicating the type of the RV that could not be marginalized it's fair game?

PR #271, opened 2023-11-23
add zerosum hierarchy

ferrine:
## Usecases

```
a + b - hierarchy without interactions, a, b - zerosum
a*b - hierarchy with interactions, a, b - zerosum, a_b - interaction effect with zerosum

(1 | a) + b - ??? should it be even considered?
```

## API

```python
with pm.Model(coords=dict(groups=range(10), stages=range(3))):
    mean = pm.Normal("mu")
    sigma = pm.Exponential("sigma")
    z = zerosum_hierarchy("z ~ groups + stages", importance=dict(groups=10), named=False)
    group_param = mean + z * sigma
```

### DIFF
--- a/pymc_experimental/tests/utils/test_hierarchy.py
+++ b/pymc_experimental/tests/utils/test_hierarchy.py
@@ -1,0 +1,60 @@
+import pymc as pm
+import pytest
+
+from pymc_experimental.utils.hierarchy import zerosum_hierarchy
+
+
+@pytest.fixture(autouse=True)
+def model():
+    with pm.Model() as model:
+        yield model
+
+
+def test_hierarchy_init(model):
+    model.add_coords(dict(a=range(2), b=range(3)))
+    var = zerosum_hierarchy("z0 ~ a", ("a",))
+    assert "z0::_a" in model.named_vars
+    assert "z0::weight" not in model.named_vars
+    var = zerosum_hierarchy(
+        "z1 ~ a+b",
+        ("a", "b"),
+        importance=dict(a=3),
+    )
+    assert "z1::_a" in model.named_vars
+    assert "z1::_b" in model.named_vars
+    assert "z1::weight" in model.named_vars
+    var = zerosum_hierarchy("z2 ~ a*b", ("a", "b"))
+    assert "z2::_a" in model.named_vars
+    assert "z2::_b" in model.named_vars
+    assert "z2::_a_b" in model.named_vars
+    assert "z2::weight" in model.named_vars
+
+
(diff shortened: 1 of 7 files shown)


pymc_experimental/utils/hierarchy.py
```diff
@@ -53,6 +53,17 @@ def zerosum_hierarchy(
     -------
     TensorVariable
         Resulting ZeroSum structured tensor with unit variance.
+
+    Examples
+    --------
+
+    .. code-block:: python
+
+        with pm.Model(coords=dict(groups=range(10), stages=range(3))):
```
ricardoV94, line 62: Nitpick: I like to include imports and fully reproducible snippet.

pymc_experimental/tests/utils/test_hierarchy.py
```diff
@@ -0,0 +1,60 @@
+import pymc as pm
+import pytest
+
+from pymc_experimental.utils.hierarchy import zerosum_hierarchy
+
+
+@pytest.fixture(autouse=True)
+def model():
+    with pm.Model() as model:
+        yield model
```
ricardoV94, line 10: Is this really needed?

pymc_experimental/utils/hierarchy.py
```diff
@@ -0,0 +1,128 @@
+from typing import Dict, List, Optional, Tuple, cast
+
+import formulae.terms
```
ricardoV94, line 3: Make import optional

requirements.txt
```diff
@@ -1,2 +1,3 @@
 pymc>=5.8.2
+formulae
```
ricardoV94, line 2: Make installation optional?

ferrine: How we structure the optional dependencies in pip?

ricardoV94: With setup.py? https://github.com/pymc-devs/pymc-experimental/blob/79c4dc13e9ea9a5cd6af6bc76f70f7b0c948af78/setup.py#L60-L65

ferrine: yep, I meant every time there is a new optional dependency, there is a new name here, easy to get lost

ricardoV94: Put it in `complete` only. I think we only need two. Otherwise people can manually install, because when they try to use a function we give them an informative message they need to install x

ferrine:
converted this PR to draft to just keep the progress

ferrine:
Meanwhile @aseyboldt suggested to use builder api, agreed with him

ricardoV94:
What's the builder API?

PR #272, opened 2023-11-24
Adapt PyMC 5.10.0 breaking changes

HasnainRaz:
Changing the name to reflect the renaming in https://github.com/pymc-devs/pymc/commit/04a03b520aa52c5586123f076c6377b32cf855b4

Otherwise importing this module raises an import error.

This need not be merged now, I am just making this PR to use the commit in my dependencies without relying on my own fork.

### DIFF
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -11,6 +11,6 @@ dependencies:
 - xhistogram
 - statsmodels
 - pip:
-  - pymc>=5.9.0  # CI was failing to resolve
+  - pymc>=5.10.0  # CI was failing to resolve
   - blackjax
   - scikit-learn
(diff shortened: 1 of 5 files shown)


ricardoV94:
@HasnainRaz did it fail or just raise a warning. The idea was that the old name was still importable but with a warning

ricardoV94:
Ahh this was importing from the right place: logprob submodule. The warning was added in the user facing one... makes sense maybe

HasnainRaz:
It was failing

ricardoV94:
We can now lower pin pymc to the `5.10.0` that was just released. That is the one with the breaking change here. The tests should start passing as well (we always test against the latest release)

HasnainRaz:
@ricardoV94 Seems the tests are failing because of a missing function in pymc. I lack the context to fix this, if you guys wanna push to this PR, feel free to do so

ricardoV94 (approved):

PR #276, opened 2023-12-02
Variationally Informed Parameterization

ferrine:
This PR adds a tool to autoparameterize your model in the following way: https://arxiv.org/abs/1906.03028

image

The intended api is like this

You need to convert tour model to change its structure

image

Then you use your samplers

image

For stability, you might want to turn on clipping for lambdas

image

### DIFF
--- a/pymc_experimental/model/transforms/autoreparam.py
+++ b/pymc_experimental/model/transforms/autoreparam.py
@@ -75,8 +75,33 @@ class VIP:
         )
 
 
-@singledispatch
 def vip_reparam_node(
+    op: RandomVariable,
+    node: Apply,
+    name: str,
+    dims: List[Variable],
+    transform: Transform,
+    eps: ModelNamed,
+    round: ModelNamed,
+) -> Tuple[ModelDeterministic, ModelNamed]:
+    if not isinstance(node.op, RandomVariable):
+        raise TypeError("Op should be RandomVariable type")
+    size = node.inputs[1]
+    if not isinstance(size, pt.TensorConstant):
+        raise ValueError("Size should be static for autoreparameterization.")
+    return _vip_reparam_node(
+        op,
+        node=node,
+        name=name,
+        dims=dims,
+        transform=transform,
+        eps=eps,
+        round=round,
+    )
+
+
+@singledispatch
+def _vip_reparam_node(
     op: RandomVariable,
     node: Apply,
     name: str,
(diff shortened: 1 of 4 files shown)


ferrine:
Anything else I should do in this PR?

ricardoV94:
Content looks good. Perhaps add an entry on the docs (we have to start introducing headers and sections though)

ferrine:
Ah, sure, docs are needed

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -0,0 +1,375 @@
...
+        scale**lam,
+        size=size,
+        rng=rng,
+    )
+    vip_rv_.name = f"{name}::tau_"
+
+    vip_rv = model_free_rv(
+        vip_rv_,
+        vip_rv_.clone(),
+        transform,
```
ferrine, line 232: @ricardoV94 should I worry about this?

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -0,0 +1,375 @@
...
+            name=name,
+            dims=dims,
+            transform=transform,
+            lam=lam,
+        ),
+        logit_lam,
+    )
+
+
+@singledispatch
```
ricardoV94, line 198: Seems a bit overkill at this point?

ferrine: I'd extend it to student T, Gamma, Logit Normal, etc

ricardoV94 (approved):

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -0,0 +1,376 @@
...
+    op: pm.Normal,
+    node: Apply,
+    name: str,
+    dims: List[Variable],
+    transform: Optional[Transform],
+    lam: pt.TensorVariable,
+) -> ModelDeterministic:
+    rng, size, _, loc, scale = node.inputs
+    if transform is not None:
+        raise NotImplementedError("Reparametrization with Transform is not implemented")
```
ricardoV94, line 221:
```suggestion
        raise NotImplementedError("Reparametrization of Normal with Transform is not implemented")
```

ricardoV94:
> @ricardoV94 this is failing for some unknown reason

I think a rewrite from the last PyTensor is leading to an error in the R2D2 prior tests. Either a rewrite bug or an issue with the R2D2. Will have to investigate

ferrine:
@ricardoV94 what are the next steps?

ricardoV94:
Need to check why the warning is being triggered and see if it can be safely ignored

ferrine:
it should be all good now

PR #277, opened 2023-12-04
Add function that caches sampling results

ricardoV94:
```python
import pymc as pm
from pymc_experimental.utils import cache_sampling

with pm.Model() as m:
    x = pm.Normal("x", 0, 1)
    y = pm.Normal("y", mu=x, observed=[0, 1, 2])

    idata = cache_sampling(pm.sample)()

with m:
    idata = cache_sampling(pm.sample)()  # Cache hit! Returning stored result
```


ricardoV94:
When rerunning notebooks

pymc_experimental/utils/cache.py
```diff
@@ -0,0 +1,174 @@
...
+        raise NotImplementedError(f"Hashing not implemented for type {c.type}")
+
+
+def get_name_and_props(obj):
+    name = str(obj)
+    props = str(getattr(obj, "_props", lambda: {})())
+    return name, props
+
+
+def hash_from_fg(fg: FunctionGraph) -> str:
```
ferrine, line 43: maybe to pytensor?

ricardoV94: Too experimental for now

ricardoV94:
I'm not familiar with MLflow, the idea here is that it pairs the saved traces to the exact model/sampling function (and arguments) that were used.

Basically the model and the function kwargs are the cache key.

Does this have any parallel to your workflow with MLflow?

pymc_experimental/utils/cache.py
```diff
@@ -0,0 +1,188 @@
...
+            arr = c.get_value(borrow=True)
+        arr_data = arr.view(np.uint8) if arr.size > 1 else arr.tobytes()
+        return hashlib.sha1(arr_data).hexdigest()
+    else:
+        raise NotImplementedError(f"Hashing not implemented for type {c.type}")
+
+
+def get_name_and_props(obj):
+    name = str(obj)
+    props = str(getattr(obj, "_props", lambda: {})())
```
lucianopaz, line 43:
```suggestion
    name = str(obj)
    if hasattr(obj, "_props"):
        prop_dict = obj._prop_dict()
        props = str(
            {k: get_name_and_props(v) for k, v in prop_dict.items()}
        )
    else:
        props = str({})
    
    name = str(obj)
    props = str(getattr(obj, "_props", lambda: {})())
```

lucianopaz: This is just to make sure that potential recursion of _props is handled.

pymc_experimental/utils/cache.py
```diff
@@ -0,0 +1,188 @@
...
+            idata_out: az.InferenceData = az.from_netcdf(file_path)
+            if force_load:
+                idata_out.load()
+
+        else:
+            idata_out = sampling_fn(*args, **kwargs, model=model, random_seed=random_seed)
+            if os.path.exists(file_path):
+                os.remove(file_path)
+            if not os.path.exists(dir):
+                os.mkdir(dir)
```
lucianopaz, line 177:
```suggestion
                os.makedirs(dir, exist_ok=True)
```

I think that it's better to use [`os.makedirs`](https://docs.python.org/3/library/os.html#os.makedirs) because it creates the intermediate directories if they are required.

pymc_experimental/utils/cache.py
```diff
@@ -0,0 +1,188 @@
...
+            az.to_netcdf(idata_out, file_path)
+
+        # We save inferencedata separately and extend if needed
+        if extend_inferencedata:
```
lucianopaz, line 181: This looks weird to me. The `sampling_fn` will go make up the hash. So if someone first calls `sample`, and then `sample_posterior_predictive` using extend, the second cache will include both the `posterior` and `posterior_predictive` groups; but the first cache will only include the `posterior` group. I think that it's cleaner to never allow for extending the idata inplace, and force users to combine the different `InferenceData` objects themselves.

lucianopaz (changes requested):

pymc_experimental/utils/cache.py
ricardoV94: Good catch

ricardoV94: Wait no, I don't see what is the problem. We only ever save new idatas coming out of th sampling_fn and these never extend the previous one.

pymc_experimental/utils/cache.py
ricardoV94: _props are not recursive

PR #278, opened 2023-12-11
Adding a generalised normal distribution

jbrinchmann:
This pull request implements a generalized Gaussian distribution (https://en.wikipedia.org/wiki/Generalized_normal_distribution) - this is useful for situations where the distribution is close to Gaussian but with some kurtosis. 

The distribution implemented is the symmetric version, but in keeping with the naming of the normal distribution I have not included this in the naming.

I have tried to follow the instructions for adding new distributions but for some reason when I do e.g.

```
In [1]: from pymc_experimental.distributions import GeneralizedNormal

In [2]: GeneralizedNormal.logcdf(1.5, mu=0.0, beta=1.3, alpha=0.5)
Out[2]: Log.0
```
which is an odd outcome since all intermediate steps are correct and the code works when embedded outside of `pymc-experimental`. As a consequence the tests do not appear to work.

### DIFF
--- a/pymc_experimental/distributions/continuous.py
+++ b/pymc_experimental/distributions/continuous.py
@@ -507,9 +507,12 @@ class GeneralizedNormal(Continuous):
         
         z = pt.abs((value-mu)/alpha)
         c =  pt.sign(value-mu)
-        cdf = 0.5 + pt.sign(value-mu)*pt.gammaincc(1.0/beta, pt.pow(z, beta))
 
-        print("CDF={0}".format(cdf))
+        # This follows the Wikipedia entry for the function - scipy has
+        # opted for a slightly different version using gammaincc instead.
+        # Note that on the Wikipedia page the unnormalised \gamma function
+        # is used, while gammainc is a normalised function
+        cdf = 0.5 + pt.sign(value-mu)*pt.gammainc(1.0/beta, pt.pow(z, beta))
         
         return pt.log(cdf)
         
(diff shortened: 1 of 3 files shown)


ricardoV94:
> which is an odd outcome since all intermediate

This is normal! PyMC uses PyTensor which return symbolic expressions until actually evaluated.

You can call `.eval()` on the output to get the numerical value. More details in https://pytensor.readthedocs.io/en/latest/tutorial/adding.html

pymc_experimental/distributions/continuous.py
```diff
@@ -351,3 +351,170 @@ def dist(cls, a, **kwargs):
...
+        mu = pt.as_tensor_variable(floatX(mu))
+        alpha = pt.as_tensor_variable(floatX(alpha))
+        beta = pt.as_tensor_variable(floatX(beta))
+
+        return super().dist([mu, alpha, beta], **kwargs)
+
+
+    # moment here returns the mean
+    def moment(rv, size, mu, alpha, beta):
+        moment, _ = pt.broadcast_arrays(mu, alpha, beta)
```
ricardoV94, line 473: Did you mean this?
```suggestion
        moment, *_ = pt.broadcast_arrays(mu, alpha, beta)
```

ricardoV94: Also missing the moment test

jbrinchmann:
Thanks for the feedback - in trying to get things working I had introduce two minor bugs in logp and logcdf. These are now fixed and double-checked to agree with scipy.stats.gennorm and a moment test has been added.

PR #285, opened 2023-12-15
Implement utility to recover marginalized variables from `MarginalModel`

### LINKED ISSUE #286: Add test for MarginalModel where variable depends on two marginalized variables
I am positively surprised this is working! Would probably be worth adding as a test.

We already have a test for when multiple RVs depend on a marginalized RV or when there are nested marginalized RVs. This will also create a nested marginalized RV even though x and y are independent a priori

```python
import pymc as pm
from pymc_experimental.model.marginal_model import MarginalModel
import numpy as np

with MarginalModel() as m:
    x = pm.Bernoulli("x", 0.1)
    y = pm.Bernoulli("y", 0.3)
    z = pm.DiracDelta("z", c=x+y)
    
m.marginalize([x, y])
logp = m.compile_logp()

np.testing.assert_allclose(np.exp(logp({"z": 0})), 0.9 * 0.7)
np.testing.assert_allclose(np.exp(logp({"z": 1})), 0.9 * 0.3 + 0.1 * 0.7)
np.testing.assert_allclose(np.exp(logp({"z": 2})), 0.1 * 0.3)
```

zaxtax:
This is a PR to add support for the `recover_marginals` method. This allows us to sample values and get access to the logps of discrete variables which we marginalized out during sampling.

Closes #286

### DIFF
--- a/pymc_experimental/model/marginal_model.py
+++ b/pymc_experimental/model/marginal_model.py
@@ -419,6 +419,9 @@ class MarginalModel(Model):
             joint_logps_norm = log_softmax(joint_logps, axis=0)
             if return_samples:
                 sample_rv_outs = pymc.Categorical.dist(logit_p=joint_logps)
+                if isinstance(rv.owner.op, DiscreteUniform):
+                    sample_rv_outs += rv_domain[0]
+
                 rv_loglike_fn = compile_pymc(
                     inputs=other_values,
                     outputs=[joint_logps_norm, sample_rv_outs],
--- a/pymc_experimental/model/marginal_model.py
+++ b/pymc_experimental/model/marginal_model.py
@@ -426,7 +426,7 @@ class MarginalModel(Model):
             joint_logps = pt.moveaxis(joint_logps, 0, -1)
 
             rv_loglike_fn = None
-            joint_logps_norm = log_softmax(joint_logps, axis=0)
+            joint_logps_norm = log_softmax(joint_logps, axis=-1)
             if return_samples:
                 sample_rv_outs = pymc.Categorical.dist(logit_p=joint_logps)
                 if isinstance(rv.owner.op, DiscreteUniform):
--- a/pymc_experimental/model/marginal_model.py
+++ b/pymc_experimental/model/marginal_model.py
@@ -84,6 +84,7 @@ class MarginalModel(Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.marginalized_rvs = []
+        self._marginalized_named_vars_to_dims = treedict()
 
     def _delete_rv_mappings(self, rv: TensorVariable) -> None:
         """Remove all model mappings referring to rv
@@ -231,6 +232,7 @@ class MarginalModel(Model):
         m.deterministics = [vars_to_clone[det] for det in self.deterministics]
 
         m.marginalized_rvs = [vars_to_clone[rv] for rv in self.marginalized_rvs]
+        m._marginalized_named_vars_to_dims = self._marginalized_named_vars_to_dims
         return m
 
     def marginalize(
@@ -254,6 +256,10 @@ class MarginalModel(Model):
                     f"RV with distribution {rv_to_marginalize.owner.op} cannot be marginalized. "
                     f"Supported distribution include {supported_dists}"
                 )
+
+            if rv_to_marginalize.name in self.named_vars_to_dims:
+                dims = self.named_vars_to_dims[rv_to_marginalize.name]
+                self._marginalized_named_vars_to_dims[rv_to_marginalize.name] = dims
 
             self._delete_rv_mappings(rv_to_marginalize)
             self.marginalized_rvs.append(rv_to_marginalize)
@@ -282,7 +288,11 @@ class MarginalModel(Model):
     def unmarginalize(self, rvs_to_unmarginalize):
         for rv in rvs_to_unmarginalize:
             self.marginalized_rvs.remove(rv)
-            self.register_rv(rv, name=rv.name)
+            if rv.name in self._marginalized_named_vars_to_dims:
+                dims = self._marginalized_named_vars_to_dims.pop(rv.name)
+            else:
+                dims = None
+            self.register_rv(rv, name=rv.name, dims=dims)
 
     def recover_marginals(
         self,
@@ -368,7 +378,7 @@ class MarginalModel(Model):
         posterior_pts = [transform_input(vs) for vs in posterior_pts]
 
         rv_dict = {}
-
+        rv_dims = {}
         for seed, rv in zip(seeds, vars_to_recover):
             supported_dists = (Bernoulli, Categorical, DiscreteUniform)
             if not isinstance(rv.owner.op, supported_dists):
@@ -453,8 +463,12 @@ class MarginalModel(Model):
             rv_dict["lp_" + rv.name] = logps.reshape(
                 tuple(len(coord) for coord in stacked_dims.values()) + logps.shape[1:],
             )
+            if rv.name in m.named_vars_to_dims:
+                rv_dims[rv.name] = list(m.named_vars_to_dims[rv.name])
+                rv_dims["lp_" + rv.name] = rv_dims[rv.name] + ["lp_" + rv.name + "_dim"]
 
         coords, dims = coords_and_dims_for_inferencedata(self)
+        dims.update(rv_dims)
         rv_dataset = dict_to_dataset(
             rv_dict,
             library=pymc,
--- a/pymc_experimental/model/marginal_model.py
+++ b/pymc_experimental/model/marginal_model.py
@@ -14,7 +14,7 @@ from pymc.logprob.basic import conditional_logp
 from pymc.logprob.transforms import IntervalTransform
 from pymc.model import Model
 from pymc.pytensorf import compile_pymc, constant_fold, inputvars
-from pymc.util import dataset_to_point_list, treedict
+from pymc.util import _get_seeds_per_chain, dataset_to_point_list, treedict
 from pytensor import Mode
 from pytensor.compile import SharedVariable
 from pytensor.compile.builders import OpFromGraph
@@ -233,9 +233,14 @@ class MarginalModel(Model):
         m.marginalized_rvs = [vars_to_clone[rv] for rv in self.marginalized_rvs]
         return m
 
-    def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorVariable]]):
+    def marginalize(
+        self, rvs_to_marginalize: Union[TensorVariable, str, Sequence[TensorVariable], Sequence[str]]
+    ):
         if not isinstance(rvs_to_marginalize, Sequence):
             rvs_to_marginalize = (rvs_to_marginalize,)
+        rvs_to_marginalize = [
+            self[var] if isinstance(var, str) else var for var in rvs_to_marginalize
+        ]
 
         supported_dists = (Bernoulli, Categorical, DiscreteUniform)
         for rv_to_marginalize in rvs_to_marginalize:
@@ -279,7 +284,12 @@ class MarginalModel(Model):
             self.register_rv(rv, name=rv.name)
 
     def recover_marginals(
-        self, idata, var_names=None, return_samples=True, extend_inferencedata=True
+        self,
+        idata,
+        var_names=None,
+        return_samples=True,
+        extend_inferencedata=True,
+        random_seed=None,
     ):
         """Computes posterior log-probabilities and samples of marginalized variables
         conditioned on parameters of the model given InferenceData with posterior group
@@ -299,6 +309,8 @@ class MarginalModel(Model):
             If True, also return samples of the marginalized variables
         extend_inferencedata : bool, default True
             Whether to extend the original InferenceData or return a new one
+        random_seed: int, array-like of int or SeedSequence, optional
+            Seed used to generating samples
 
         Returns
         -------
@@ -323,14 +335,18 @@ class MarginalModel(Model):
 
         """
         if var_names is None:
-            var_names = {var.name for var in self.marginalized_rvs}
-        else:
-            var_names = {var_names}
+            var_names = [var.name for var in self.marginalized_rvs]
 
+        var_names = [var if isinstance(var, str) else var.name for var in var_names]
         vars_to_recover = [v for v in self.marginalized_rvs if v.name in var_names]
-        missing_names = var_names.difference(v.name for v in vars_to_recover)
+        missing_names = [v.name for v in vars_to_recover if v not in self.marginalized_rvs]
         if missing_names:
             raise ValueError(f"Unrecognized var_names: {missing_names}")
+
+        if return_samples and random_seed is not None:
+            seeds = _get_seeds_per_chain(random_seed, len(vars_to_recover))
+        else:
+            seeds = [None] * len(vars_to_recover)
 
         posterior = idata.posterior
 
@@ -351,9 +367,8 @@ class MarginalModel(Model):
         posterior_pts = [transform_input(vs) for vs in posterior_pts]
 
         rv_dict = {}
-        rv_dims_dict = {}
 
-        for rv in vars_to_recover:
+        for seed, rv in zip(seeds, vars_to_recover):
             supported_dists = (Bernoulli, Categorical, DiscreteUniform)
             if not isinstance(rv.owner.op, supported_dists):
                 raise NotImplementedError(
@@ -365,7 +380,21 @@ class MarginalModel(Model):
             rv = m.vars_to_clone[rv]
             m.unmarginalize([rv])
             dependent_vars = find_conditional_dependent_rvs(rv, m.basic_RVs)
-            joint_logp = m.logp(vars=dependent_vars + [rv])
+            joint_logps = m.logp(vars=dependent_vars + [rv], sum=False)
+
+            marginalized_value = m.rvs_to_values[rv]
+            other_values = [v for v in m.value_vars if v is not marginalized_value]
+
+            # Handle batch dims for marginalized value and its dependent RVs
+            joint_logp = joint_logps[-1]
+            for dv in joint_logps[:-1]:
+                dbcast = dv.type.broadcastable
+                mbcast = marginalized_value.type.broadcastable
+                mbcast = (True,) * (len(dbcast) - len(mbcast)) + mbcast
+                values_axis_bcast = [
+                    i for i, (m, v) in enumerate(zip(mbcast, dbcast)) if m and not v
+                ]
+                joint_logp += dv.sum(values_axis_bcast)
 
             rv_shape = constant_fold(tuple(rv.shape))
             rv_domain = get_domain_of_finite_discrete_rv(rv)
@@ -379,27 +408,28 @@ class MarginalModel(Model):
                 0,
             )
 
-            marginalized_value = m.rvs_to_values[rv]
-            other_values = [v for v in m.value_vars if v is not marginalized_value]
-
             joint_logps = vectorize_graph(
                 joint_logp,
                 replace={marginalized_value: rv_domain_tensor},
             )
+            joint_logps = pt.moveaxis(joint_logps, 0, -1)
 
             rv_loglike_fn = None
+            joint_logps_norm = log_softmax(joint_logps, axis=0)
             if return_samples:
                 sample_rv_outs = pymc.Categorical.dist(logit_p=joint_logps)
                 rv_loglike_fn = compile_pymc(
                     inputs=other_values,
-                    outputs=[log_softmax(joint_logps, axis=0), sample_rv_outs],
+                    outputs=[joint_logps_norm, sample_rv_outs],
                     on_unused_input="ignore",
+                    random_seed=seed,
                 )
             else:
                 rv_loglike_fn = compile_pymc(
                     inputs=other_values,
-                    outputs=log_softmax(joint_logps, axis=0),
+                    outputs=joint_logps_norm,
                     on_unused_input="ignore",
+                    random_seed=seed,
                 )
 
             logvs = [rv_loglike_fn(**vs) for vs in posterior_pts]
@@ -409,18 +439,16 @@ class MarginalModel(Model):
             if return_samples:
                 logps, samples = zip(*logvs)
                 logps = np.array(logps)
-                rv_dict[rv.name] = np.reshape(
-                    samples, tuple(len(coord) for coord in stacked_dims.values())
+                samples = np.array(samples)
+                rv_dict[rv.name] = samples.reshape(
+                    tuple(len(coord) for coord in stacked_dims.values()) + samples.shape[1:],
                 )
-                rv_dims_dict[rv.name] = sample_dims
             else:
                 logps = np.array(logvs)
 
-            rv_dict["lp_" + rv.name] = np.reshape(
-                logps,
+            rv_dict["lp_" + rv.name] = logps.reshape(
                 tuple(len(coord) for coord in stacked_dims.values()) + logps.shape[1:],
             )
-            rv_dims_dict["lp_" + rv.name] = sample_dims + ("lp_" + rv.name + "_dims",)
 
         coords, dims = coords_and_dims_for_inferencedata(self)
         rv_dataset = dict_to_dataset(
@@ -433,8 +461,7 @@ class MarginalModel(Model):
         )
 
         if extend_inferencedata:
-            rv_dict = {k: (rv_dims_dict[k], v) for (k, v) in rv_dict.items()}
-            idata = idata.posterior.assign(**rv_dict)
+            idata.posterior = idata.posterior.assign(rv_dataset)
             return idata
         else:
             return rv_dataset
--- a/pymc_experimental/model/marginal_model.py
+++ b/pymc_experimental/model/marginal_model.py
@@ -234,7 +234,8 @@ class MarginalModel(Model):
         return m
 
     def marginalize(
-        self, rvs_to_marginalize: Union[TensorVariable, str, Sequence[TensorVariable], Sequence[str]]
+        self,
+        rvs_to_marginalize: Union[TensorVariable, str, Sequence[TensorVariable], Sequence[str]],
     ):
         if not isinstance(rvs_to_marginalize, Sequence):
             rvs_to_marginalize = (rvs_to_marginalize,)
--- a/pymc_experimental/model/marginal_model.py
+++ b/pymc_experimental/model/marginal_model.py
@@ -2,26 +2,36 @@ import warnings
 from typing import Sequence, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


pymc_experimental/tests/model/test_marginal_model.py
```diff
@@ -251,6 +256,26 @@ def test_marginalized_change_point_model_sampling(disaster_model):
     )
 
 
+@pytest.mark.slow
+@pytest.mark.filterwarnings("error")
+def test_unmarginalized_basic(disaster_model):
```
ricardoV94, line 261: I would like a simpler model where we can assert the actual logps are what's expected

ricardoV94 (commented):
Looks great, I will try and see how we can get transforms out of the way

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +248,127 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
         # Raise errors and warnings immediately
         self.clone()._marginalize(user_warnings=True)
 
+    def unmarginalize(
+        self, idata, var_names=None, include_samples=False, extend_inferencedata=True
+    ):
+        """Computes log-likelihoods of marginalized variables conditioned on parameters
```
ricardoV94, line 254: Does this handle multiple marginalized variables?

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +248,127 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+            other_values = [v for v in m.value_vars if v is not marginalized_value]
+
+            # TODO: Handle constants
+            # TODO: Handle transformed variables
+            joint_logp_op = OpFromGraph(
+                [marginalized_value] + other_values, [joint_logp], inline=True
+            )
+            joint_logps = [
+                joint_logp_op(rv_domain_tensor[i], *other_values) for i in range(len(rv_domain))
+            ]
```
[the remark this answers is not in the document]
zaxtax: What would the code look like for that?

pymc_experimental/model/marginal_model.py
zaxtax: Yes! Though no test yet for that

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +248,127 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+
+            rv_loglike_fn = None
+            if include_samples:
+                sample_rv_outs = pm.Categorical.dist(logit_p=joint_logps)
```
ricardoV94, line 322: Are these `joint_logps` normalized? `pm.Categorical` won't do it under the hood

pymc_experimental/model/marginal_model.py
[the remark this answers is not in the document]
ricardoV94: Something like `joint_logps = vectorize_graph(joint_logp, replace={marginalized_value: rv_domain_tensor}`

pymc_experimental/model/marginal_model.py
ricardoV94: Can you add some info on the docstrings saying how you're handling. Sounds like you are doing one variable at a time evaluation while keeping the remaining ones marginalized?

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +248,127 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        Returns
+        -------
+        idata : InferenceData
+            InferenceData with var_names added to posterior
+
+        """
+        if var_names is None:
+            var_names = self.marginalized_rvs
+
+        joint_logp = self.logp()
```
ricardoV94, line 277: `self.logp` will return the logp with the variables marginalized so they won't be part of the graph. I guess that's why you have `on_unused_inputs` issues later? I imagine you want the original logp that does not marginalize variables so you can give them values. 

That's why I was asking how you handle multiple related marginalized variables, seems like you have to either evaluate one at a time conditioned on the previous variables already evaluated or create a joint logp for all the combinations of the marginalized variables.

ricardoV94: Here is a test that creates nested marginalized variables: https://github.com/pymc-devs/pymc-experimental/blob/8046695e600970bb30a107376281d3e477a66dd0/pymc_experimental/tests/model/test_marginal_model.py#L169-L204

They get represented as an OpFromGraph where there are more than one output RVs without values. You can raise NotImplementedError for these cases for now so that you know you're working only with independent simple marginalized variables. Still I think you don't want to work with `self.logp`

ricardoV94: `self._logp` skips the marginalization step, maybe that's what you need?

pymc_experimental/model/marginal_model.py
zaxtax: Yes that's exactly what I do. I sample each discrete variable while the rest of them marginalized

pymc_experimental/model/marginal_model.py
zaxtax: I chose to marginalise each discrete variable one at a time. Presumably I still need marginalise for that reason

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +248,130 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
         self.clone()._marginalize(user_warnings=True)
 
+    def unmarginalize(
+        self, idata, var_names=None, include_samples=False, extend_inferencedata=True
+    ):
+        """Computes log-likelihoods of marginalized variables conditioned on parameters
+        of the model given InferenceData with posterior group
+
+        When there are multiple marginalized variables, each marginalized variable is
+        conditioned on both the parameters and the other marginalized variables
```
ricardoV94, line 258:
```suggestion
        conditioned on both the parameters and the other variables still marginalized
```

pymc_experimental/model/marginal_model.py
ricardoV94: It will when you pass `logit_p`, but the `logit_p` added to the inferencedata will still not be normalized. I think it may be more intuitive if they are but not sure.

pymc_experimental/model/marginal_model.py
```diff
@@ -248,14 +248,20 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
         # Raise errors and warnings immediately
         self.clone()._marginalize(user_warnings=True)
 
-    def unmarginalize(
+    def unmarginalize(self, rvs_to_unmarginalize):
+        for r in rvs_to_unmarginalize:
+            self.register_rv(rv, name=rv.name)
+
+        self.marginalized_rvs = list(set(self.marginalized_rvs) - set(rvs_to_unmargalize))
```
ricardoV94, line 255: Slight preference because it won't alter the order of the remaining marginalized_rvs. You had a typo though: for `r` instead of for `rv`

```suggestion
        for rv in rvs_to_unmarginalize:
            rv = self.marginalized_rvs.pop(rv)
            self.register_rv(rv, name=rv.name)
```

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +249,134 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
         # Raise errors and warnings immediately
         self.clone()._marginalize(user_warnings=True)
 
+    def unmarginalize(self, rvs_to_unmarginalize):
+        for rv in rvs_to_unmarginalize:
+            self.marginalized_rvs.remove(rv)
+            self.register_rv(rv, name=rv.name)
+
+    def recover_marginals(
+        self, idata, var_names=None, include_samples=False, extend_inferencedata=True
```
ricardoV94, line 258: I would perhaps default `include_samples=True` by default. Also maybe a different name? `return_samples` or just `sample`?

pymc_experimental/tests/model/test_marginal_model.py
```diff
@@ -251,6 +251,26 @@ def test_marginalized_change_point_model_sampling(disaster_model):
     )
 
 
+@pytest.mark.slow
+@pytest.mark.filterwarnings("error")
```
ricardoV94, line 255: I think this is the default for the whole test suite so you don't need to add

pymc_experimental/model/marginal_model.py
zaxtax: The discrete samples are often not as good for understanding the tail probabilities

This can be seen by comparing the changepoint's logps vs the discrete samples discussed on https://mc-stan.org/docs/stan-users-guide/change-point.html

ricardoV94: I know that but I still think it doesn't hurt to include by default. logits are not something many users grok intuitively

ricardoV94 (commented):
Looks pretty good, I left 2 comments. 

I would a test with multiple marginalized dependent variables and after that I think this would be pretty much there (except for the transforms ofc)

pymc_experimental/model/marginal_model.py
zaxtax: Updated name

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,134 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+
+            joint_logps = vectorize_graph(
+                joint_logp,
+                replace={marginalized_value: rv_domain_tensor},
+            )
+
+            rv_loglike_fn = None
+            if include_samples:
+                sample_rv_outs = pymc.Categorical.dist(logit_p=joint_logps)
+                rv_loglike_fn = compile_pymc(
```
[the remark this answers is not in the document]
ricardoV94: I still think this is worthwhile exploring

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,162 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        fn = self.compile_fn(inputs=self.free_RVs, outs=transformed_rvs)
+        return fn, transformed_names
+
+    def unmarginalize(self, rvs_to_unmarginalize):
+        for rv in rvs_to_unmarginalize:
+            self.marginalized_rvs.remove(rv)
+            self.register_rv(rv, name=rv.name)
+
+    def recover_marginals(
+        self, idata, var_names=None, return_samples=False, extend_inferencedata=True
```
ricardoV94, line 282: I still think we should return samples by default.

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,162 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+            return dict(zip(transform_names, transform_fn(inputs)))
+
+        posterior_pts = [transform_input(vs) for vs in posterior_pts]
+
+        rv_dict = {}
+        rv_dims_dict = {}
+
+        for rv in var_names:
+            m = self.clone()
+            rv = m.vars_to_clone[rv]
```
ricardoV94, line 335: Add a check and raise NotImplementedError if this RV is not one of the supported cases. For instance when we add MarginalHMM the logic here won't suffice to recover the variables.

ricardoV94 (commented):
This looks pretty good 😊

Just two comments plus the question of compiling a single function for when we have multiple marginalized RVs. I am not sure the speed benefits outweigh the extra complexity at this point so fine to leave as is

pymc_experimental/model/marginal_model.py
zaxtax: How do I check the supported cases?

ricardoV94:
Ah one reason I see for why we may want to normalize the lps is that we actually don't need to evaluate the joint logp of the whole model, but only those variables that depend on the marginalized one.

In the future we may want to be more efficient and compile a logp with `vars=[marginalized, **dependent_RVs]` and the unnormalized LPs don't make as much sense there. In contrast, the normalized one should be exactly the same.

zaxtax:
> Ah one reason I see for why we may want to normalize the lps is that we actually don't need to evaluate the joint logp of the whole model, but only those variables that depend on the marginalized one.
> 
> In the future we may want to be more efficient and compile a logp with `vars=[marginalized, *dependent_RVs]` and the unnormalized lps don't make as much sense there. In contrast, the normalized lps should come out exactly the same.

Yep, I'm convinced. Will make the changes

zaxtax:
I think in the future we can include the optimisations where we compile for joint_logps all at once. As well as only using a logp that includes terms which contain `marginalized_value`.

But this should work for now

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        All log-probabilities are within the transformed space
+
+        Parameters
+        ----------
+        idata : InferenceData
+            InferenceData with posterior group
+        var_names : sequence of str, optional
+            List of Observed variable names for which to compute log_likelihood. Defaults to all observed variables
+        return_samples : bool, default True
+            If True, also return samples of the marginalized variables
```
ricardoV94, line 299: Docstrings are wrong. Also would be nice to add a code example as we do for other methods

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        extend_inferencedata : bool, default True
+            Whether to extend the original InferenceData or return a new one
+
+        Returns
+        -------
+        idata : InferenceData
+            InferenceData with var_names added to posterior
```
ricardoV94, line 306: Would be good to emphasize lps will be called `lp_{varname}` and be found in the posterior group, same for samples.

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+                    inputs=other_values,
+                    outputs=joint_logps,
+                    on_unused_input="ignore",
+                )
+
+            logvs = [rv_loglike_fn(**vs) for vs in posterior_pts]
+
+            if return_samples:
+                logps, samples = zip(*logvs)
+                logps = np.array(logps)
```
ricardoV94, line 387: This shouldn't be needed?
```suggestion
```

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+            else:
+                logps = np.array(logvs)
+                rv_dict["lp_" + rv.name] = log_softmax(
+                    np.reshape(
+                        logps,
+                        tuple(len(coord) for coord in stacked_dims.values()) + logps.shape[1:],
+                    ),
+                    axis=len(stacked_dims),
+                )
+                rv_dims_dict["lp_" + rv.name] = sample_dims + ("lp_" + rv.name + "_dims",)
```
ricardoV94, line 409: Since this is done in both branches, move it out and write only once?

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+            rv_shape = constant_fold(tuple(rv.shape))
+            rv_domain = get_domain_of_finite_discrete_rv(rv)
+            rv_domain_tensor = pt.swapaxes(
+                pt.full(
+                    (*rv_shape, len(rv_domain)),
+                    rv_domain,
+                    dtype=rv.dtype,
+                ),
+                axis1=0,
+                axis2=-1,
```
ricardoV94, line 355: Do you mean perhaps `moveaxis(..., -1, 0)`? This will fail if `rv_shape` is larger than 1 no?

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+            )
+
+            marginalized_value = m.rvs_to_values[rv]
+
+            other_values = [v for v in m.value_vars if v is not marginalized_value]
+
+            # TODO: Handle constants
```
ricardoV94, line 362:
```suggestion
```

pymc_experimental/model/marginal_model.py
ricardoV94: If that was a case we should add a test as well

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        conditioned on both the parameters and the other variables still marginalized
+
+        All log-probabilities are within the transformed space
+
+        Parameters
+        ----------
+        idata : InferenceData
+            InferenceData with posterior group
+        var_names : sequence of str, optional
+            List of Observed variable names for which to compute log_likelihood. Defaults to all observed variables
```
ricardoV94, line 297: This is wrong, it's a log-probability right? And the default is "all marginalized variables"

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,176 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        extend_inferencedata : bool, default True
+            Whether to extend the original InferenceData or return a new one
+
+        Returns
+        -------
+        idata : InferenceData
+            InferenceData with var_names added to posterior
+
+        """
+        if var_names is None:
```
ricardoV94, line 309: Name is misleading, because it's not a string you can pass but the RV itself?

Also should add a check that it actually corresponds to a marginalized_rv?

ricardoV94 (commented):
Small comments, hopefully that's all on my end :)

pymc_experimental/model/marginal_model.py
zaxtax: This adopts logic from `replace_finite_discrete_marginal_subgraph` is there a reason this might break and that won't?

ricardoV94: I think you mean the `finite_discrete_marginal_rv_logp` and I think it's a bug there as well

pymc_experimental/model/marginal_model.py
zaxtax: Because we use a list comprehension over the samples in idea we have a list not a numpy array. I don't think there is a nice way to get around this.

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,192 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+
+    def unmarginalize(self, rvs_to_unmarginalize):
+        for rv in rvs_to_unmarginalize:
+            self.marginalized_rvs.remove(rv)
+            self.register_rv(rv, name=rv.name)
+
+    def recover_marginals(
+        self, idata, var_names=None, return_samples=True, extend_inferencedata=True
+    ):
+        """Computes normalized posterior probabilities of marginalized variables
```
ricardoV94, line 284: The only reason I took the normalized was to reuse the posterior word, what is more important was to mention it also samples

```suggestion
        """Computes posterior log-probabilities and samples of marginalized variables
```

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,192 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        conditioned on both the parameters and the other variables still marginalized
+
+        All log-probabilities are within the transformed space
+
+        Parameters
+        ----------
+        idata : InferenceData
+            InferenceData with posterior group
+        var_names : sequence of str, optional
+            List of Observed variable names for which to compute posterior probabilities. Defaults to all marginalized variables
```
ricardoV94, line 297:
```suggestion
            List of variable names for which to compute posterior log-probabilities and samples. Defaults to all marginalized variables
```

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,192 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+        return_samples : bool, default True
+            If True, also return samples of the marginalized variables
+        extend_inferencedata : bool, default True
+            Whether to extend the original InferenceData or return a new one
+
+        Returns
+        -------
+        idata : InferenceData
+            InferenceData with where a lp_{varname} and {varname} for each marginalized variable in var_names added to posterior
```
ricardoV94, line 306:
```suggestion
            InferenceData with where a lp_{varname} and {varname} for each marginalized variable in var_names added to the posterior group
```

pymc_experimental/tests/model/test_marginal_model.py
```diff
@@ -251,6 +268,125 @@ def test_marginalized_change_point_model_sampling(disaster_model):
...
+        true_logp(idata.y.values.flatten(), idata.sigma.values.flatten()),
+        idata.lp_k[0].values,
+    )
+
+
+def test_recover_batched_marginal():
+    """Test that marginalization works for batched random variables"""
+    with MarginalModel() as m:
+        sigma = pm.HalfNormal("sigma")
+        idx = pm.Bernoulli("idx", p=0.7, shape=(2, 2))
```
ricardoV94, line 319: Here I would give a different length to each dim and check the come out correctly in the idata

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +255,192 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+                    dtype=rv.dtype,
+                ),
+                -1,
+                0,
+            )
+
+            marginalized_value = m.rvs_to_values[rv]
+            other_values = [v for v in m.value_vars if v is not marginalized_value]
+
+            joint_logps = vectorize_graph(
```
ricardoV94, line 384: I don't see the normalization anywhere anymore?

pymc_experimental/model/marginal_model.py
ricardoV94: You're right

pymc_experimental/tests/model/test_marginal_model.py
ricardoV94: I nice feature (no need to block this PR, can be a follow up issue) would be to reuse dims in the computed variables if the user specified dims for the marginalized variables

pymc_experimental/model/marginal_model.py
ricardoV94: Ah you're doing it with scipy? Any reason not to do it with Pytensor with the rest of the computations?

ricardoV94:
As a follow up we may want to standardize the signature of marginalize and recover marginals to allow passing strings or the variables in either case. Right now each is restricted to a different type which feels suboptimal

pymc_experimental/model/marginal_model.py
```diff
@@ -222,9 +233,14 @@ def clone(self):
         m.marginalized_rvs = [vars_to_clone[rv] for rv in self.marginalized_rvs]
         return m
 
-    def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorVariable]]):
+    def marginalize(
+        self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorVariable], Sequence[str]]
```
ricardoV94, line 237:
```suggestion
        self, rvs_to_marginalize: Union[TensorVariable, str, Sequence[TensorVariable], Sequence[str]]
```

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +260,202 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+                m.recover_marginals(idata, var_names=["x"])
+
+
+        """
+        if var_names is None:
+            var_names = {var.name for var in self.marginalized_rvs}
+        else:
+            var_names = {var_names}
+
+        var_names = {var if isinstance(var, str) else var.name for var in var_names}
```
ricardoV94, line 335: One reason I don't like sets is they introduce randomness all over the place. 

This made me realize we should allow users to pass a seed and split it for each of the `compile_pymc` when we are sampling the Categorical (there's a `get_seeds_per_chain` utility in PyMC).

But even with a seed the draws will be different depending on the order we end up creating the functions due to this set

pymc_experimental/model/marginal_model.py
```diff
@@ -244,6 +260,202 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+                replace={marginalized_value: rv_domain_tensor},
+            )
+            joint_logps = pt.moveaxis(joint_logps, 0, -1)
+
+            rv_loglike_fn = None
+            if return_samples:
+                sample_rv_outs = pymc.Categorical.dist(logit_p=joint_logps)
+                rv_loglike_fn = compile_pymc(
+                    inputs=other_values,
+                    outputs=[log_softmax(joint_logps, axis=0), sample_rv_outs],
```
ricardoV94, line 413: nitpick, move the repeated log_softmax before the if/else

pymc_experimental/model/marginal_model.py
```diff
@@ -238,12 +257,233 @@ def marginalize(self, rvs_to_marginalize: Union[TensorVariable, Sequence[TensorV
...
+            )
+
+            joint_logps = vectorize_graph(
+                joint_logp,
+                replace={marginalized_value: rv_domain_tensor},
+            )
+            joint_logps = pt.moveaxis(joint_logps, 0, -1)
+
+            rv_loglike_fn = None
+            joint_logps_norm = log_softmax(joint_logps, axis=0)
```
ricardoV94, line 429: Shouldn't this the be last axis now?

pymc_experimental/tests/model/test_marginal_model.py
```diff
@@ -251,6 +268,158 @@ def test_marginalized_change_point_model_sampling(disaster_model):
...
+        sigma = sigma.repeat(len(p)).reshape(len(sigma), -1)
+        return log_softmax(
+            np.log(p)
+            + norm.logpdf(y, loc=mu, scale=sigma)
+            + halfnorm.logpdf(sigma)
+            + np.log(sigma),
+            axis=1,
+        )
+
+    np.testing.assert_almost_equal(
```
ricardoV94, line 310: Can you also add a sanity check assert that logsumexp(lps) is close to 0?

ricardoV94 (commented):
Small confusion, otherwise everything looks ready

ricardoV94 (approved):

PR #294, opened 2024-01-01
Implement Generalized Pareto distribution

xieyj17:
Generalized Pareto distribution is a commonly used distribution for modelling the tail of another distribution. It has wide applications in risk management, finance, ans quality assuarance. See [wiki](https://en.wikipedia.org/wiki/Generalized_Pareto_distribution) page. 

I added a new distribution to the pymc-experimental branch.


xieyj17:
When I am rebasing, I accidently added the recent PR by @zaxtax .

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,163 @@ def moment(rv, size, mu, sigma, xi):
         return mode
 
 
+# Generalized Pareto Distribution
+class GenParetoRV(RandomVariable):
+    name: str = "Generalized Pareto Distribution"
```
ricardoV94, line 226:
```suggestion
    name: str = "Generalized Pareto"
```

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,163 @@ def moment(rv, size, mu, sigma, xi):
         return mode
 
 
+# Generalized Pareto Distribution
+class GenParetoRV(RandomVariable):
```
ricardoV94, line 225: Should subclass `ScipyRandomVariable` because Scipy RVs (sometimes) do something dumb with `size=(1,)`

```suggestion
class GenParetoRV(ScipyRandomVariable):
```

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,163 @@ def moment(rv, size, mu, sigma, xi):
...
+    name: str = "Generalized Pareto Distribution"
+    ndim_supp: int = 0
+    ndims_params: List[int] = [0, 0, 0]
+    dtype: str = "floatX"
+    _print_name: Tuple[str, str] = ("Generalized Pareto Distribution", "\\operatorname{GP}")
+
+    def __call__(self, mu=0.0, sigma=1.0, xi=1.0, size=None, **kwargs) -> TensorVariable:
+        return super().__call__(mu, sigma, xi, size=size, **kwargs)
```
ricardoV94, line 233: This is not strictly necessary because most users will never call the RV directly. We usually provide default values through the PyMC distribution class

```suggestion
```

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,163 @@ def moment(rv, size, mu, sigma, xi):
...
+        mu: np.ndarray,
+        sigma: np.ndarray,
+        xi: np.ndarray,
+        size: Tuple[int, ...],
+    ) -> np.ndarray:
+        # using scipy's parameterization
+        return stats.genpareto.rvs(c=xi, loc=mu, scale=sigma, random_state=rng, size=size)
+
+
+gp = GenParetoRV()
```
ricardoV94, line 248:
```suggestion
gen_pareto = GenParetoRV()
```

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,163 @@ def moment(rv, size, mu, sigma, xi):
         return mode
 
 
+# Generalized Pareto Distribution
+class GenParetoRV(RandomVariable):
+    name: str = "Generalized Pareto Distribution"
+    ndim_supp: int = 0
+    ndims_params: List[int] = [0, 0, 0]
+    dtype: str = "floatX"
+    _print_name: Tuple[str, str] = ("Generalized Pareto Distribution", "\\operatorname{GP}")
```
ricardoV94, line 230:
```suggestion
    _print_name: Tuple[str, str] = ("Generalized Pareto", "\\operatorname{GenPareto}")
```

pymc_experimental/tests/distributions/test_continuous.py
```diff
@@ -138,6 +138,64 @@ class TestGenExtreme(BaseTestDistributionRandom):
...
+    """
+    Wrapper class so that tests of experimental additions can be dropped into
+    PyMC directly on adoption.
+
+    pm.logp(GenPareto.dist(mu=0.,sigma=1.,xi=5.),value=1.)
+    """
+
+    def test_logp(self):
+        def ref_logp(value, mu, sigma, xi):
+            if xi == 0:
```
ricardoV94, line 151: Scipy genpareto logpdf fails for xi = 0?

pymc_experimental/tests/distributions/test_continuous.py
```diff
@@ -138,6 +138,64 @@ class TestGenExtreme(BaseTestDistributionRandom):
...
+            else:
+                return sp.genpareto.logpdf(value, c=xi, loc=mu, scale=sigma)
+
+        check_logp(
+            GenPareto,
+            R,
+            {"mu": R, "sigma": Rplusbig, "xi": Rplusbig},
+            ref_logp,
+            decimal=select_by_precision(float64=6, float32=2),
+            skip_paramdomain_outside_edge_test=True,
```
ricardoV94, line 162: Why are you skipping the outside edge test?

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,163 @@ def moment(rv, size, mu, sigma, xi):
...
+    ========  =========================================================================
+
+    Parameters
+    ----------
+    mu : float
+        Location parameter.
+    sigma : float
+        Scale parameter (sigma > 0).
+    xi : float
+        Shape parameter (xi >= 0)
```
ricardoV94, line 305: Could be worth a note saying this is more restrictive than other definitions of the GenPareto (in wikipedia there seems to be special cases for xi < 0?)

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,163 @@ def moment(rv, size, mu, sigma, xi):
...
+            pt.log(1 - pt.exp(-1 * scaled)),
+            pt.log(1 - pt.pow((1 + xi * scaled), (-1 / xi))),
+        )
+        logc = pt.switch(pt.ge(scaled, 0), logc_expression, -np.inf)
+
+        return check_parameters(logc, sigma > 0, xi >= 0, msg="sigma > 0 and xi >= 0")
+
+    def moment(rv, size, mu, sigma, xi):
+        r"""
+        Mean is defined when :math:`\xi < 1`
```
ricardoV94, line 372: We don't need to provide a real "moment", just anything that always has finite logp. So in this case `moment = mu` may be good enough?

ricardoV94 (commented):
Thanks for opening up the PR. I left some comments and questions.

For the rebase, what did you try to do exactly?

ricardoV94:
You should add an entry in the docs API: https://github.com/pymc-devs/pymc-experimental/blob/main/docs/api_reference.rst

pymc_experimental/tests/distributions/test_continuous.py
xieyj17: Yes, I do noticed a tiny bug in scipy's function when calculating pdf for general pareto distribution with xi==0. Will double check and sumbit a PR to fix that as well.

pymc_experimental/tests/distributions/test_continuous.py
xieyj17: Since I am bounding the xi to be >= 0, I'd like to skip the outside edge test.

pymc_experimental/distributions/continuous.py
xieyj17: xi < 0 are less seen for modelling extreme values. I will add a note here.

xieyj17: xi < 0 are less seen for modelling extreme values. I will add a note here.

pymc_experimental/distributions/continuous.py
xieyj17: Are you suggesting that we only need to return `mu` instead of the true mean? Or shall I just leave it as it?

xieyj17:
@ricardoV94  Thanks for you comments! I have addressed your suggested changes. Thank you!

pymc_experimental/tests/distributions/test_continuous.py
ricardoV94: The point of that test is to make sure the bounding is defined correctly, so you shouldn't skip

pymc_experimental/distributions/continuous.py
ricardoV94: Yes, tha tyou can just return `mu` and take away the restriction

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,160 @@ def moment(rv, size, mu, sigma, xi):
...
+        Parameters
+        ----------
+        value: numeric
+            Value(s) for which log-probability is calculated. If the log probabilities for multiple
+            values are desired the values must be provided in a numpy array or Pytensor tensor
+
+        Returns
+        -------
+        TensorVariable
+        """
```
ricardoV94, line 329: These docstrings are incomplet, and the `logp` function is not really user facing, so it's better to not include

pymc_experimental/distributions/continuous.py
```diff
@@ -221,6 +221,160 @@ def moment(rv, size, mu, sigma, xi):
...
+        ----------
+        value: numeric or np.ndarray or `TensorVariable`
+            Value(s) for which log CDF is calculated. If the log CDF for
+            multiple values are desired the values must be provided in a numpy
+            array or `TensorVariable`.
+
+        Returns
+        -------
+        TensorVariable
+        """
```
ricardoV94, line 356: Same here

pymc_experimental/tests/distributions/test_continuous.py
```diff
@@ -138,6 +138,64 @@ class TestGenExtreme(BaseTestDistributionRandom):
     ]
 
 
+class TestGenParetoClass:
```
ricardoV94, line 141: Missing the test for moment

ricardoV94 (commented):
A few more comments, thanks for the work so far!

rxd199682:
I think the logp for Generalized Pareto distribution should be     def logp(value, mu, sigma, xi):
        """
        Calculate log-probability of Generalized Pareto distribution
        at specified value.

        Parameters
        ----------
        value: numeric
            Value(s) for which log-probability is calculated. If the log probabilities for multiple
            values are desired the values must be provided in a numpy array or Pytensor tensor

        Returns
        -------
        TensorVariable
        """

        scaled = (value - mu) / sigma

        logp_expression = pt.switch(
            pt.isclose(xi, 0),
            -1 * scaled,
            -1 * pt.log(sigma) - ((xi + 1) / xi) * pt.log1p(xi * scaled),
        )
        logp = pt.switch(pt.gt(1 + xi * scaled, 0), logp_expression, -np.inf)
        return check_parameters(logp, sigma > 0, pt.and_(xi > -1, xi < 1), msg="sigma > 0 or -1 < xi < 1")

pymc_experimental/tests/distributions/test_continuous.py
rxd199682: I think the logp for Generalized Pareto distribution should be
我认为广义帕累托分布的 logp 应该是
def logp(value, mu, sigma, xi):
"""
Calculate log-probability of Generalized Pareto distribution
计算广义帕累托分布的对数概率
at specified value.  在指定值。

    Parameters
    ----------
    value: numeric
        Value(s) for which log-probability is calculated. If the log probabilities for multiple
        values are desired the values must be provided in a numpy array or Pytensor tensor

    Returns
    -------
    TensorVariable
    """

    scaled = (value - mu) / sigma

    logp_expression = pt.switch(
        pt.isclose(xi, 0),
        -1 * scaled,
        -1 * pt.log(sigma) - ((xi + 1) / xi) * pt.log1p(xi * scaled),
    )
    logp = pt.switch(pt.gt(1 + xi * scaled, 0), logp_expression, -np.inf)
    return check_parameters(logp, sigma > 0, pt.and_(xi > -1, xi < 1), msg="sigma > 0 or -1 < xi < 1")

pymc_experimental/distributions/continuous.py
rxd199682: I think the logp for Generalized Pareto distribution should be
我认为广义帕累托分布的 logp 应该是
def logp(value, mu, sigma, xi):
"""
Calculate log-probability of Generalized Pareto distribution
计算广义帕累托分布的对数概率
at specified value.  在指定值。

    Parameters
    ----------
    value: numeric
        Value(s) for which log-probability is calculated. If the log probabilities for multiple
        values are desired the values must be provided in a numpy array or Pytensor tensor

    Returns
    -------
    TensorVariable
    """

    scaled = (value - mu) / sigma

    logp_expression = pt.switch(
        pt.isclose(xi, 0),
        -1 * scaled,
        -1 * pt.log(sigma) - ((xi + 1) / xi) * pt.log1p(xi * scaled),
    )
    logp = pt.switch(pt.gt(1 + xi * scaled, 0), logp_expression, -np.inf)
    return check_parameters(logp, sigma > 0, pt.and_(xi > -1, xi < 1), msg="sigma > 0 or -1 < xi < 1")

ricardoV94:
Closing in favor of #689

PR #300, opened 2024-01-28
Expand models supported by automatic marginalization

ricardoV94:
This PR allows more kinds of graphs to be marginalized. Previously, we were limiting it to Elemwise operations to ensure that information across batch dimensions was not mixed between the marginalized and dependent RVs, so as to generate an efficient logp expression that did not grow with number of batch dimensions, but was constant on the domain of the marginalized variables.

We still have the same constraint, but can now analyze it more carefully by propagating information about the batch dimensions of the marginalized RV across the intermediate operations. This allows operations like DimShuffle (transposition, expand_dims, squeeze), Blockwise, Reductions and, rather important, flavors of basic and advanced indexing like the following:

```python
import pymc as pm
import pytensor.tensor as pt

from pymc_experimental import MarginalModel

with MarginalModel() as m:
    state = pm.Categorical("state", p=[0.1, 0.3, 0.6], shape=(4,))
    # Advanced indexing was not supported before
    # The indexed variable could be an RV as well!
    mu = pt.as_tensor([-10, 0, 10])[state]
    sigma = pm.HalfNormal("sigma")
    emission = pm.Normal("emission", mu, sigma, observed=[-9.0, -0.5, 1.0, 11.0])
    
m.marginalize(state)
m.point_logps()
# {'sigma': -0.73, 'emission': -10.52}
```

This should expand the range of models supported and open room for further expansions.

## TODO:
- [x] Give more informative error for explicit broadcastable dims
- [x] Document internal logic more carefully
- [x] Allow dependent multivariate RVs (it should almost work out of the box, just need some tweaks and tests)
- [x] Test `subgrah_dims` utility directly
- [x] Test MvNormal k-clusters model that showed up in Discourse some time ago
- [x] Test dependent RVs batch to the left restriction
- [x] Allow a couple more simple Ops


pymc_experimental/model/marginal_model.py
```diff
@@ -586,7 +825,7 @@ def replace_finite_discrete_marginal_subgraph(fgraph, rv_to_marginalize, all_rvs
         )
 
     marginalized_rv_input_rvs = find_conditional_input_rvs([rv_to_marginalize], all_rvs)
-    dependent_rvs_input_rvs = [
+    other_direct_rv_ancestors = [
```
zaxtax, line 828: What does other mean here in terms of the variable name?

zaxtax (commented):
It would be nice if there was some intuition for why some programs can be marginalized and some cannot. The distinction is a bit and for some cases it feels like we should be able to.

ricardoV94:
> It would be nice if there was some intuition for why some programs can be marginalized and some cannot. The distinction is a bit and for some cases it feels like we should be able to.

You can marginalize models as long as they don't mix independent dimensions of the marginal RV (and as long as we can be sure of that)

pymc_experimental/model/marginal_model.py
ricardoV94: it's the set of model RVs that are direct ancestors to the dependent rvs, excluding the marginal rv

ricardoV94:
This PR now depends on https://github.com/pymc-devs/pymc/pull/7159

zaxtax:
Aside from being re-based, are there any blockers to merging this?

twiecki:
> Aside from being re-based, are there any blockers to merging this?

@ricardoV94 just had a baby so probably won't respond. Do you want to rebase and we can just merge for now?

ricardoV94:
Nah I can still type now and then. PR was snot yet done

zaxtax:
Congrats @ricardoV94 :partying_face:

ricardoV94:
Failing CI is due to weird stuff between PyMC and PyTensor:  

Hopefully it will be sorted out soon. Tests pass locally on my machine.

pymc_experimental/model/marginal/distributions.py
```diff
@@ -45,31 +71,77 @@ def get_domain_of_finite_discrete_rv(rv: TensorVariable) -> tuple[int, ...]:
...
+    return reduced_logp
+
+
+def align_logp_dims(dims: tuple[tuple[int, None]], logp: TensorVariable) -> TensorVariable:
+    """Align the logp with the order specified in dims."""
+    dims_alignment = [dim for dim in dims if dim is not None]
+    return logp.transpose(*dims_alignment)
+
+
+def recreate_ofg_outputs(op: OpFromGraph, inputs: Sequence[Variable]) -> tuple[Variable]:
```
zaxtax, line 129: This would benefit from a docstring or comment

pymc_experimental/model/marginal/marginal_model.py
```diff
@@ -552,61 +549,47 @@ def collect_shared_vars(outputs, blockers):
...
+    output_rvs = [rv_to_marginalize, *dependent_rvs]
 
-    outputs = rvs_to_marginalize
     # We are strict about shared variables in SymbolicRandomVariables
-    inputs = input_rvs + collect_shared_vars(rvs_to_marginalize, blockers=input_rvs)
+    inputs = input_rvs + collect_shared_vars(output_rvs, blockers=input_rvs)
 
     if isinstance(rv_to_marginalize.owner.op, DiscreteMarkovChain):
-        marginalize_constructor = DiscreteMarginalMarkovChainRV
+        marginalize_constructor = MarginalDiscreteMarkovChainRV
```
zaxtax, line 584: Good rename!

tests/model/test_marginal_model.py
```diff
@@ -373,7 +373,6 @@ def test_recover_batched_marginal():
     assert post.lp_idx.shape == (*post.idx.shape, 2)
 
 
-@pytest.mark.xfail(reason="Still need to investigate")
```
zaxtax, line 383: Does this test pass now?

zaxtax (commented):
This looks really good. It's a little annoying that the refactor is mixed in with expansion of support for models, but looks pretty good.

ricardoV94:
> It's a little annoying that the refactor is mixed in with expansion of support for models, but looks pretty good

It's in separate commits

tests/model/test_marginal_model.py
ricardoV94: Yup

zaxtax:
> > It's a little annoying that the refactor is mixed in with expansion of support for models, but looks pretty good
> 
> It's in separate commits

Yea, I reviewed it commit by commit. I saw the bulk of the work was in the first one.

zaxtax:
We should probably bump the minimum version required to 5.17

ricardoV94:
> We should probably bump the minimum version required to 5.17

Done

pymc_experimental/model/marginal/distributions.py
ricardoV94: Added a comment

zaxtax (approved):
LGTM!

PR #302, opened 2024-02-05
Statespace: Don't automatically save statespace matrices as `Deterministic` variables

jessegrabowski:
I had originally done this to facilitate out-of-sample sampling tasks, so I could do e.g. `pm.Flat('T', T)`. The result was that all matrices were saved to the `idata`, creating pretty horrible `pm.model_to_graphviz` outputs like this:

image

This also was extremely memory wasteful. Many of the matrices are not random at all, and they were being saved `(chain, draw)` times. 

After this refactor, the matrices are dynamically rebuilt as needed from the parameter samples. The new graphs look like this:

image

There are also some en-passant changes to how exogenous data are handled that breaks the Structural example notebook. I will open a new PR after this one to address that, because I think I finally have it set up to handle forecasting with exogenous data. Basically, I was previously treating exogenous data like a type of "parameter". It's been upgraded to a first-class object, and custom models subclassing `PyMCStateSpace` that use exogenous data will now need to implement `data_names` and `data_info` properties.

### DIFF
--- a/pymc_experimental/statespace/core/statespace.py
+++ b/pymc_experimental/statespace/core/statespace.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Optional, Sequence, Union
 
 import numpy as np
 import pandas as pd
@@ -30,16 +30,13 @@ from pymc_experimental.statespace.utils.constants import (
     ALL_STATE_AUX_DIM,
     ALL_STATE_DIM,
     FILTER_OUTPUT_DIMS,
-    FILTER_OUTPUT_NAMES,
     FILTER_OUTPUT_TYPES,
     JITTER_DEFAULT,
     MATRIX_DIMS,
     MATRIX_NAMES,
     OBS_STATE_DIM,
-    SHOCK_AUX_DIM,
     SHOCK_DIM,
     SHORT_NAME_TO_LONG,
-    SMOOTHER_OUTPUT_NAMES,
     TIME_DIM,
     VECTOR_VALUED,
 )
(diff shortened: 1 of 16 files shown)


pymc_experimental/statespace/core/statespace.py
```diff
@@ -69,6 +66,16 @@ def _verify_group(group):
         raise ValueError(f'Argument "group" must be one of "prior" or "posterior", found {group}')
 
 
+def _matrix_vector_dot(A: pt.TensorVariable, x: pt.TensorVariable) -> pt.TensorVariable:
```
jessegrabowski, line 69: Is there a way to do this in "pure pytensor", without writing a helper?

ricardoV94:
Matrices that don't change could be defined as constant or mutabledata

pymc_experimental/statespace/core/statespace.py
ricardoV94: `(A @ x[..., None]).squeeze(-1)`?

jessegrabowski:
They're stored as TensorVariables in the statespace representation. This new way of doing this will directly use those once instead of copying them over and over into the idata.

ricardoV94:
> They're stored as TensorVariables in the statespace representation. This new way of doing this will directly use those once instead of copying them over and over into the idata. 

ConstantData and MutableData are also stores once in the idata as opposed to Deterministics which are stored per draw. Seems like that's what you wanted?

jessegrabowski:
Something like that, but I don't want to have a bunch of logic to decide if a matrix is static or contains parameters. If the user really wants to inspect matrices, he can ask for them manually and save them however he wants. In general, I think the important outputs for most users is going to be the parameters and the states. The rest can be more hidden away.

pymc_experimental/statespace/core/statespace.py
```diff
@@ -291,7 +314,7 @@ def _unpack_statespace_with_placeholders(self) -> Tuple:
 
         return a0, P0, c, d, T, Z, R, H, Q
 
-    def unpack_statespace(self) -> Tuple:
+    def unpack_statespace(self) -> list[pt.TensorVariable, ...]:
```
ricardoV94, line 317: List doesn't need ellipsis 

```suggestion
    def unpack_statespace(self) -> list[pt.TensorVariable]:
```

ricardoV94 (approved):
Just a tiny type-hint thing

pymc_experimental/statespace/core/statespace.py
jessegrabowski: Pycharm actually complains when I make this change, it wants me to literally list out `tuple[
        TensorVariable, TensorVariable, TensorVariable, TensorVariable, TensorVariable, TensorVariable, TensorVariable, TensorVariable, TensorVariable]`, which seems silly. `list[TensorVariable, ...]` was my hack-around.

jessegrabowski:
What's up with these jax test failures on the Ubuntu CI?

pymc_experimental/statespace/core/statespace.py
ricardoV94: Ignore pycharm

ricardoV94:
> What's up with these jax test failures on the Ubuntu CI?

https://github.com/pymc-devs/pymc-experimental/issues/305 ?

Issue #305, opened 2024-02-09
`test_histogram_approximation` failing due to warning in newer JAX release


ricardoV94:
https://github.com/pymc-devs/pymc-experimental/actions/runs/7842985013/job/21402357791?pr=304#step:7:2388

jessegrabowski:
Appears to be due to https://github.com/google/jax/pull/18989

ricardoV94:
Is the histogram functionality using JAX @ferrine? We may need to redesign if that's the case because it can't be use with PyMC multiprocessing in that case

jessegrabowski:
I think it's related to `pm.sample`, not histogram. There are test in `test_linear_model` and `test_model_builder` that are also failing because they call `pm.sample`

ricardoV94:
So what if JAX is loaded it will issue a warning on `os.fork` even if it's not being used?

ricardoV94:
Then let's just ignore the warning, it's stupid

PR #306, opened 2024-02-11
Fix JAX warnings in tests

jessegrabowski:
JAX warning mentioned #305 seems to be caused by using multiple cores when sampling with `pm.sample` in tests. This PR sets `cores=1` in all the tests to avoid it.

I also fixed a future warning emitted by `arviz` related to the use of the `coords` property to get the size of dimension -- the `sizes` property should be used instead.

Finally (and perhaps controversially?), I took the opportunity to move the test folder up one level, out of the project module. This matches the directory structure of `pymc` and `pytensor`, and makes more sense to me.


jessegrabowski:
Sorry I posted cringe, fixing the git history on this branch.

PR #307, opened 2024-02-11
Fix JAX warnings in tests

### LINKED ISSUE #305: `test_histogram_approximation` failing due to warning in newer JAX release

jessegrabowski:
Closes #305 

JAX warning mentioned https://github.com/pymc-devs/pymc-experimental/issues/305 seems to be caused by using multiple cores when sampling with pm.sample in tests. The JAX test appears to be over-eager, because it is issued even when JAX is not being used, as long as JAX has been imported. 

This PR adds the warning to `filterwarnings` in the `pyproject.toml`, since it doesn't appear to be relevant to the tests its causing to fail.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,9 @@ filterwarnings =[
 
   # bool8, find_common_type, cumproduct, and product had deprecation warnings added in numpy 1.25
   'ignore:.*(\b(pkg_resources\.declare_namespace|np\.bool8|np\.find_common_type|cumproduct|product)\b).*:DeprecationWarning',
+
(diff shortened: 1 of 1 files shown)


jessegrabowski:
I wanted to filter by module (`jax._src.xla_bridge`), but it didn't work. This does, so I'm going with it.

By far my worst work, but it fixes it issue.

ricardoV94 (approved):

ricardoV94:
> I'd still like to move `tests` up one level, out of the project files.

Let's do that in a separate PR

Issue #317, opened 2024-03-25
new `pymc` version breaks import of `pymc_experimental`


RZachLamberty:
if you install `pymc==5.11` and `pymc-experimental==0.0.17`, changes to the pymc.distributions.distributrion` module make it impossible to import `pymc_experimental`:

```python
import pymc_experimental
```

will result in

```
ImportError: cannot import name '_moment' from 'pymc.distributions.distribution' (xxx/lib/python3.10/site-packages/pymc/distributions/distribution.py)

----> 1 import pymc_experimental
File xxx/lib/python3.10/site-packages/pymc_experimental/__init__.py:26
     23         handler = logging.StreamHandler()
     24         _log.addHandler(handler)
---> 26 from pymc_experimental import distributions, gp, statespace, utils
     27 from pymc_experimental.inference.fit import fit
     28 from pymc_experimental.model.marginal_model import MarginalModel
File xxx/lib/python3.10/site-packages/pymc_experimental/distributions/__init__.py:28
     26 from pymc_experimental.distributions.histogram_utils import histogram_approximation
     27 from pymc_experimental.distributions.multivariate import R2D2M2CP
---> 28 from pymc_experimental.distributions.timeseries import DiscreteMarkovChain
     30 __all__ = [
     31     "BetaNegativeBinomial",
     32     "DiscreteMarkovChain",
   (...)
     39     "Maxwell",
     40 ]
File xxx/lib/python3.10/site-packages/pymc_experimental/distributions/timeseries.py:9
      7 import pytensor.tensor as pt
      8 from pymc.distributions.dist_math import check_parameters
----> 9 from pymc.distributions.distribution import (
     10     Distribution,
     11     SymbolicRandomVariable,
     12     _moment,
     13     moment,
     14 )
     15 from pymc.distributions.shape_utils import (
     16     _change_dist_size,
     17     change_dist_size,
     18     get_support_shape_1d,
     19 )
     20 from pymc.logprob.abstract import _logprob
```

ricardoV94:
Thanks for flagging, we failed to deprecate softly in https://github.com/pymc-devs/pymc/pull/7166

@aerubanov do you happen to have the time to fix things on this end?

aerubanov:
@ricardoV94 yeah, will take a look today

ricardoV94:
It's still useful to update pymc-experimental not to depend on the deprecated methods, as the tests will start to fail anyway

rklees:
[...]

PR #320, opened 2024-03-28
Update for PyMC v5.11

zaxtax:
This should let `pymc-experimental` build with pymc v5.11.

This relates to #317

### DIFF
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -11,6 +11,6 @@ dependencies:
 - xhistogram
 - statsmodels
 - pip:
-  - pymc>=5.10.0  # CI was failing to resolve
+  - pymc>=5.11.0  # CI was failing to resolve
   - blackjax
   - scikit-learn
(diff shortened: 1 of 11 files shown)


ricardoV94:
@zaxtax you also need to update the versions here: https://github.com/pymc-devs/pymc-experimental/tree/main/conda-envs

ricardoV94:
Tests are failing because `moment` is still being used internally?

zaxtax:
It seems to fail locally for me as well.

ricardoV94:
> You need to replace these methods as well:
> 
> https://github.com/pymc-devs/pymc-experimental/blob/ee07caa259517c14697e285f5029c1cca076d1b9/pymc_experimental/distributions/continuous.py#L214

I just picked one, I think there are more

zaxtax:
Oh weird grep wasn't finding that for me for some reason

zaxtax:
The current errors since unrelated to moment but happy to fix them.

ricardoV94 (approved):

Issue #297, opened 2024-01-09
`statespace`: Leveraging RegressionComponent yields error


A108669:
Hello!
I have been experimenting with the Structural Time Series module; however, I have been running into trouble when I attempt to add an external regressor to a model.   Below is an example including dummy data as well as the full stack trace.  I have not been able to get any external regressors added to the models. 

```python

import pandas as pd
import numpy as np

# Generate dummy data with monthly seasonality and trend
n_samples = 1000
np.random.seed(100)

trend_data = np.arange(n_samples) * .1
regressor_data = np.random.normal(scale=2, size=n_samples)
y = trend_data + regressor_data + np.random.normal(scale=2, size=n_samples) + 10
df = pd.DataFrame(
    data={
        'time_index': pd.date_range("2001-01-01", freq="M", periods=len(y)),
        'x': regressor_data,
        'y': y,
    }
)

trend = st.LevelTrendComponent(name="linear_trend", order=2, innovations_order=0)
regressor = st.RegressionComponent(name="xreg", k_exog=1)
error = st.MeasurementError(name="error")

df = df.set_index("time_index")
df.index.freq = 'M'

mod = trend + error + regressor
ss_mod = mod.build(name="test")
sigma_trend_dims, sigma_obs_dims, regressor_dims, _, P0_dims = ss_mod.param_dims.values()
coords = ss_mod.coords

with pm.Model(coords=coords) as model_1:
    P0_diag = pm.Gamma("P0_diag", alpha=2, beta=5, dims=P0_dims[0])
    P0 = pm.Deterministic("P0", pt.diag(P0_diag), dims=P0_dims)

    initial_trend = pm.Normal("initial_trend", dims=initial_trend_dims)
    sigma_error = pm.Gamma("sigma_error", alpha=2, beta=5, dims=["observed_state"])

    beta_xreg = pm.Normal("beta_xreg", .2, 1)
    data_xreg = pm.MutableData("data_xreg", df[["x"]])

    ss_mod.build_statespace_graph(df[['y']], mode="JAX")
    idata = pm.sample(nuts_sampler="numpyro", target_accept=0.9)
```

```python
ValueError: Argument [[-0.85651 ... 33556989]] given to the scan node is not compatible with its corresponding loop function variable *4-<Matrix(float64, shape=(?, ?))>
```

Full Stack Trace:
```python
The following parameters should be assigned priors inside a PyMC model block: 
[...]

PR #326, opened 2024-04-07
Bug fixes for `statespace`

### LINKED ISSUE #318: New error message when building statespace_graph
I installed the latest version of pymc-experimental and get an error message pointing to 

ssm_model.build_statespace_graph(data['meas'], mode="JAX")

The error message is 

TypeError: Cannot convert Type Scalar(float64, shape=()) (of Variable sigma_SA_cycle) into Type Vector(float64, shape=(1,)). You can try to manually convert sigma_SA_cycle into a Vector(float64, shape=(1,)).

It refers to the pytensor variable 

sigma_SA_cycle = pm.Gamma("sigma_SA_cycle", alpha=2, beta=5)

which is defined in pm.model()

I have installed pytensor version 2.18.6.

Any idea of how to fix this?

jessegrabowski:
When I first wrote `StructuralComponent`, I had an idea that every model parameter should have dims associated with it. This had the consequence that even if a parameter should *always* be a scalar, such as standard deviations on shocks, it needed to be given `shape=(1,)` if `dims` were not used. This was a bit silly, so since #296, it has been my intention that scalar-only parameters should be scalars (and have no associated `dims`). #318 points out that the code is currently inconsistent -- some parameters are still having dims generated for them that don't need them, or vice-versa. 

This PR should hopefully clean all this up. Parameters that are always scalars will never have dims associated with them, do not expect a shape argument, and will show `shape=None` in the model info output when `structural_model.build(verbose=True)` is called.

In addition: 
- I re-run the structural notebook to work after recent changes.
- The name "time" can now be used as a dimension before calling `model.build_statespace_graph(observed_data)`. This is useful when setting exogenous data for an `ExogenousComponent`. In this case, a check is added that the index on the data passed to the `build_statespace_graph` is the same as that found on any existing `time` dim.
- `TimeSeasonality` previously always removed the first state because it is not identified. I added an option to keep it to I could test if `ZeroSumNormal` could be used to identify all the states.

Closes #318 , #297

### DIFF
--- a/pymc_experimental/statespace/core/statespace.py
+++ b/pymc_experimental/statespace/core/statespace.py
@@ -284,6 +284,8 @@ class PyMCStateSpace:
         Prints a short report to the terminal about the data needed for the model, including their names, shapes,
         and named dimensions.
         """
+        if not self.data_info:
+            return
 
         out = ""
         for data, info in self.data_info.items():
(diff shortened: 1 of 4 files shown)


ricardoV94 (approved):

jessegrabowski:
Can we force merge this, the failures are unrelated (we need to update tests after https://github.com/pymc-devs/pymc/pull/7211)

ricardoV94:
> Can we force merge this, the failures are unrelated (we need to update tests after https://github.com/pymc-devs/pymc/pull/7211). Also looks like the windows CI is totally borked by something in arviz.

sure

rklees:
I downloaded the software and run your Structural Timeseries Modelling notebook. Get an error message  in

ss_mod.build_statespace_graph(nile, mode="JAX")

saying

ValueError: All variables needed to compute inner-graph must be provided as inputs under strict=True. The inner-graph implicitly depends on the following shared variables [RandomGeneratorSharedVariable(<Generator(PCG64) at 0x165D86C00>)]

ricardoV94:
@rklees the code will only work with the latest version of PyMC after #329 

You can downgrade to PyMC 5.12.0 in the meantime

rklees:
I installed from https://github.com/pymc-devs/pymc-experimental/pull/326 and downgraded to pymc=5.12.0. After that, I could run the updated Structural Time Series modeling notebook and got the same results. However, running pymc-experimental on a synthetic datasets, which I am using already for months, provides results, which are completely different from those obtained with pymc-experimental version 0.0.16. Whereas version 0.0.16 provides very reasonable results, the 326 version provides unrealistic results. For instance, HDI intervals for the posterior predictive, are orders of magnitude too large.

jessegrabowski:
If you think there's a bug I'm happy to track it down, but this PR shouldn't have touched anything related to that. If something was messed up my guess is that it would have been introduced in #302. I'd need to see some code.

rklees:
I can send you two Jupyter notebooks, but dont know how to do that using github. If there is a simple way to do that, please let me know
Roland

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #337, opened 2024-04-20
Use vectorize in `finite_discrete_marginal_rv_logp`

ricardoV94:
This should be more performant and also faster to compile due to smaller computational graphs.

I've been waiting for the `vectorize` functionality to mature, and I think it's good now


zaxtax (approved):
LGTM

jessegrabowski (approved):

PR #345, opened 2024-06-01
Implement Laplace (quadratic) approximation

carsten-j:
This is an early version of q quadratic approximation implementation that I have developed while reading Statistical Rethinking by Richard McElreath.

There is a short discussion about this in the [issue](https://github.com/pymc-devs/pymc-experimental/issues/341) and maybe @theorashid can help with feedback of this draft PR. 

This work is partly based on the Python package [pymc3-quap](https://github.com/rasmusbergpalm/pymc3-quap) but pymc3-quap is based on PYMC3 and a lot happend bewteen version 3 and 5 of PYMC. Optimizers works better when provided with a good initial guess and hence a (optional) starting point has been added to function arguments. Please see [Github](https://github.com/pymc-devs/pymc/issues/5443#issuecomment-1030609090) for a discussion about the differences between PYMC version 3 and 5 for computing the Hessian.


carsten-j:
I am looking for the best way to return not just a posterior sample distribution but also the mean vector and covariance matrix of the Gaussian distribution. Any suggestion for this. So far my only idea is to add another section to the inferenceData returned containing this information. Thoughts on this?

zaxtax:
I'm not sure the InferenceData is the best place to put it. We should copy
whatever we do with Variational Inference

aloctavodia:
Historically, Inferencedata has been focused on mcmc. But we have discussed a few times extend it to better handle other inference methods, like SMC or variational methods. It just that there has not been enough momentum to agree and implement and schema that works for those methods.

carsten-j:
@zaxtax and @aloctavodia are you saying that I should not return inferencedata at all or just not return the gaussian mean and covariance in the inferencedata object? I am new to both PYMC and Bayesian statistics so I do not know the history of this package.
Best, Carsten

zaxtax:
Oh, it's more that we haven't decided how to handle this within the
library. Don't treat this as a blocker, though we should raise it for
discussion more broadly

twiecki:
CC @ferrine

ricardoV94:
Suggestion, have two groups in the returned inferencedata: 
1. A `fit` group that includes the mean and covariance of the laplace fit, and 
2. a `posterior` group that includes draws from this fit, and has all the bells and whistles like dimensions, deterministics, etc... Include the default extra groups like observed and constant data. This will look just like a fit from mcmc sampling. This can be disabled by the user by setting `draws = 0`

We could even try different fits from distinct initialization points (optionally) and save those as distinct "chains" in the `fit` and corresponding `posterior` groups. Although usually multiple initialization are used with the goal of finding the best fit, they could still be useful to detect multi-modality / pervasiveness of local optima.

pymc_experimental/inference/quadratic.py
```diff
@@ -0,0 +1,140 @@
...
+
+    map = pm.find_MAP(vars=vars)
+
+    m = pm.modelcontext(model)
+
+    for var in vars:
+        if m.rvs_to_transforms[var] is not None:
+            m.rvs_to_transforms[var] = None
+            var_value = m.rvs_to_values[var]
+            var_value.name = var.name
```
ricardoV94, line 87: Use the utility `remove_value_transforms`: https://www.pymc.io/projects/docs/en/stable/api/model/generated/pymc.model.transform.conditioning.remove_value_transforms.html

This must be done outside of a model context for now, we'll fix that soon. That means for testing purposes you must call `pm.fit(..., model=model)` outside of a model context

pymc_experimental/inference/quadratic.py
```diff
@@ -0,0 +1,140 @@
...
+    especially those with significant skewness or multimodality.
+
+    See Also
+    --------
+    fit : Calling the inference function 'fit' like pmx.fit(method="quadratic", vars=[mu, logsigma], model=m)
+          will forward the call to 'quadratic'.
+
+    """
+
+    random_seed = np.random.default_rng(seed=random_seed)
```
ricardoV94, line 77: Nitpick: More standard abbrevation, corresponds to a more precise name
```suggestion
    rng = np.random.default_rng(seed=random_seed)
```

pymc_experimental/inference/quadratic.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def quadratic(
+    vars: Sequence[Variable],
+    draws=1_000,
+    model=None,
+    random_seed: Optional[RandomSeed] = None,
+):
+    """
+    Create a quadratic approximation for a posterior distribution.
```
ricardoV94, line 25: Mention laplace somewhere early in the docstrings. Some folks have expressed a preference for calling it laplace actually (in which case quadratic should be mentioned early in the docstrings). I don't have a preference myself.

theorashid: I would prefer Laplace. numpyro also use Laplace https://docs.pyro.ai/en/1.4.0/infer.autoguide.html#autolaplaceapproximation

pymc_experimental/inference/quadratic.py
```diff
@@ -0,0 +1,140 @@
...
+    idata = az.convert_to_inference_data(ds)
+
+    idata = addDataToInferenceData(model, idata)
+
+    return idata
+
+
+def addDataToInferenceData(model, trace):
+    # Add deterministic variables to inference data
+    trace.posterior = pm.compute_deterministics(trace.posterior, model=model, merge_dataset=True)
```
ricardoV94, line 114:
```suggestion
    trace.posterior = pm.compute_deterministics(trace.posterior, model=model, merge_dataset=True, progressbar=False)
```

pymc_experimental/inference/quadratic.py
```diff
@@ -0,0 +1,140 @@
...
+    See Also
+    --------
+    fit : Calling the inference function 'fit' like pmx.fit(method="quadratic", vars=[mu, logsigma], model=m)
+          will forward the call to 'quadratic'.
+
+    """
+
+    random_seed = np.random.default_rng(seed=random_seed)
+
+    map = pm.find_MAP(vars=vars)
```
ricardoV94, line 79: Allow controlling progressbar
```suggestion
    map = pm.find_MAP(vars=vars, progressbar=progressbar)
```

ricardoV94:
@carsten-j PR looks great! I left some comment above

pymc_experimental/tests/test_quadratic.py
```diff
@@ -0,0 +1,64 @@
+#   Copyright 2022 The PyMC Developers
```
twiecki, line 1:
```suggestion
#   Copyright 2024 The PyMC Developers
```

carsten-j:
Thanks you @ricardoV94 and @twiecki for the review comments. I believe that all of them expect one has been fixed. I have not figured out how to use `remove_value_transforms`. I tried to browse through PYMC source code but that did not really help.

ricardoV94:
> Thanks you @ricardoV94 and @twiecki for the review comments. I believe that all of them expect one has been fixed. I have not figured out how to use `remove_value_transforms`. I tried to browse through PYMC source code but that did not really help.

The docs contains code example: https://www.pymc.io/projects/docs/en/stable/api/model/generated/pymc.model.transform.conditioning.remove_value_transforms.html

carsten-j:
I should have mentioned that I did read the doc and looked at the example. But I have not been able to figure out how to apply it to my case. I will try again ...

ricardoV94:
To be able to use it inside the model context, it will need this change to get merged first: https://github.com/pymc-devs/pymc/pull/7352

But you should be able to already test by doing the object way with `pm.fit(..., model=model)` outside of the model context

carsten-j:
@ricardoV94 I figured out how to replace the for loop with remove_value_transforms. Is the PR ready for merge or are there additional review comments?

pymc_experimental/inference/laplace.py
```diff
@@ -99,17 +100,14 @@ def laplace(
 
     rng = np.random.default_rng(seed=random_seed)
 
-    map = pm.find_MAP(vars=vars, progressbar=progressbar)
+    with pm.modelcontext(model) as transformed_m:
+        map = pm.find_MAP(vars=vars, progressbar=progressbar)
```
ricardoV94, line 104:
```suggestion
    transformed_m = pm.modelcontext(model)
    map = pm.find_MAP(vars=vars, progressbar=progressbar, model=transformed_m)
```

pymc_experimental/inference/laplace.py
```diff
@@ -99,17 +100,14 @@ def laplace(
...
 
-    m = pm.modelcontext(model)
+    # See https://www.pymc.io/projects/docs/en/stable/api/model/generated/pymc.model.transform.conditioning.remove_value_transforms.html
+    with remove_value_transforms(transformed_m) as untransformed_m:
+        untransformed_vars = [untransformed_m[v.name] for v in vars]
+        hessian = pm.find_hessian(point=map, vars=untransformed_vars)
```
ricardoV94, line 109:
```suggestion
    untransformed_m = remove_value_transforms(transformed_m):
    untransformed_vars = [untransformed_m[v.name] for v in vars]
    hessian = pm.find_hessian(point=map, vars=untransformed_vars, model=untransformed_m)
```

ricardoV94 (changes requested):
Does this work? If we already have the model, there's no need to enter the model context and force the inner functions to retrieve it again, we can just pass the model directly

carsten-j:
@ricardoV94 it does work but I have updated the code according to your comment. It makes sense and I am still learning about PYMC :-)

pymc_experimental/inference/laplace.py
```diff
@@ -0,0 +1,183 @@
...
+def laplace(
+    vars: Sequence[Variable],
+    draws=1_000,
+    chains=1,
+    model=None,
+    random_seed: Optional[RandomSeed] = None,
+    progressbar=True,
+):
+    """
+    Create a Laplace approximation for a posterior distribution.
```
ricardoV94, line 42:
```suggestion
    Create a Laplace (quadratic) approximation for a posterior distribution.
```

ricardoV94 (approved):
Looks pretty slick, left a minor docs suggestion

carsten-j:
@ricardoV94 tests are failing but I do not see any relationship between the failing tests and this new Laplace feature. It there a general problem with tests?

carsten-j:
@ricardoV94 tests are still failing. I am not sure what to do here as I do not see why these tests should fail due to my PR. Please advice.

ricardoV94:
@carsten-j I can confirm those failures are unrelated to your changes, and related to changes in PyMC or other dependencies. We'll have to fix them in a separate PR

carsten-j:
Thanks for clarifying @ricardoV94. What is next step from here. Do we need approval for additional maintainers or can the PR be merged now?

ricardoV94:
@carsten-j we would fix the tests elsewhere, then update this PR on top of the fixes, and if the tests are passing we can merge, unless any of the reviewers blocks it / requests more changes. You don't need to do anything :)

zaxtax:
Looks good. Once the tests pass, I think it's good to merge

pymc_experimental/inference/laplace.py
```diff
@@ -0,0 +1,183 @@
...
+    Parameters
+    ----------
+    vars : Sequence[Variable]
+        A sequence of variables for which the Laplace approximation of the posterior distribution
+        is to be created.
+    draws : int, optional, default=1_000
+        The number of draws to sample from the posterior distribution for creating the approximation.
+    chains : int, default=1
+        The number of chains to sample. For chains=0 only the fit of the Laplace approximation is
+        returned.s
```
theorashid, line 57: +returned.

pymc_experimental/tests/test_laplace.py
```diff
@@ -0,0 +1,65 @@
...
+    y = np.array([2642, 3503, 4358], dtype=np.float64)
+    n = y.size
+    draws = 100000
+    chains = 1
+
+    with pm.Model() as m:
+        logsigma = pm.Uniform("logsigma", 1, 100)
+        mu = pm.Uniform("mu", -10000, 10000)
+        yobs = pm.Normal("y", mu=mu, sigma=pm.math.exp(logsigma), observed=y)
+        vars = [mu, logsigma]
```
theorashid, line 44: Question: say you only did `vars=[mu]`, how would the variable `logsigma` be estimated?

ricardoV94: I think find_MAP in that case uses the initial_point for the excluded variable. I never found that behavior useful tbh

ricardoV94: Sounds like we may want to pass a similar `start` kwarg to laplace to set the value of variables that are not being optimized?

pymc_experimental/inference/laplace.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Parameters
+    ----------
+    vars : Sequence[Variable]
+        A sequence of variables for which the Laplace approximation of the posterior distribution
+        is to be created.
+    draws : int, optional, default=1_000
+        The number of draws to sample from the posterior distribution for creating the approximation.
+    chains : int, default=1
+        The number of chains to sample. For chains=0 only the fit of the Laplace approximation is
```
ricardoV94, line 56: Only returning the `fit` should be draws=0?

Re: chains, right now it doesn't make much sense because all chains will be the same. This changes if: 1) Users can pass different starts for different chains or 2) We add some random jitter across the initial point so that multiple chains could indeed arrive at different conclusions. I'm not sure this is useful, so maybe limiting to 1 chain / optimization (and removing the kwarg) may be fine?

ricardoV94: This was suggested by @theorashid

pymc_experimental/tests/test_laplace.py
theorashid: worth adding a test on this to confirm the behaviour

pymc_experimental/inference/laplace.py
theorashid: Just adding that it is pedagogically confusing to have multiple chains, because the user might try to incorrectly use MCMC diagnostics on the posterior to see if their "Laplace approximation" has converged. This is (probably) the same reason that the variation inference objects return idata with only one chain (the user again cannot specify multiple chains).

carsten-j: I think it makes sense to limit to 1 chain and not having chains as part of the method arguments.

carsten-j: The latest commit fixes this issue.

pymc_experimental/tests/test_laplace.py
carsten-j: I am not sure I fully understand this. I will give it a second go with the documentation for find_MAP.

pymc_experimental/inference/laplace.py
```diff
@@ -0,0 +1,182 @@
...
+    find_observations,
+)
+from pymc.model.transform.conditioning import remove_value_transforms
+from pymc.util import RandomSeed
+from pytensor import Variable
+
+
+def laplace(
+    vars: Sequence[Variable],
+    draws=1_000,
```
theorashid, line 35: nit: I think it would be `draws: Optional[int] = 1000` and then later down `if draws is not None:`

carsten-j: This is fixed with latest commit

pymc_experimental/inference/laplace.py
carsten-j: This has been fixed.

pymc_experimental/tests/test_laplace.py
theorashid: Yeah, just a little confirmation of what happens in the test case when you don't pass all the variables into `pm.fit` e.g. `vars = [logsigma]`. Doesn't need to be a full test if it doesn't break things, it can be a warning or a line in the docs at this point.

Rest of the PR looks great. Nice job

carsten-j: @theorashid it sound like a good idea with a warning in case one does not pass all variable into the fit method. I was going through the documentation and wondering if the PYMC model attribute 'value_vars' is the proper attribute to use to check if number of variables in 'vars' is the same as all the variables in the model.

carsten-j: Hi @theorashid. I am not sure how to handle if only a subset of the variables are passed in, e.g. vars=[mu] and log_sigma is left out. If this should raise a warning I need some way of figuring out the number of model parameters and compare that with the number of parameters in vars. I am not sure how to determine the number of model parameters

ricardoV94: `model.free_RVs`

carsten-j: @theorashid and @ricardoV94, I committed an update that will raise a warning in case number of variables in vars does not equal number of model variables.

ricardoV94:
@carsten-j tests are no longer failing in main. You can rebase/merge into your branch

pymc_experimental/tests/test_laplace.py
theorashid: LGTM

carsten-j:
@ricardoV94  I have rebased the laplace branch but it looks like someone needs to approve Github worksflows.

zaxtax:
Looks like there are still a few failing tests, but once those pass this is probably good to merge

carsten-j:
@zaxtax, all tests passed. Are you also able to merge the PR? Thanks.

twiecki:
Congrats @carsten-j, this is a big one!

carsten-j:
Thank you @twiecki. Really happy to contribute and thanks to all those that helped. After the summer I will try to work on documentation for building and running locally. I took me some time to figure out how this works!

zaxtax:
Congrats @carsten-j this is really neat!

theorashid:
Brilliant work @carsten-j . Hope to see you contribute to PyMC again!

PR #346, opened 2024-06-03
More bugfixes for statespace

jessegrabowski:
The statespace model is still broken, this PR is another round of bugfixes.

- [x] Update structural notebook to conform to API changes after #326 
- [x] Remove the numpy helper functions from the structural notebook, use pytensor only
- [x] Fix JAX-based forward sampling
- [x] Re-run all notebooks without errors
- [x] Get all current tests to pass
- [x] Add JAX tests

I need some help with fixing the JAX forward sampling. I'm doing something wrong, because even after freezing I have dynamic shape errors. This is the major blocker to considering the module "working" again.


jessegrabowski:
This is ready for review. I will need to make changes again after the next pymc/pytensor releases, but for now this all works -- tests pass, and all notebooks run.

For review I tried to organize the commits into chunks. It would be good to have eyeballs on the changes related to the distributions, since those were creating trouble for me in the first place.

jessegrabowski:
No idea what's going on with the CI, there seems to be a lot of broken stuff unrelated to this.

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -18,6 +22,64 @@
...
+    rv_op = MvNormalSVDRV()
+
+
+@jax_sample_fn.register(MvNormalSVDRV)
+def jax_sample_fn_mvnormal_svd(op, node):
+    def sample_fn(rng, size, dtype, *parameters):
+        rng_key = rng["jax_state"]
+        rng_key, sampling_key = jax.random.split(rng_key, 2)
+        sample = jax.random.multivariate_normal(
+            sampling_key, *parameters, shape=size, dtype=dtype, method="svd"
```
ricardoV94, line 75: Interesting.

More importantly, aren't you making JAX non-optional with the imports?

ricardoV94 (changes requested):

pymc_experimental/statespace/filters/distributions.py
jessegrabowski: Yes, good catch. I'll import inside the rewrite.

ricardoV94: I think importing the dispatch already forces jax to be importable. Usually I have had to have a try except import of jax

jessegrabowski: Oh at the top of the file you mean?

jessegrabowski: let me know if I did it right

ricardoV94: If you can import the package or the statespace submodule without jax istalled you did it right

pymc_experimental/statespace/filters/distributions.py
```diff
@@ -302,71 +364,67 @@ class SequenceMvNormal(Continuous):
...
+    def dist(cls, mus, covs, logp, **kwargs):
+        return super().dist([mus, covs, logp], **kwargs)
 
     @classmethod
-    def rv_op(cls, mus, covs, logp, steps, support_shape, size=None):
-        if size is not None:
-            batch_size = size
+    def rv_op(cls, mus, covs, logp, size=None):
+
+        # TODO: None of this does anything -- what am I doing wrong?
```
jessegrabowski, line 376: Need guidance here. I'm not sure how `support_shape` is meant to be used now that signatures are implemented. It seemed to do nothing so I removed it, but I saw that the timeseries distributions in PyMC core don't do this dimshuffling for scan

ricardoV94: Usually core shape gives us nsteps in the pymc timeseries but that's only needed because we don't have a parameter changing with time.

If you do here, like mu? you should have all the info you need without an explicit "core_shape" argument. 

Was this related to your question?

jessegrabowski: So I'm safe to just remove this in my case, since I get the time from the inputs and I can safely always assume batches will be added to the left?

ricardoV94: sounds right

ricardoV94 (approved):
Approving so you can merge, let me know if you want me to look at some specific changes carefully

PR #353, opened 2024-06-27
Marginalize continuous variable via QMC integration

ricardoV94:
CC @theorashid


pymc_experimental/model/marginal_model.py
```diff
@@ -870,3 +932,68 @@ def step_alpha(logp_emission, log_alpha, log_P):
...
+    rng = marginalized_rv_op.rng_param(marginalized_rv_node)
+    shape = constant_fold(tuple(marginalized_rv.shape))
+    size = np.prod(shape).astype(int)
+    n_draws = 2**op.qmc_order
+
+    # TODO: Wrap Sobol in an Op so we can control the RNG and change whenever
+    qmc_engine = scipy.stats.qmc.Sobol(d=size, seed=rng.get_value(borrow=False))
+    uniform_draws = qmc_engine.random(n_draws).reshape((n_draws, *shape))
+
+    if isinstance(marginalized_rv_op, MvNormal):
```
ferrine, line 966: how about using multipledispatch on this? Could be in a different PR. The rationale is that one can implement a new distribution without hacking pmx

ricardoV94: Most cases default to inverse icdf implementation, just mvnormal needs special case because it has no icdf and yet is probably pretty common

ricardoV94: I rather not make everything dispatch until it's shown to actually be necessary. Feels like premature modularity if you will

PR #357, opened 2024-07-05
Contribution page

michaelcao28:
This PR updates CONTRIBUTING.md to provide a basic workflow when working with both the pymc and pymc-experimental repos.
p.s. this is my first PR so please be kind :sweat_smile:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,45 @@
 # Contributing guide
 
+Thank you for your interest in contributing to PyMC and PyMC-experimental!
+
+This page outlines the steps to follow if you wish to contribute to the pymc-experimental repo and clone the repo locally.
+
+## Install locally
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,3 +1,113 @@
...
 
-Page in construction, for now go to https://github.com/pymc-devs/pymc-experimental#questions.
+Thank you for your interest in contributing to PyMC and PyMC-experimental!
+
+This page outlines the steps to follow if you wish to contribute to the pymc-experimental repo and clone the repo locally.
+
+## Install locally
+**1**: Create a folder `pymc-devs` in your local machine and follow the steps on [cloning PyMC locally](https://www.pymc.io/projects/docs/en/latest/contributing/pr_tutorial.html).
+
+Since PyMC-experimental should integrate with the latest version of PyMC, it is recommended that any development work on PyMC-experimental must also work with the latest version of PyMC.
```
twiecki, line 10: I don't think that's true, don't we pin a version?

ricardoV94: We don't but we test against the latest

PR #359, opened 2024-07-11
Specialized DiscreteMarkovChain step sampler

### LINKED ISSUE #324: Add specialized step sampler for DiscreteMarkovChain
A subclass of the Categorical sampler from PyMC should suffice

ricardoV94:
Closes #324 

It's just `CategoricalGibbsMetropolis` but accepts `DiscreteMarkovChain` and knows how to get the dimensionality of P.

Once we merge DMC into PyMC we can just extend the CategoricalGibbsMetropolis to accept it

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,6 @@
+ci:
+  autofix_prs: false
+
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.6.0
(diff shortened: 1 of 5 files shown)


ricardoV94:
Also I'm sure there must be more clever DMC samplers. This will update each step at a time, which I can imagine may be suboptimal

jessegrabowski (approved):
Seems like a fine stop-gap change

Maybe wait for the #361 to merge first then change where the test file is?

ricardoV94:
> Seems like a fine stop-gap change
> 
> Maybe wait for the #361 to merge first then change where the test file is?

Sure

PR #363, opened 2024-07-19
fix autoreparam because dims are no longer static

ferrine:


pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -174,15 +176,18 @@ def vip_reparam_node(
 ) -> Tuple[ModelDeterministic, ModelNamed]:
     if not isinstance(node.op, RandomVariable | SymbolicRandomVariable):
         raise TypeError("Op should be RandomVariable type")
-    rv = node.default_output()
-    try:
-        [rv_shape] = constant_fold([rv.shape])
-    except NotConstantValueError:
-        raise ValueError("Size should be static for autoreparametrization.")
+    _, size, *_ = node.inputs
+    eval_size = size.eval()
```
ricardoV94, line 180: How is this better? You can do `constant_fold(raise_if_not_constant=False)`

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -174,15 +176,18 @@ def vip_reparam_node(
...
+    if eval_size is not None:
+        rv_shape = tuple(eval_size)
+    else:
+        rv_shape = ()
+    lam_name = f"{name}::lam_logit__"
+    _log.debug(f"Creating {lam_name} with shape of {rv_shape}")
     logit_lam_ = pytensor.shared(
```
ricardoV94, line 187: Does this need to be a shared variable? How/who is controlling it?

pymc_experimental/model/transforms/autoreparam.py
ferrine: I remember hitting missing input errors, but I was doin plain eval

ricardoV94: Hmm the problem is you do need constant values with the shares variable

pymc_experimental/model/transforms/autoreparam.py
ferrine: it will appear named variable in the model, like pm.Data

ferrine: Is there a better way to create new pm.Data?

ricardoV94: Or you can have a shared variable but update the shape later? Does your algorithm now the shape of the rv value variable? It could set the this shared variable then

ricardoV94: Okay maybe this is fine

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -174,15 +176,18 @@ def vip_reparam_node(
...
-    except NotConstantValueError:
-        raise ValueError("Size should be static for autoreparametrization.")
+    _, size, *_ = node.inputs
+    eval_size = size.eval()
+    if eval_size is not None:
+        rv_shape = tuple(eval_size)
+    else:
+        rv_shape = ()
+    lam_name = f"{name}::lam_logit__"
+    _log.debug(f"Creating {lam_name} with shape of {rv_shape}")
```
ricardoV94, line 186: Do we need the log or is it just temporary for debugging?

ferrine: this log level is not printed by default, I decided to add that, since the shape can change some day, and this log message could be helpful.

ferrine: The debug logging is a very nice thing, you rarely need that, but when you do..., we have a very bad handling of this functionality so far, it can be improved one day

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -175,14 +177,13 @@ def vip_reparam_node(
     if not isinstance(node.op, RandomVariable | SymbolicRandomVariable):
         raise TypeError("Op should be RandomVariable type")
     rv = node.default_output()
-    try:
-        [rv_shape] = constant_fold([rv.shape])
-    except NotConstantValueError:
-        raise ValueError("Size should be static for autoreparametrization.")
+    [rv_shape] = constant_fold([rv.shape], raise_if_not_constant=False)
```
ricardoV94, line 180: Ignore my idea, this won't work for you. When it can't constant_fold, it returns the best graph it can but you still need constants for `np.zeros`

pymc_experimental/model/transforms/autoreparam.py
ferrine: ok, I reverted this due to this uncertainty. I only need this shape once, the computation of it can be addressed better. I saw in one place you did manual function compute with linker="py" and no optimizations, this would be suitable here as well but I do not want to duplicate code

ricardoV94: You can do `eval(mode="FAST_COMPILE")` if you want it to be fast. Still does some rewrites but not much and avoids C compilation

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -174,15 +176,18 @@ def vip_reparam_node(
...
     if not isinstance(node.op, RandomVariable | SymbolicRandomVariable):
         raise TypeError("Op should be RandomVariable type")
-    rv = node.default_output()
-    try:
-        [rv_shape] = constant_fold([rv.shape])
-    except NotConstantValueError:
-        raise ValueError("Size should be static for autoreparametrization.")
+    _, size, *_ = node.inputs
+    eval_size = size.eval(mode="FAST_COMPILE")
+    if eval_size is not None:
```
ricardoV94, line 181: eval_size will never be `None`

ricardoV94:
Looks good to me, just a useless `eval_size` check perhaps from when I told you to try the `constant_fold` function? Shouldn't be needed

pymc_experimental/model/transforms/autoreparam.py
ferrine: It can
image

ricardoV94: Wait why are you evaling size instead of rv.shape?

ferrine:
> Looks good to me, just a useless `eval_size` check perhaps from when I told you to try the `constant_fold` function? Shouldn't be needed

I'll have a branching condition to support both then

image

pymc_experimental/model/transforms/autoreparam.py
ferrine: I was afraid I get the dependence on the original RV

ricardoV94: If you use eval you can only get a numerical output

ricardoV94:
> > Looks good to me, just a useless `eval_size` check perhaps from when I told you to try the `constant_fold` function? Shouldn't be needed
> 
> I'll have a branching condition to support both then
> 
> image
> 

I think you cannot because you need to define the shared variable with a valid initial size?

pymc_experimental/model/transforms/autoreparam.py
ferrine: It does not seem to work, it complains on missing input error, probably the value var, I'm reverting back

ricardoV94: This is wrong max. If size is none the RV shape is implied by the parameters, it doesn't mean it is as scalar ()

ricardoV94: What missing input? There shouldn't be any inputs in the forward graph

ricardoV94: Value var should not ever show up in the forward graph

ricardoV94: Ah this is working on the fgraph...

ferrine: I have a dedicated method to update lambdas in the VIP class, but It is not handy after model transformations. It would be a good proposal to change that API later

ferrine: > So different question: can we create shared variable with data=np.zeros((1,)*rv.ndim) and only set correct shape later?
It should be possible, I assume that would mean I only know the shape after transformation is already done

ricardoV94: Do you need shape of transformed or original? If it is the transformed, you could request model.initial_point before starting the dispatch process?

Alternatively we should be able to use shape inference but may need to implement the `infer_shape` of the ModelVariable. The The helper `infer_static_shape` possibly followed by an eval should do the job and not complain about value variables because those shouldn't be in the graph after `infer_shape`

ricardoV94: Can you point me to which test failed when you tried `rv.shape.eval()`?

ferrine: this one: https://github.com/pymc-devs/pymc-experimental/actions/runs/10058239228/job/27800877978#step:5:1461

ferrine: > Do you need shape of transformed or original? If it is the transformed, you could request model.initial_point before starting the dispatch process?

This looks like an interesting solution

ferrine: > This looks like an interesting solution

```
model.initial_point()
```

Seems to give a point in the sampling space, I need the untransformed one

ricardoV94 (approved):
This was a tricky one!

ricardoV94:
Oh still failing :( ?

ferrine:
Yeah, I still get this input error
```
FAILED tests/model/transforms/test_autoreparam.py::test_set_truncate - pytensor.graph.utils.MissingInputError: Input 1 (s_log__) of the graph (indices start from 0), used to compute ModelFreeRV{transform=<pymc.logprob.transforms.LogTransform object at 0x7f4afe540550>}(s, s_log__), was not provided and not given a value. Use the PyTensor flag exception_verbosity='high', for more information on this error.
```

ferrine:
@ricardoV94 should I revert back to an initial solution? Without the fix, autoreparam can't be used on any model with dims(

ricardoV94:
Can you dprint the graph to see why the value variables are there?

ferrine:
I reverted to the solution which is working but not ideal. I'll open an issue to address the future-proof refactoring after we resolve this bug

ricardoV94:
Cab you add an xfail test with the car we know to be broken now?

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -174,15 +176,18 @@ def vip_reparam_node(
 ) -> Tuple[ModelDeterministic, ModelNamed]:
     if not isinstance(node.op, RandomVariable | SymbolicRandomVariable):
         raise TypeError("Op should be RandomVariable type")
```
ricardoV94, line 178: Can you also raise NotImplementedError for op.ndim_supp>0?

pymc_experimental/model/transforms/autoreparam.py
```diff
@@ -174,15 +176,18 @@ def vip_reparam_node(
...
-    rv = node.default_output()
-    try:
-        [rv_shape] = constant_fold([rv.shape])
-    except NotConstantValueError:
-        raise ValueError("Size should be static for autoreparametrization.")
+    _, size, *_ = node.inputs
```
ricardoV94, line 179:
```suggestion
    # FIXME: This is wrong when size is None
    _, size, *_ = node.inputs
```

pymc_experimental/model/transforms/autoreparam.py
ferrine: it will not pass the dispatch anyway, it is not needed

ferrine:
> Can you add an xfail test with the case we know to be broken now?

Can you please elaborate on this?

PS I wish github has threads

pymc_experimental/model/transforms/autoreparam.py
ferrine: > one lambda per shape or size

Per size for sure 100%

ricardoV94 (approved):

Issue #350, opened 2024-06-23
Update pre-commit hooks to use `ruff` and `ruff-format`


jessegrabowski:
The current one is still using black, which is slower. We should get with the times!

PR #366, opened 2024-07-23
update pre-commit-config replacing black and pylint with ruff

theorashid:
This is for #350, because I too wanted to replace black with ruff. There's a lot going on here but it's largely cosmetic. After changing the pre-commit file following the pytensor pre-commit config I:

- ran `pre-commit run --all-files`
- manually changed the remaining errors rather than using the `--unsafe-fixes` option. These were largely import errors (F401) or Local variable `mu` is assigned to but never used (F841, common when people write pymc models in tests).

The dubious ones that I changed which might be bugs were:
- `gp.latent_approx.py` L144 `mu = self.mean_func(X)` never used? Should it be?
- statespace e.g. `def rv_op( # noqa: F811`. I added the skip here but `rv_op` is specified above (e.g. `rv_op = LinearGaussianStateSpaceRV`) then overwritten in a class method. Just checking the desired behaviour

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,17 +20,12 @@ repos:
   hooks:
     - id: pyupgrade
       args: [--py37-plus]
-- repo: https://github.com/psf/black
-  rev: 24.1.1
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.5.4
   hooks:
-    - id: black
-    - id: black-jupyter
-- repo: https://github.com/PyCQA/pylint
-  rev: v3.0.3
-  hooks:
-    - id: pylint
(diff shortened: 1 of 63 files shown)


theorashid:
Couple of follow ups:
- why is `conftest.py` in the root. is that normal?
- can the setup.py/setupegg.py/pyproject.toml be consolidated into just `pyproject.toml` (happy to give this a go)
- can I add `.venv/` and `python-version` to the `.gitignore` for those of us who use venv/pyenv instead of conda

zaxtax:
Should this be combined with #368 ?

theorashid:
Working with @jessegrabowski to merge ours together. No need to review

PR #368, opened 2024-07-26
Update pre-commit

### LINKED ISSUE #350: Update pre-commit hooks to use `ruff` and `ruff-format`

jessegrabowski:
Closes #350 

Switches our pre-commit to use ruff. I also changed the pyupgrade to version 3.10, since that's the minimum we support these days.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.5.0
+  rev: v4.6.0
   hooks:
     -   id: check-merge-conflict
     -   id: check-toml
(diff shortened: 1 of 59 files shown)


maresb:
Tests are failing with `fixture 'rng' not found`

maresb:
This would be easier to review with code fixes in a separate commit.

Ruff works much better if rules have been explicitly defined in `pyproject.toml`.

Adding the rule `UP` makes `pyupgrade` mostly redundant, although last time I checked `pyupgrade` did a more thorough job.

maresb (approved):
Nice! This looks great now!

jessegrabowski:
@theorashid can you have a look at this and make sure I didn't miss anything you wanted to see in the configuration?

Regarding the points raised in #366 :

- I put `#noqa` on `gp.latent_approx.py L144 mu = self.mean_func(X)`. Not sure if it is supposed to be there, but I kept it (@bwengals )
- The `rv_op` thing in statespace didn't trigger, but I'll look into it.

For the root directory cleanup questions, you should open a separate issue (and I promise and I won't step on your toes again, sorry)

.pre-commit-config.yaml
```diff
@@ -10,37 +10,18 @@ repos:
     -   id: no-commit-to-branch
         args: [--branch, main]
     -   id: trailing-whitespace
-- repo: https://github.com/PyCQA/isort
-  rev: 5.13.2
+
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.4.8
```
maresb, line 15:
```suggestion
  rev: v0.5.5
```

.pre-commit-config.yaml
```diff
@@ -10,37 +10,18 @@ repos:
...
-    - id: black
-    - id: black-jupyter
-- repo: https://github.com/PyCQA/pylint
-  rev: v3.0.3
-  hooks:
-    - id: pylint
-      args: [--rcfile=.pylintrc]
-      files: ^pymc_experimental/
+    - id: ruff
+      args: [ --fix, --exit-non-zero-on-fix ]
```
maresb, line 18:
```suggestion
      args: [ --fix, --unsafe-fixes, --exit-non-zero-on-fix ]
```

maresb (commented):
> The difference is that it forces you to make the changes yourself, whereas pyupgrade does it automatically. I'm lazy so I like pyupgrade, but maybe one less dependency is worth it.

Sorry, GitHub glitched and I didn't see your comment when I previously approved. Give this a whirl and it should fix (almost) everything automatically now.

If you really truly love your reviewer then you can separate your commits even further via `git commit --no-verify` to commit the automatic fixes while leaving in the remaining errors so that by-hand fixes can be viewed separately.

jessegrabowski:
I do really love my reviewer, so I did as you asked.

maresb (approved):
Immaculate!

theorashid:
I'm happy with this. I'll close my PR and let's close the original issue too. Maybe I'll pick up more typos when we clean up the project root

PR #369, opened 2024-07-27
Add ETS model

jessegrabowski:
Adds `BayesianETS` as an available statespace model.

According to [this book](https://www.otexts.com/fpp3/ets.html), this is great baseline model for forecasting. I don't have a lot of experience with it, but statsmodels implements it, so I did too. 

Todo:

- [x] Write an example notebook
- [ ] Add a class method utility to decompose the time series into level, trend, and seasonal components
- [x] Test time series generation against statsmodels
- [x] Check whether the first state is correctly interpreted in the Kalman filter 

The last point is a but subtle. From the [statsmodel implementation](https://github.com/statsmodels/statsmodels/blob/d02d785b618afc3ce79859b9417ea7c0165d89ea/statsmodels/tsa/statespace/exponential_smoothing.py#L103C1-L107C78):

>    One consequence is that the "initial state" corresponds to the "filtered" state at time t=0, but this is different from the usual state space initialization used in Statsmodels, which initializes the model with the "predicted" state at time t=1. 

I have no idea whether this is important or not for the present implementation, I need to look at it more closely.

**Some comments:** 
In general I just follow the statsmodel implementation, and I'm testing that all the statespace matrices match those of statsmodels. The one difference is in the selection matrix. Looking at the book equations, for example:

$$
\begin{align}
y_t &= \ell_t + \epsilon_t \\
\ell_t &= \ell_{t-1} + \alpha \epsilon_t
\end{align}
$$

The general statespace equation is:

$$
\begin{align}
x_t &= T x_{t-1} + c + R \epsilon_t \\
y_t &= Z x_t + d + H
\end{align}
$$

This implies states: 

$$x_t = \begin{bmatrix} \epsilon_t \\ 
\ell_t \end{bmatrix}$$

So:

$$T = \begin{bmatrix} 0 & 0 \\
0 & 1
\end{bmatrix}, R = \begin{bmatrix} 1 \\ 
\alpha
\end{bmatrix}
$$

This all seems straightforward, but statsmodels sets the (0, 0) element of $R$ to be $1 - \alpha$. That really doesn't seem right, so I didn't do it.


jessegrabowski:
So statsmodels was right of course. My equations above were slightly wrong. We want to have:

$$
\begin{align}
y_t &= \ell_{t-1} + \epsilon_t \\
\ell_t &= \ell_{t-1} + \alpha \epsilon_t
\end{align}
$$

This could be done by making $y_t$ a hidden state, then:

$$T = \begin{bmatrix} 0 & 1 \\ 
0 & 1 \end{bmatrix},  Z_t = \begin{bmatrix} 1 & 0 \end{bmatrix}$$

But we can also get back to the timing that I had above, with $y_t = \ell_t + \epsilon_t$, if we do some algebra. Specifically, $\ell_{t-1} = \ell_t - \alpha \epsilon_t$, so $y_t = \ell_t + \epsilon_t - \alpha \epsilon_t = \ell_t + (1 - \alpha) \epsilon_t$

So that's where the $R$ adjustments in statsmodels were coming from. The same thing happens when you bring in a seasonal component; it becomes $y_t = \ell_t + s_{t-m) + (1 - \alpha - \gamma) \epsilon_t$

I also added an alternative parameterization that I thing might be nicer for priors. Under the $\alpha, \beta, \gamma$ parameterization, you need $\alpha \in (0, 1)$, $\beta \in (0, \alpha)$, and $\gamma \in (0, 1 - \alpha)$. Which is pretty hairy to set priors for. If instead you go with:

$$
\begin{align}
y_t &= \ell_{t-1} + s_{t-m-1} + \epsilon_t \\
\ell_t &= \ell_{t-1} + b_{t-1} + \alpha \epsilon_t \\
b_t &= b_{t-1} + \alpha \beta^\star \epsilon_t \\
s_{t-m} &= s_{t-m-1} + (1 - \alpha)\gamma^\star \epsilon_t
\end{align}
$$

Then all of $\alpha, \beta^\star, \gamma^\star$ are between 0 and 1, which is nice. So that's the default now, and you can ask for the other one with `use_transformed_parameterization=True`.

pyproject.toml
```diff
@@ -20,6 +20,9 @@ filterwarnings =[
 
   # Warning coming from blackjax
   'ignore:jax\.tree_map is deprecated:DeprecationWarning',
+
+  # shutil error
+  'ignore: Use shutil\.which instead of find_executable:DeprecationWarning',
```
ricardoV94, line 25: Not needed anymore, fixed in the last released PyTensor

ricardoV94:
Typo: 3 instances of "forcasting" (e missing) in the notebook

pymc_experimental/statespace/models/ETS.py
```diff
@@ -0,0 +1,678 @@
...
+
+        Parameters
+        ----------
+        order: tuple of string, Optional
+            The exponential smoothing "order". This is a tuple of three strings, each of which should be one of 'A', 'Ad',
+            or 'N'.
+
+                - The first element indicates the type of errors to use. Only 'A' is allowed.
+                - The second element indicates the type of trend to use. 'A', Ad' or 'N' are allowed.
+                - The third element indicates the type of seasonal component to use. 'A' or 'N' are allowed.
```
ricardoV94, line 158: Could be encoded in type hints?

pymc_experimental/statespace/models/ETS.py
```diff
@@ -0,0 +1,678 @@
...
+            seasonal = order[2] == "A"
+
+        self.trend = trend
+        self.damped_trend = damped_trend
+        self.seasonal = seasonal
+        self.seasonal_periods = seasonal_periods
+        self.use_transformed_parameterization = use_transformed_parameterization
+        self.stationary_initialization = stationary_initialization
+
+        if (initialization_dampening >= 1.0) or (initialization_dampening < 0.0):
```
ricardoV94, line 253: Does not matter:
```suggestion
        if not (0 < initialization_dampening < 1.0):
```

pymc_experimental/statespace/models/ETS.py
```diff
@@ -0,0 +1,678 @@
...
+
+        return coords
+
+    def _stationary_initialization(self, T_stationary, mode=None):
+        # Solve for matrix quadratic for P0
+        R = self.ssm["selection"]
+        Q = self.ssm["state_cov"]
+
+        # TODO: How to get mode information to here? It's not available when the model is created.
+        method = "direct" if ((self.k_states < 50) or (mode == "JAX")) else "bilinear"
```
ricardoV94, line 464: Make it a JAX rewrite (if you can never use bilinear in JAX)

ricardoV94 (approved):

pymc_experimental/statespace/models/ETS.py
jessegrabowski: I like this idea, but I want to handle it on the pytensor side rather than as a specialized rewrite in this library. I'll open an issue

ricardoV94: Actually should just be part of the lapyunov dispatch? No need for rewrite

jessegrabowski: I think it has to be a rewrite. The dispatcher is just an if/else switch. If `mode == "bilinear"` a specialized `Op` is introduced that has no JAX mode. Otherwise, we just do a computation using pytensor, no Op.

junpenglao:
Oh wow, I have seen folks express ETS as a states pace model, but never in a way that so clear and straightforward - very nice!

PR #372, opened 2024-08-11
Allow forecasting with exogenous variables

jessegrabowski:
Enable scenario-based forcasting in VARMAX, SARIMAX, and structural models with regressions.

Couple to-dos before its ready for review:

- [x] Carefully check the timing on the forecasting. Things seem to be shifted forward by 1 or 2 periods.
- [x] Recycle the exogenous forecasting code to allow in-sample counterfactual scenarios using `sample_conditional_posterior` or `sample_conditional_prior`. 
- [ ] Update relevant example notebooks to show how to use the scenario tools

Otherwise things seem to be working

### DIFF
--- a/pymc_experimental/statespace/core/statespace.py
+++ b/pymc_experimental/statespace/core/statespace.py
@@ -1514,7 +1514,7 @@ class PyMCStateSpace:
 
     @staticmethod
     def _validate_forecast_args(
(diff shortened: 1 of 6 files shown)


jessegrabowski:
I agree that this needs some notebooks to show the new usage, but I will do them in a follow up PR

PR #380, opened 2024-10-03
Make jax optional in testing suite

ricardoV94:


ricardoV94:
CI should be fixed after picking pytensor 2.25.6

PR #381, opened 2024-10-03
Update version (and pin failing micromamba)

ricardoV94:


maresb:
As a PR this confuses me because it seemingly does two seemingly unrelated things:
1. Fix the CI
2. Update the version number in preparation for a release

You might consider updating the PR title. And then as soon as it's merged, so that we don't forget, consider immediately creating a (currently-failing) PR to unpin micromamba.

ricardoV94:
The goal is to do a release, it happens that the CI is failing, so we pinned. They are separate commits and the PR itself won't show up in the release notes. I'll open the issue to unpin micromamba

maresb (approved):
Thanks for clarifying, sounds good!

Last week I had an incident where I cut a release, and the behavior changed in an unexpected way. I spent a long time scouring my PR titles for a potential explanation, and had to do a painstaking bisection. The culprit was a PR where I made an incidental fix that obviously resulted in the change, but I forgot to note this anywhere. So at the moment I'm a bit jumpy about this stuff. :joy:

ricardoV94:
> Thanks for clarifying, sounds good!
> 
> Last week I had an incident where I cut a release, and the behavior changed in an unexpected way. I spent a long time scouring my PR titles for a potential explanation, and had to do a painstaking bisection. The culprit was a PR where I made an incidental fix that obviously resulted in the change, but I forgot to note this anywhere. So at the moment I'm a bit jumpy about this stuff. 😂

My policy is unrelated changes can go in the same PR as long as they are in separate clean commits and we don't have to squash merge. Otherwise they must go on separate PRs. In this case it seemed silly to run the CI twice as well

maresb:
Makes sense, thanks!

PR #382, opened 2024-10-14
Summarize model as rich table

ricardoV94:
From https://github.com/pymc-devs/pymc/discussions/7210#discussioncomment-10921154

```python
import numpy as np
import pymc as pm

from pymc_experimental.printing import model_table

coords = {"subject": range(20), "param": ["a", "b"]}
with pm.Model(coords=coords) as m:
    x = pm.Data("x", np.random.normal(size=(20, 2)), dims=("subject", "param"))
    y = pm.Data("y", np.random.normal(size=(20,)), dims="subject")

    beta = pm.Normal("beta", mu=0, sigma=1, dims="param")
    mu = pm.Deterministic("mu", pm.math.dot(x, beta), dims="subject")
    sigma = pm.HalfNormal("sigma", sigma=1)

    y_obs = pm.Normal("y_obs", mu=mu, sigma=sigma, observed=y, dims="subject")

model_table(m)
```
```
 Variable  Expression         Dimensions             
─────────────────────────────────────────────────────
      x =  Data               subject[20] × param[2] 
      y =  Data               subject[20]            
                                                     
   beta ~  Normal(0, 1)       param[2]               
  sigma ~  HalfNormal(0, 1)                          
                              Parameter count = 3    
                                                     
     mu =  f(beta)            subject[20]            
                                                     
  y_obs ~  Normal(mu, sigma)  subject[20]     
```

### DIFF
--- a/pymc_experimental/printing.py
+++ b/pymc_experimental/printing.py
@@ -1,0 +1,182 @@
+import numpy as np
+
+from pymc import Model
+from pymc.printing import str_for_dist, str_for_potential_or_deterministic
+from pytensor import Mode
+from pytensor.compile.sharedvalue import SharedVariable
+from pytensor.graph.type import Constant, Variable
+from rich.box import SIMPLE_HEAD
+from rich.table import Table
+
+
+def variable_expression(
+    model: Model,
+    var: Variable,
+    truncate_deterministic: int | None,
+) -> str:
+    """Get the expression of a variable in a human-readable format."""
+    if var in model.data_vars:
+        var_expr = "Data"
+    elif var in model.deterministics:
+        str_repr = str_for_potential_or_deterministic(var, dist_name="")
+        _, var_expr = str_repr.split(" ~ ")
+        var_expr = var_expr[1:-1]  # Remove outer parentheses (f(...))
+        if truncate_deterministic is not None and len(var_expr) > truncate_deterministic:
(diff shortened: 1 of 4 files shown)


ricardoV94:
The output will be automatically improved once these issues are solved:
1. https://github.com/pymc-devs/pymc/issues/7536
1. https://github.com/pymc-devs/pymc/issues/7538

pymc_experimental/printing.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def _extract_value(var: SharedVariable | Constant) -> np.ndarray:
+    if isinstance(var, SharedVariable):
+        return var.get_value(borrow=True)
+    else:
+        return var.data
+
+
+def model_table(
```
jessegrabowski, line 20: I would suggest splitting out the functionality for actually creating the string representation from this function. That intermediate representation could return a list of (name, expression, dims) tuples, which could then be fed into different types of string representation (of which `model_table` is one)

ricardoV94: Yeah I was thinking the same

ricardoV94: I refactored the utilities that convert model info (variable expressions, dims) so they can be used elsewhere

jessegrabowski (approved):

Issue #340, opened 2024-05-15
Issue tracker for PyMC implementation of INLA


theorashid:
This is for https://github.com/pymc-devs/pymc/issues/3242 and https://github.com/pymc-devs/pymc/issues/6992.
cc: @ricardoV94 @zaxtax 

Approximate the marginal posterior distribution of some subset of the parameters, referred to as the marginal Laplace approximation. Then, integrate out the remaining parameters using another method.

This is great for [latent Gaussian models](https://theorashid.github.io/notes/INLA).

## 1. (Marginal) Laplace approximation

- Implement Laplace approximation
- Use Newton solver to find the mode of the Gaussian
- Use adjoint method to find the gradient of the Laplace approximation/mode

## 2. API

- Implement marginal Laplace approximation using step sampler interface

## 3. Sparse matrix operations

- Implement sparse matrix in pytensor with ability to:
  - add to the diagonal
  - multiply by a vector
  - solve a linear system
  - compute a log determinant
- Implement a MVN distribution with a sparse precision matrix

## 4. Interface

- Create an interface with bambi that can be accessed like R-INLA

## 5. Documentation and examples

- generalised linear mixed model example
- Spatial stats example
- pymc ICAR example but rewritten using INLA
- Time series example setting up AR model with ICAR (see [Dan footnote 41](https://dansblog.netlify.app/posts/2022-03-22-a-linear-mixed-effects-model/a-linear-mixed-effects-model#so-why-isnt-this-just-a-gaussian-process))

Note, I will update and link to the issues/PRs once they are made. If you want to tackle one of these issues, comment below and I will update the list with your name.

If you have any more things to add, please comment and I will add them to the list and create issues.

elizavetasemenova:
I'm interested in working on #341 and spatial stats

ColtAllen:
@theorashid @zaxtax do you feel the recent merge of [PR 513](https://github.com/pymc-devs/pymc-extras/pull/513) is enough to check off the `Newton Solver` box under _Section 1. Laplace Approximation_?

My thoughts are that the minimum requirements are met, but in the future we'll want to expand the API for better Hessian support and perhaps `pytensor.tensor.optimize.root`. The analytical expression for the "full" Laplace approximate requires a Hessian, but for now we're good to proceed with a simplified approximation (i.e., a third order Taylor series expansion).

ColtAllen:
[...]

PR #385, opened 2024-10-27
Add `find_MAP` with close JAX integration and fix bug with Laplace fit

### LINKED ISSUE #376: Laplace approximation not handling non-scalar parameters
I have been trying to utilize the new `laplace` function for estimating the posterior distribution using a quadratic approximation. It seems that there is an issue when trying to use non-scalar parameters. This is a slight modification of the example that is provided in the documentation for the function:

```
y = np.array([2642, 3503, 4358]*10)
cat = np.array([0]*15 + [1]*15)
with pm.Model() as m:
    logsigma = pm.Uniform("logsigma", 1, 100)
    mu = pm.Uniform("mu", -10000, 10000, shape=2)
    yobs = pm.Normal("y", mu=mu[cat], sigma=pm.math.exp(logsigma), observed=y)
    idata = laplace([mu, logsigma], model=m)
```

The following error is generated:

`TypeError: ('Wrong number of dimensions: expected 1, got 0 with shape ().', 'Container name "mu"')`

Interestingly, this error occurs even when the shape is not greater than 1 and no indexing is performed:

```
y = np.array([2642, 3503, 4358]*10)
with pm.Model() as m:
    logsigma = pm.Uniform("logsigma", 1, 100)
    mu = pm.Uniform("mu", -10000, 10000, shape=1)
    yobs = pm.Normal("y", mu=mu, sigma=pm.math.exp(logsigma), observed=y)
    idata = laplace([mu, logsigma], model=m)
```

This seems to indicate that there is an issue with setting the shape parameter to any value.

jessegrabowski:
Closes #376 

This PR adds code to run `find_MAP` using JAX. I'm using JAX for gradients, because I found the compile times were faster. Open to suggestions/rebuke.

It also adds a `fit_laplace` function, which is bad because we already have a `fit_laplace` function. This one has slightly different objective though -- it isn't meant to be used as a step sampler on a subset of model variables. Instead, it is meant to be used on the MAP result to give an approximation to the full posterior. My function also lets you do the Laplace approximation in the transformed space, then do sample-wise reverse transformation. I think this is legit, and lets you obtain approximate posteriors that respect the domain of the prior. Tagging @theorashid so we can resolve the differences.

Last point is that I added a dependency on `better_optimize`. This is a package I wrote that basically rips out the wrapper code used in PyMC `find_MAP` and applies it to arbitrary optimization problems. It is more feature complete than the PyMC wrapper -- it supports all optimizer modes for `scipy.optimize.minimize` *and* `scipy.optimize.root`, and also helps get keywords to the right place in those functions (who can ever remember if an argument goes in `method_kwargs` or in the funciton itself?). I plan to add support for `basinhopping` as well, which will be nice for really hairy minimizations.

I could see an objection to adding another dependency, but 1) it's a lightweight wrapper around functionality that doesn't *really* belong in PyMC anyway, and 2) it's a big value-add compared to working directly with the `scipy.optimize` functions, which have gnarly, inconsistent signatures.


theorashid:
Hey, nice one, yeah I agree, we should only have one `fit_laplace` function.

> it isn't meant to be used as a step sampler on a subset of model variables

The current `fit_laplace` isn't either. It isn't a step sampler. (The INLA stuff #340 still has a few blockers so that's separate and not yet in the library.) The implementation was made by a user following Statistical Rethinking, where McElreath fits some models using the Laplace approximation of all parameters.

Current behaviour of `fit_laplace` is:

- Take the MAP estimate of the parameters (https://github.com/pymc-devs/pymc-experimental/blob/main/pymc_experimental/inference/laplace.py#L109)
- Move to untransformed space (https://github.com/pymc-devs/pymc-experimental/blob/main/pymc_experimental/inference/laplace.py#L112)
- Get the Hessian (https://github.com/pymc-devs/pymc-experimental/blob/main/pymc_experimental/inference/laplace.py#L114)
- `cov = np.linalg.inv(hessian)` https://github.com/pymc-devs/pymc-experimental/blob/main/pymc_experimental/inference/laplace.py#L119
- simulate from this posterior approx `rng.multivariate_normal(mean, cov, size=(1, draws))` and turn into `idata`. (Looking at this now actually, we should probably transform back into the constrained space? Otherwise I can see e.g. standard deviation variables with samples below zero.)

The behaviour when you only pass a subset of variables isn't really desirable in my opinion (see https://github.com/pymc-devs/pymc-experimental/pull/345#discussion_r1642892896), so we put a warning. So as you say:

> Instead, it is meant to be used on the MAP result to give an approximation to the full posterior.

Agree, that's the best plan for `fit_laplace`.

Judging by your docs and a quick glance at your code, I think you're basically doing the same thing. The current implementation is few lines of code and a few docs, so I reckon

1) make sure you can pass the test case with your method, which is an example from BDA3 https://github.com/pymc-devs/pymc-experimental/blob/main/tests/test_laplace.py
2) throw any of the useful code and docs into your method

Then it should be safe to delete the existing code and we can go back to one `fit_laplace`.

> I could see an objection to adding another dependency

I would _love_ a generic optimiser in _p u r e_ pytensor, but I can see looking at your code that there a lot of fancy extras that would take a large effort to write in pytensor. Still, if we want to go back to one of our efforts with a fixed point operator (https://github.com/pymc-devs/pytensor/pull/978 and https://github.com/pymc-devs/pytensor/pull/944), we could probably write `find_MAP` with that in some form, with fewer bells and whistles though.

Happy to look at your code and review properly later in the week if you'd like me to. Let me know. Otherwise, I'll leave to the core devs.

pymc_experimental/inference/jax_find_map.py
```diff
@@ -0,0 +1,450 @@
...
+
+    for i, var in enumerate(variables):
+        raveled_shape = np.prod(optimized_point[var.name].shape).astype(int)
+        rv = model.values_to_rvs.get(var, var)
+
+        idx = slice(cursor, cursor + raveled_shape)
+        slices[rv] = idx
+        out_shapes[rv] = tuple(
+            [len(model.coords[dim]) for dim in model.named_vars_to_dims.get(rv.name, [])]
+        )
```
ricardoV94, line 74: What if an rv doesn't have dims specified?

pymc_experimental/inference/jax_find_map.py
```diff
@@ -0,0 +1,450 @@
...
+
+    Returns
+    -------
+    InferenceData
+        arviz.InferenceData object storing posterior, observed_data, and constant_data groups.
+
+    """
+    frozen_model = freeze_dims_and_data(model)
+    if not transform_samples:
+        untransformed_model = remove_value_transforms(frozen_model)
```
ricardoV94, line 153: Are you share you want to remove the value transforms? That will take the optimizer out of an unconstrained space

pymc_experimental/inference/jax_find_map.py
```diff
@@ -0,0 +1,450 @@
...
+        Whether to transform the samples back to the original parameter space. Default is True.
+    zero_tol: float
+        Value below which an element of the Hessian matrix is counted as 0.
+        This is used to stabilize the computation of the inverse Hessian matrix. Default is 1e-8.
+    diag_jitter: float | None
+        A small value added to the diagonal of the inverse Hessian matrix to ensure it is positive semi-definite.
+        If None, no jitter is added. Default is 1e-8.
+    progressbar : bool
+        Whether or not to display progress bar. Default is True.
+    mode : str
```
ricardoV94, line 142: No option is being given besides JAX

pymc_experimental/inference/jax_find_map.py
```diff
@@ -0,0 +1,450 @@
...
+    progressbar: bool = True,
+    mode: str = "JAX",
+) -> az.InferenceData:
+    """
+    Compute the Laplace approximation of the posterior distribution.
+
+    The posterior distribution will be approximated as a Gaussian
+    distribution centered at the posterior mode.
+    The covariance is the inverse of the negative Hessian matrix of
+    the log-posterior evaluated at the mode.
```
ricardoV94, line 115: ? Less ambiguous since there is a `mode` argument
```suggestion
    the log-posterior evaluated at the optimized_point.
```

ricardoV94:
> Happy to look at your code and review properly later in the week if you'd like me to. Let me know. Otherwise, I'll leave to the core devs.

That would be appreciated

ricardoV94:
Agree with what @theorashid said. This `fit_laplace` is going for the same goal as the previous one. Happy to replace it, if it's not married to JAX backend. Still fine to allow using JAX for the autodiff. What you're offering is very similar to nutpie `gradient_backend` kwarg, so we could use the same terminology

ricardoV94:
No objections about your custom library wrapper

pymc_experimental/inference/jax_find_map.py
jessegrabowski: Obviously because it's all you need : )

pymc_experimental/inference/jax_find_map.py
jessegrabowski: is there already a helper in pymc to figure out shapes of an rv?

ricardoV94: If you froze the dims you should have everything in `rv.type.shape`. Otherwise we usually use `model.initial_point`

jessegrabowski:
tagging @theorashid -- I couldn't pick you as a reviewer?

I did a major refactor of this. I broke the marriage to jax and generalized the find_MAP function. Files have been renamed to reflect this. I also merged the two laplace approaches. The biggest change is that I removed the ability to choose `vars`. I think the idea here was to be able to partially marginalize some variables in a model? But I think this would require a somewhat different approach.

theorashid:
yea sorry I'm just a normal, but I'll give it a review. Will do it at some point in the next 2 weeks.

pymc_experimental/inference/find_map.py
```diff
@@ -0,0 +1,415 @@
...
+        untransformed_X = f.backward(X[..., idx]) if f is not None else X[..., idx]
+
+        if rv in out_shapes:
+            new_shape = (chains, draws) + out_shapes[rv]
+            untransformed_X = untransformed_X.reshape(new_shape)
+
+        out.append(untransformed_X)
+
+    f_untransform = pytensor.function([X], out, mode="JAX")
+    return f_untransform(posterior_draws)
```
ricardoV94, line 89: Do we want to compile a function for a single eval? When I checked in JAX I found it was not worth it to e.g., JIT it. I would check if `mode=Mode(linker="py", optimizer=None)` would be worth it here. Also use `pytensor.In(borrow=True)` and `pytensor.Out(borrow=True)` to avoid deepcopy of variables that are not transformed.

jessegrabowski: yeah that's reasonable. I think the mode='JAX' is just a left-over from an intermediate stage of refactor when I was still hung up on JAX

pymc_experimental/inference/find_map.py
```diff
@@ -0,0 +1,415 @@
...
+        hess=f_hess,
+        hessp=f_hessp,
+        progressbar=progressbar,
+        method=method,
+        **optimizer_kwargs,
+    )
+
+    raveled_optimized = RaveledVars(optimizer_result.x, initial_params.point_map_info)
+    unobserved_vars = get_default_varnames(model.unobserved_value_vars, include_transformed)
+    unobserved_vars_values = model.compile_fn(unobserved_vars)(
```
ricardoV94, line 404:
```suggestion
    unobserved_vars_values = model.compile_fn(unobserved_vars, mode="FAST_COMPILE")(
```

pymc_experimental/inference/laplace.py
```diff
@@ -181,10 +221,334 @@ def addDataToInferenceData(model, trace, progressbar):
...
+        if on_bad_cov == "warn":
+            _log.warning(
+                "Inverse Hessian is not positive semi-definite at the provided point, using the closest PSD "
+                "matrix in L1-norm instead"
+            )
+
+    return mu, H_inv
+
+
+def laplace(
```
ricardoV94, line 326: usually better not to have a function with the same name as the module file. If you can think of a different name

ricardoV94 (approved):
Minor suggestions, PR looks amazing!

ricardoV94:
@jessegrabowski can we close https://github.com/pymc-devs/pymc-experimental/issues/376

Do you have a test that covers something like it?

jessegrabowski:
Yes, I think [this test](https://github.com/jessegrabowski/pymc-experimental/blob/f2504e98addc8c082038866dac3ead34925eecbc/tests/test_laplace.py#L115) and [this test](https://github.com/jessegrabowski/pymc-experimental/blob/f2504e98addc8c082038866dac3ead34925eecbc/tests/test_laplace.py#L166) should cover that issue

pymc_experimental/inference/find_map.py
```diff
@@ -230,8 +238,8 @@ def scipy_optimize_funcs_from_loss(
         Whether to compile a function that computes the Hessian of the loss function.
     use_hessp: bool
         Whether to compile a function that computes the Hessian-vector product of the loss function.
-    use_jax_gradients: bool
-        If True, use JAX to compute gradients. This is only possible when ``compile_kwargs["mode"]`` is set to "JAX".
+    gradient_backend: str, one of "jax" or "pytensor"
```
ricardoV94, line 241: should mention default

theorashid:
sweet, all done?

jessegrabowski:
For now, though I'd still appreciate it if you could have a look and open issues on any bugs/shortcomings you find

pymc_experimental/inference/find_map.py
```diff
@@ -0,0 +1,431 @@
...
+
+        def f_hess_jax(x):
+            return jax.numpy.stack(_f_hess_jax(x))
+
+        f_hess = jax.jit(f_hess_jax)
+
+    return f_loss_and_grad, f_hess, f_hessp
+
+
+def _compile_functions(
```
theorashid, line 155: NIT: Maybe `_compile_functions` and `_compile_jax_gradients` are _slightly_ too generic function names. I found it a little tricky to remember exactly what they were doing when reading through the code

pymc_experimental/inference/find_map.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+def set_optimizer_function_defaults(method, use_grad, use_hess, use_hessp):
+    method_info = MINIMIZE_MODE_KWARGS[method].copy()
+
+    use_grad = use_grad if use_grad is not None else method_info["uses_grad"]
+    use_hess = use_hess if use_hess is not None else method_info["uses_hess"]
+    use_hessp = use_hessp if use_hessp is not None else method_info["uses_hessp"]
+
+    if use_hess and use_hessp:
```
theorashid, line 37: I was going through all the methods thinking when you would need `hess` and `hessp` and then came back to this. I would probably warn the user / not let them pass _both_ `use_hess` and `use_hessp`

pymc_experimental/inference/laplace.py
```diff
@@ -181,10 +222,349 @@ def addDataToInferenceData(model, trace, progressbar):
...
         {"observed_data": observed_data, "constant_data": constant_data},
         coords=coords,
         dims=dims,
     )
 
-    return trace
+    return idata
+
+
+def fit_mvn_to_MAP(
```
theorashid, line 234: `fit_mvn_at_MAP`? I mean technically this function just fits a MVN at a `point`, the user doesn't necessarily have to pass the MAP

pymc_experimental/inference/laplace.py
```diff
@@ -181,10 +222,349 @@ def addDataToInferenceData(model, trace, progressbar):
...
+        np.linalg.cholesky(H_inv)
+    except np.linalg.LinAlgError:
+        if on_bad_cov == "error":
+            raise np.linalg.LinAlgError(
+                "Inverse Hessian not positive-semi definite at the provided point"
+            )
+        H_inv = get_nearest_psd(H_inv)
+        if on_bad_cov == "warn":
+            _log.warning(
+                "Inverse Hessian is not positive semi-definite at the provided point, using the closest PSD "
```
theorashid, line 324: For my understanding, what sort of scenarios/models would get a not PSD hessian. And is using closest PSD a good ideas?

pymc_experimental/inference/laplace.py
```diff
@@ -181,10 +222,349 @@ def addDataToInferenceData(model, trace, progressbar):
...
+    progressbar: bool = True,
+    random_seed: int | np.random.Generator | None = None,
+    compile_kwargs: dict | None = None,
+) -> az.InferenceData:
+    """
+    Generate samples from a multivariate normal distribution with mean `mu` and inverse covariance matrix `H_inv`.
+
+    Parameters
+    ----------
+    mu
```
theorashid, line 347: add docs here

pymc_experimental/inference/laplace.py
```diff
@@ -181,10 +222,349 @@ def addDataToInferenceData(model, trace, progressbar):
...
+        Whether to display a progress bar during optimization. Defaults to True.
+    fit_in_unconstrained_space: bool, default False
+        Whether to fit the Laplace approximation in the unconstrained parameter space. If True, samples will be drawn
+        from a mean and covariance matrix computed at a point in the **unconstrained** parameter space. Samples will
+        then be transformed back to the original parameter space. This will guarantee that the samples will respect
+        the domain of prior distributions (for exmaple, samples from a Beta distribution will be strictly between 0
+        and 1).
+
+        .. warning::
+            This argumnet should be considered highly experimental. It has not been verified if this method produces
```
theorashid, line 475: *argument

pymc_experimental/inference/laplace.py
```diff
@@ -181,10 +222,349 @@ def addDataToInferenceData(model, trace, progressbar):
...
+            valid draws from the posterior. **Use at your own risk**.
+
+    gradient_backend: str, default "pytensor"
+        The backend to use for gradient computations. Must be one of "pytensor" or "jax".
+    chains: int, default: 2
+        The number of sampling chains running in parallel.
```
theorashid, line 481: I'd probably add something here reiterating that this isn't a sampling inference method. This is just sampling from the approximated posterior. There was already people in the forum asking about the differences in these methods

tests/test_find_map.py
```diff
@@ -0,0 +1,98 @@
...
+
+    hess_val = np.array(f_hess(x_val))
+    np.testing.assert_allclose(hess_val.squeeze(), np.array([[2, 1], [1, 0]]))
+
+    hessp_val = np.array(f_hessp(x_val, np.array([1.0, 0.0])))
+    np.testing.assert_allclose(hessp_val.squeeze(), np.array([2, 1]))
+
+
+@pytest.mark.parametrize(
+    "method, use_grad, use_hess",
```
theorashid, line 57: any `use_hessp` tests? or are we just testing if scipy works here

theorashid (commented):
I managed to follow the code through and it looks good to me. Happy you got rid of the option to fit on a subset of variables, which didn't make sense to me anyway. If it passes the original test then it should be good. You can do something about the other comments if you want, but maybe not because we are e x p e r i m e n t a l

PR #393, opened 2024-11-22
Remove dead code from `kalman_filter.py`

jessegrabowski:
My original vision for `kalman_filter.py` was quite bad. Following `statsmodels`, I made a bunch of classes that instantiate various types of kalman filters, with minor differences between them. The idea was to let the user pick what would work best in his given situation.

Looking at it now, this should all be accomplished via rewrites. Some of it already is. For example, the only difference between the `StandardFilter` and the `SingleTimeSeriesFilter` was that I replaced a matrix inverse with a division. This is done automatically in rewrites, so there's no need for it.

I removed the `SteadyStateFilter` because that never made sense. Instead, all of the filters should be checking for convergence to the steady state -- that will be an upcoming PR.

`CholeskyFilter` has been renamed to `SquareRootFilter`, and it actually *is* a square root filter. I let the user pick it for now, but my plan is to rewrite to this when we see that the user has provided a lower triangular `P0`. I haven't benchmarked it, but it should be faster in that case (since solves are replaced with solve_triangular everywhere)

### DIFF
--- a/pymc_experimental/statespace/core/statespace.py
+++ b/pymc_experimental/statespace/core/statespace.py
@@ -19,7 +19,6 @@ from pytensor.compile import get_mode
 from pymc_experimental.statespace.core.representation import PytensorRepresentation
 from pymc_experimental.statespace.filters import (
     KalmanSmoother,
-    SingleTimeseriesFilter,
     SquareRootFilter,
     StandardFilter,
     UnivariateFilter,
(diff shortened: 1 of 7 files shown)


AlexAndorra (approved):
Much simpler indeed, thanks @jessegrabowski !
I'm guessing we need to update some examples after this is merged? Shall I merge BTW?

jessegrabowski:
Yeah some examples mention these other filter options, but none actually use them. It will be good to go through and delete all of that.

PR #396, opened 2024-12-05
Add `opt_sample` that performs optimizations prior to calling `pymc.sample`

ricardoV94:


pymc_experimental/sampling/optimizations/conjugacy.py
```diff
@@ -0,0 +1,153 @@
...
+    value = rv.type(name=name)
+    fgraph.add_input(value)
+    free_rv = model_free_rv(rv, value, transform, *dims)
+    free_rv.name = name
+    return free_rv
+
+
+@node_rewriter(tracks=[ModelFreeRV])
+def beta_binomial_conjugacy(fgraph: FunctionGraph, node):
+    [beta_free_rv] = node.outputs
```
ricardoV94, line 110: Do not apply rewrite if inputs depend on other model variables, as that can lead to less efficient sampling due to forcing a Gibbs scheme

wd60622: Possible to make this optional condition? Do we expect this will always happen (different distributions)? And expect for different conjugate models?

ricardoV94: From a discussion with @aseyboldt it will happen as long as there are correlations in the posteriors. But we can make optional by having a parametrized rewrite and tags: "conjugacy" vs "force_conjugacy" or something like that

wd60622: Makes sense. Does less efficient sampling mean taking more samples would be need to get the same results? If sampling conjugate step is faster then the trade-off of sampling more might be worth it. Or does it not work like that?

wd60622:
Need to try out. This example in description is different than the docstrings. Both work?

> ```python
> import pymc as pm
> import pymc_experimental as pmx
> 
> with pm.Model() as m:
>     p = pm.Beta("p", 1, 1, shape=(1000,))
>     y = pm.Binomial("y", n=100, p=p, observed=[1, 50, 99, 50]*250)
> 
>     idata = pm.sample(verbose=True) 
> 
> # Applied optimization: beta_binomial_conjugacy 1x
> # ConjugateRVSampler: [p]
> ```

ricardoV94:
> Need to try out. This example in description is different than the docstrings. Both work?
> 
> > ```python
> > import pymc as pm
> > import pymc_experimental as pmx
> > 
> > with pm.Model() as m:
> >     p = pm.Beta("p", 1, 1, shape=(1000,))
> >     y = pm.Binomial("y", n=100, p=p, observed=[1, 50, 99, 50]*250)
> > 
> >     idata = pm.sample(verbose=True) 
> > 
> > # Applied optimization: beta_binomial_conjugacy 1x
> > # ConjugateRVSampler: [p]
> > ```

Yes, I wanted an example that is faster than NUTS, but for testing it would be overkill

pymc_experimental/sampling/optimizations/conjugacy.py
```diff
@@ -0,0 +1,153 @@
...
+    beta_rv, beta_value, *beta_dims = node.inputs
```
wd60622, line 111: Is the `beta_value` used anywhere?

pymc_experimental/sampling/optimizations/conjugacy.py
```diff
@@ -0,0 +1,153 @@
...
+    # We want to replace free_rv by ConjugateRV()->(free_rv, conjugate_posterior_rv)
+    a, b = get_dist_params(beta_rv)
+    n, _ = get_dist_params(binomial_rv)
+
+    # Use value of y in new graph to avoid circularity
+    y = binomial_model_var.owner.inputs[1]
+
+    conjugate_a = a + y
+    conjugate_b = b + (n - y)
+    extra_dims = range(binomial_rv.type.ndim - beta_rv.type.ndim)
```
wd60622, line 138: Does this work in all cases?

pymc_experimental/sampling/optimizations/conjugacy.py
```diff
@@ -0,0 +1,153 @@
...
+    if extra_dims:
+        conjugate_a = conjugate_a.sum(extra_dims)
+        conjugate_b = conjugate_b.sum(extra_dims)
+    conjugate_beta_rv = Beta.dist(conjugate_a, conjugate_b)
```
wd60622, line 142: Would summing through data to make summary stats used be more general? 

For instance, y.sum(...) and (n - y).sum(...)

pymc_experimental/sampling/optimizations/conjugacy.py
wd60622: I guess the cases where there is some type of ExpandDims on one of the parameters

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: Added the option and an example of how to trigger it in the docstrings

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: Now it is actually replaced by a new value variable. I can underscore it.

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: Yes, I'm being lazy and only allowing broadcasting on the left. I wanted to keep things simple for the expression of the conjugate beta alpha and beta params. We can make it more flexible

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: That would mix independent dimensions of the beta. The model would no longer be equivalent to the original one

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: To answer your first question, it does not. If alpha already had one dimension, I should sum contributions from y/n-y and keep that dimension. Basically I have to unbroadcast uses of the sampe alpha/beta parameters

ricardoV94: Btw this sort of things are what's usually missing from the literature on conjugacy and bayesian model rewrites. Things are much more tricky once you start working with arrays of parameters

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: Specifically a could itself not be a scalar

wd60622: I am saying that the parameter, p, and its shape defines the model and to keep that. Make n and y broadcastable with p only as that will make broadcastable with a and b. For instance,

a: ()
b: ()
p: ()
n: (100, )
y: (100, )

Then to sum n and y to match dim of p, (). n.sum(axis=0), y.sum(axis=0)

Another example :

a: (3, )
b: ()
p: (4, 3) # Defined from shape 
n: ()
y: (100, 4, 1)

n summed to something broadcastable with (4, 3)
n, y.sum(axis=0)

wd60622:
```python
# Determined shape from p
y_stat = ...
n_stat = ...

# Cookie-cutter conjugate logic
a_post = a + y_stat
b_post = b + (n_stat - y_stat)
# Translate the conjugate parameters to PyMC distribution
conjugate_beta_rv = Beta.dist(alpha=a_post, beta=b_post) # Same shape as p before. Maybe shape needs to be passed as well

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: I fixed the second example. The first one is explicitly excluded right now.

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: Yup, that was the goal originally but as you found out it was not implemented correctly. I was only handing new dims. The last push has the `sum_bcast_dims` that reduces a_post/b_post to the shape of the original beta RV

ricardoV94: > a: (3, )
> b: ()
> p: (4, 3) # Defined from shape
> n: ()
> y: (100, 4, 1)

This example is not possible/not supported. the (100, 4, 1) cannot map to the (4, 3) of the p, unless you slice or something, which we don't allow

wd60622: Sure, but in the operation to get to posterior parameters, they are fine. 

Also, why not pass shape to new Beta.dist?

ricardoV94: > Sure, but in the operation to get to posterior parameters, they are fine and broadcast to the desired shape.

It matters how that shape went from 3 to 1. Where they summed/sliced/multiplied. Was there another operation in between like exponentiation/ reciprocal?

By only allowing direct use or left expand dims I rule out all these cases that I wouldn't know how to conjugate.

> Also, why not pass shape to new Beta.dist?

Yup, it's doing that now.

ricardoV94:
@wd60622 I had forgot to push the updates. I wasn't trying to gaslight you :D

jessegrabowski:
Any objection to cleaning this up and merging it? I think it's nice to have around, and I might have found an actual use case in DPs, see [this discussion](https://discourse.pymc.io/t/how-to-improve-ess-and-r-hat-for-dpmm/17165/4)

ricardoV94:
No objections

pymc_experimental/sampling/optimizations/conjugacy.py
ricardoV94: No, but we can easily convert a model to a version without dims and then apply the optimizations

williambdean:
I've resolved the conflicts

PR #417, opened 2025-01-24
Fix minor typo in Making a Custom Statespace Model.ipynb.

### LINKED ISSUE #416: Timing in State Space models
I noticed a timing inconsistency in the Making a Custom Statespace Model.ipynb notebook. 

Image

To be consistent with the general form in (1)-(2), the ar(3) model should be:

$$x(t) = \rho_1 x_{t-1} + \rho_2 x_{t-2}  + \rho_3 x_{t-3} + \varepsilon_{t-1}$$

instead of

$$x(t) = \rho_1 x_{t-1} + \rho_2 x_{t-2}  + \rho_3 x_{t-3} + \varepsilon_{t}$$

Alternatively, in (1) instead of $\varepsilon_{t}$ it should be $\varepsilon_{t+1}$. I realize that this doesn't matter in many cases, but in some models there is a big difference between the new state being driven by the same period innovation or the one from the last period. I apologise if this is explained elsewhere in the documentation, I have only looked at this notebook.

niskrev:
Change $x_{t+1}$ to $x_t$ on the left-hand side and $x_t$ to $x_{t-1}$ on the right-hand side of eq (1). 
Fixed #416

### DIFF
--- a/notebooks/Making a Custom Statespace Model.ipynb
+++ b/notebooks/Making a Custom Statespace Model.ipynb
@@ -73,7 +73,7 @@
     "\n",
     "The statespace module is designed to make it easy for users to create their own statespace models. At its core, a statspace model is just a system of two linear equatons:\n",
     "\n",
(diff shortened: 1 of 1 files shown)


jessegrabowski (approved):
Looks great! Thanks for fixing this!

PR #420, opened 2025-01-29
fix: progress bar display

aphc14:
Issue
- Progress bar percentage is not refreshed after the pathfinder finishes, leaving the completion percentage less than 100%

Changes
- Progress bar percentage is refreshed after completion to display progress of 100%
- Add progress bar details of the number of paths complete. Users can opt to stop progress (via Keyboard Interrupt) knowing the number of paths completed rather than as a percentage.


aphc14:
I think parts of the pymc-extras/.github/workflows/pypi.yml needs to be changed: from `actions/download-artifact@v3` to `actions/download-artifact@v4`

zaxtax (approved):

PR #422, opened 2025-02-05
Bugfixes and Improvements for MAP and Laplace

jessegrabowski:
Specifically responding to suggestions by @theorashid,  but the code is also quite ugly in general. I'd like to think about how it could be done better.


jessegrabowski:
rebased to group commits logically. I ended up touching code unrelated to the stated PR to get tests to pass, so this should not be squashed when merged.

zaxtax (commented):
Other than minor changes, looks good

PR #426, opened 2025-02-17
Fix group selection in `sample_posterior_predictive` when `predictions=True` is passed in kwargs

butterman0:
### Summary
Fixes hard-coded group selection in `sample_posterior_predictive` which unnecessarily restricts usage of predict functions. Previously, if `predictions=True` (ideally set in `pm.sample_posterior_predictive' when predicting out-of-sample) is passed to the predict functions the function always extracted from `"posterior_predictive"` which doesn't exist in the dataframe when `predictions = True`.

### Changes
Selects appropriate group depending if `predictions` is passed.

### DIFF
--- a/pymc_extras/model_builder.py
+++ b/pymc_extras/model_builder.py
@@ -650,8 +650,11 @@ class ModelBuilder:
             if extend_idata:
                 self.idata.extend(post_pred, join="right")
 
+        # Determine the correct group dynamically
+        group_name = "predictions" if kwargs.get("predictions", False) else "posterior_predictive"
+
         posterior_predictive_samples = az.extract(
-            post_pred, "posterior_predictive", combined=combined
+            post_pred, group_name, combined=combined
         )
 
         return posterior_predictive_samples


butterman0:
I'm sorry I haven't opened an issue first - I thought it was such a minor change that it wasn't necessary. This is also my first contribution so not 100% on the process!

pymc_extras/model_builder.py
```diff
@@ -650,8 +650,11 @@ def sample_posterior_predictive(self, X_pred, extend_idata, combined, **kwargs):
             if extend_idata:
                 self.idata.extend(post_pred, join="right")
 
+        # Determine the correct group dynamically
+        group_name = "predictions" if kwargs.get("predictions", False) else "posterior_predictive"
```
ricardoV94, line 654: Would it be better to make `predictions` an explicit kwarg (with the same default as PyMC) and use that directly?

butterman0: Yes I initially had that! Although I wasn't sure what was best practice. It is nice to make it explicit, but it means passing `predictions=False` as explicit args through the `predict` method and then to `sample_posterior_predictive` which is used in other methods - although this shouldn't be a problem if keeping the same default as PyMC as you suggest.

butterman0: In fact, the class method `sample_posterior_predictive` is called twice, on both occasions it is for prediction: methods `predict` and `predict_posterior`.

I think I would argue that in this case we would like the default to be `predictions=True` (as opposed to the pymc [pm.sample_posterior_predictive](https://www.pymc.io/projects/docs/en/latest/api/generated/pymc.sample_posterior_predictive.html) default) since it is always used for prediction. 

When False, the `posterior_predictive` group is overridden meaning we would have to run `fit` or `sample_model` again if we wanted to do posterior predictive checks?

pymc_extras/model_builder.py
```diff
@@ -559,7 +560,7 @@ def predict(
         """
 
         posterior_predictive_samples = self.sample_posterior_predictive(
-            X_pred, extend_idata, combined=False, **kwargs
+            X_pred, extend_idata, predictions, combined=False, **kwargs
```
ricardoV94, line 563: pass by keyword to be on the safe side

pymc_extras/model_builder.py
```diff
@@ -723,7 +728,7 @@ def predict_posterior(
 
         X_pred = self._validate_data(X_pred)
         posterior_predictive_samples = self.sample_posterior_predictive(
-            X_pred, extend_idata, combined, **kwargs
+            X_pred, extend_idata, predictions, combined, **kwargs
```
ricardoV94, line 731: pass by keyword argument

ricardoV94 (commented):
The predictions argument should be mentioned in the docstrings now that it is explicit

pymc_extras/model_builder.py
butterman0: I was aiming to keep it in the same format as current implementation. e.g. x_pred, extend_idata and combined are not keywords - should these be changed as well?

pymc_extras/model_builder.py
```diff
@@ -624,7 +625,7 @@ def sample_prior_predictive(
 
         return prior_predictive_samples
 
-    def sample_posterior_predictive(self, X_pred, extend_idata, combined, **kwargs):
+    def sample_posterior_predictive(self, X_pred, extend_idata, predictions, combined, **kwargs):
```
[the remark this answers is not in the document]
butterman0: The other arguments do not have defaults. The `sample_posterior_predictive` is only called through the predict functions, which do have defaults.

Would you be able to explain why we would want `predictions` to have a default, when the other arguments do not?

pymc_extras/model_builder.py
butterman0: Just checking you agree with setting `predictions=True` as default @ricardoV94 ?

ricardoV94: Yeah makes sense in the predict oriented methods

pymc_extras/model_builder.py
butterman0: Hi @ricardoV94, let me know what you think and I can adjust.

butterman0:
Hi @ricardoV94, I've made the requested changes.

Two commits updating the doc strings.

Most recent commit passing by keyword argument and setting default as requested. 

I'm still a little unclear as to why we would treat the predictions argument differently to the other arguments in the `sample_posterior_predictive` method (i.. combined and extend_idata). Similarly, the same question as to why we would pass by keyword when calling the method when we don't with the other variables. 

Harry

ricardoV94:
We should always pass by keyword argument, but since you didn't write the previous code I didn't ask you to change those lines

ricardoV94 (commented):
Can you add a test that confirms this is working now?

butterman0:
pre-commit.ci autofix

butterman0:
Hi @ricardoV94, let me know if anything else.

ricardoV94 (approved):

ricardoV94:
@butterman0 sorry for the delay, running the tests, we can merge if they pass

butterman0:
pre-commit.ci autofix

butterman0:
Hi @ricardoV94, one main problem arose in the testing which was solved. There were a few more under the hood that arose when I added another test: to test the `predict_posterior` function. 

The `predict_posterior` function calls `_validate_data`, which requires a) a 2D array and b) returns a numpy array. Therefore I changed _data_setter in the test setup to handle this.

I assume `_validate_data` should actually be called in `predict` as well (I have another pull request open regarding this, which I will close now), so I added it. Subsequently, I updated the `predict` calls throughout the tests such that they all pass 2D arrays to `_validate_data` via `predict` with `data_setter` handling it as before.

Let me know if anything is off.

butterman0:
Hi @ricardoV94, I'm not totally sure what the problem was as all the tests are passing on my side. 

I am getting an error from an import in `__init__.py` which is unrelated to the tests I am running, since I don't need the imports. The error message is below, the tests run fine if you comment out the offending import.

`ImportError while importing test module '/home/harold/Code/pymc-extras/tests/test_model_builder.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniconda3/envs/spGDMM/lib/python3.12/site-packages/_pytest/python.py:493: in importtestmodule
    mod = import_path(
../../miniconda3/envs/spGDMM/lib/python3.12/site-packages/_pytest/pathlib.py:587: in import_path
    importlib.import_module(module_name)
../../miniconda3/envs/spGDMM/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1381: in _gcd_import
    ???
<frozen importlib._bootstrap>:1354: in _find_and_load
    ???
<frozen importlib._bootstrap>:1325: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:929: in _load_unlocked
    ???
../../miniconda3/envs/spGDMM/lib/python3.12/site-packages/_pytest/assertion/rewrite.py:185: in exec_module
    exec(co, module.__dict__)
tests/test_model_builder.py:25: in <module>
    from pymc_extras.model_builder import ModelBuilder
pymc_extras/__init__.py:20: in <module>
    from pymc_extras.inference import find_MAP, fit, fit_laplace, fit_pathfinder
pymc_extras/inference/__init__.py:17: in <module>
pymc_extras/inference/laplace_approx/laplace.py:38: in <module>
    from pytensor.tensor.optimize import minimize
E   ModuleNotFoundError: No module named  #'pytensor.tensor.optimize`

butterman0:
Hi @ricardoV94, what do you think of the updates?

I also am considering writing a python package for the biodiversity model I'm making using PyMC, and in particular model builder. There's a few design choices specifically regarding model builder I wanted to ask about to help make this. Where would be the best forum for that?

PR #427, opened 2025-02-18
Fix: importance_sampling=None produces error

aphc14:
**Fix handling of `importance_sampling="none"` or `None` option**

**Changes**
- Modified the handling of importance sampling parameter to treat `"none"` and `None` consistently
- Updated function signatures and documentation to clarify the importance sampling options

**Added/Modified Tests**
- When importance sampling is disabled (None), samples retain shape `(num_paths, num_draws_per_path, N)` (where `N` are the number of model parameters)
- Other importance sampling methods return samples of shape `(num_draws, N)`
- Tighten `np.testing.assert_allclose` tolerance as increasing `jitter` brings parameter estimates closer to the reference value

### DIFF
--- a/pymc_extras/inference/pathfinder/pathfinder.py
+++ b/pymc_extras/inference/pathfinder/pathfinder.py
@@ -156,7 +156,7 @@ def convert_flat_trace_to_idata(
     postprocessing_backend: Literal["cpu", "gpu"] = "cpu",
     inference_backend: Literal["pymc", "blackjax"] = "pymc",
     model: Model | None = None,
-    importance_sampling: Literal["psis", "psir", "identity", "none"] = "psis",
+    importance_sampling: Literal["psis", "psir", "identity"] | None = "psis",
 ) -> az.InferenceData:
     """convert flattened samples to arviz InferenceData format.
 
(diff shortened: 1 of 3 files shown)


pymc_extras/inference/pathfinder/pathfinder.py
```diff
@@ -1690,6 +1697,16 @@ def fit_pathfinder(
     """
 
     model = modelcontext(model)
+
+    valid_importance_sampling = {"psis", "psir", "identity", None}
+
+    if importance_sampling is not None:
+        importance_sampling = importance_sampling.lower()
+        importance_sampling = None if importance_sampling == "none" else importance_sampling
```
fonnesbeck, line 1705: I thought you had globally removed the use of "none"?

aphc14: Good point. I thought about allowing a string for `None` as `"none"` to keep it consistent with the other `importance_sampling` options (i.e., `"psis"`, `"psir"`, `"identity"`). Come to think of it, `None` would be consistent with Python APIs in general with mixed argument types (`str | None`, `int | None`, etc).

I'll send another commit to remove the use of "none".

zaxtax:
Is this good to merge?

aphc14:
Yes I believe so. @fonnesbeck  ?

zaxtax:
I just went ahead and merged this. Thanks @aphc14 !

aphc14:
Thanks @zaxtax !

PR #429, opened 2025-02-21
Add to requirements-dev.txt and MvNormalSVD method

aphc14:
closes #428

Let me know if opening a git issue wasn't necessary. Wasn't sure if I simple PR would do.

PR #431, opened 2025-03-01
Bump PyMC version requirement 

### LINKED ISSUE #430: BayesianVARMAX import leads to TypeError
Error:
TypeError: MvNormalRV.__init__() missing 1 required keyword-only argument: 'method'

Reprex:

```bash
mamba create -n sbvar python=3.10 pymc pandas matplotlib pymc-extras jupyter
```

```python
import pymc as pm
from pymc_extras.statespace import BayesianVARMAX
```

System:
Apple M1 Max Macbook Pro
System Version: macOS 15.3.1 (24D70)
Kernel Version: Darwin 24.3.0

Found when trying out code from [this](https://discourse.pymc.io/t/structural-bayesian-vector-autoregression/16603/4) S-BVAR discussion.

jessegrabowski:
Closes #430 
Related to #429 

The latest pytensor release breaks several places in the statespace code. I was using a rewrite to introduce `method='svd'` to multivariate_normal forward samples; this is no longer needed. There was also an `if Variable:` check, which now needs to be `if Variable is not None`.

### DIFF
--- a/pymc_extras/statespace/filters/distributions.py
+++ b/pymc_extras/statespace/filters/distributions.py
@@ -62,29 +62,8 @@ class MvNormalSVD(MvNormal):
     A JAX MvNormal robust to low-rank covariance matrices
     """
 
-    rv_op = MvNormalSVDRV()
-
-
-try:
-    import jax.random
-
-    from pytensor.link.jax.dispatch.random import jax_sample_fn
-
-    @jax_sample_fn.register(MvNormalSVDRV)
-    def jax_sample_fn_mvnormal_svd(op, node):
-        def sample_fn(rng, size, dtype, *parameters):
-            rng_key = rng["jax_state"]
-            rng_key, sampling_key = jax.random.split(rng_key, 2)
-            sample = jax.random.multivariate_normal(
-                sampling_key, *parameters, shape=size, dtype=dtype, method="svd"
-            )
-            rng["jax_state"] = rng_key
-            return (rng, sample)
-
-        return sample_fn
-
-except ImportError:
-    pass
+    # TODO: Remove this entirely on next PyMC release; method will be exposed directly in MvNormal
+    rv_op = MvNormalSVDRV(method="svd")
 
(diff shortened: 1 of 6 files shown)


pymc_extras/statespace/filters/distributions.py
```diff
@@ -62,29 +62,8 @@ class MvNormalSVD(MvNormal):
...
-            )
-            rng["jax_state"] = rng_key
-            return (rng, sample)
-
-        return sample_fn
-
-except ImportError:
-    pass
+    # TODO: Remove this entirely on next PyMC release; method will be exposed directly in MvNormal
+    rv_op = MvNormalSVDRV(method="svd")
```
ricardoV94, line 66: Don't you want to wait?

jessegrabowski: Not really, I have to work 100% in a locally patched version right now, and at least 2 people have already hit the issue.

jessegrabowski: Unless we're releasing over the next few days?

ricardoV94: That's the plan but not a big deal either way

ricardoV94:
Seems like we need to ignore the numpy import warning or pin tests to numpy<2.0. It's coming from pymc

jessegrabowski:
Why aren't tests failing on the pymc side? We also do `xfail_strict = true` over there and we're not filtering any warnings.

ricardoV94:
We're not failing on warnings there. It was easier her because we started from scratch with the testing suite

jessegrabowski:
oh duh it's `filterwarnings = ["error"]` that does it, not `xfail_strict`. I'm on a losing streak this week.

AlexAndorra (approved):
Thanks @jessegrabowski , this looks good! Shall I merge?

jessegrabowski:
yes please!

PR #434, opened 2025-03-15
Make index check on statespace data less strict

### LINKED ISSUE #384: Check on Index of data too strict in `statespace`
As discussed with @jessegrabowski , the `statespace` module has a check on indices of passed observed data that seems to strict: if you pass an integer Index, it will not accept it because it's technically not a RangeIndex:
```python
a = pd.DataFrame(
    index=np.arange(8), columns=["A", "B", "C", "D"], data=np.arange(32).reshape(8, 4)
)

mod = st.LevelTrendComponent(order=2, innovations_order=[0, 1])
ss_mod = mod.build(name="a")

initial_trend_dims, sigma_trend_dims, P0_dims = (
    ss_mod.param_dims.values()
)
coords = ss_mod.coords

with pm.Model(coords=coords) as model_1:
    P0_diag = pm.Gamma("P0_diag", alpha=5, beta=5)
    P0 = pm.Deterministic("P0", pyt.eye(ss_mod.k_states) * P0_diag, dims=P0_dims)

    initial_trend = pm.Normal("initial_trend", dims=initial_trend_dims)
    sigma_trend = pm.Gamma("sigma_trend", alpha=2, beta=50, dims=sigma_trend_dims)

    ss_mod.build_statespace_graph(
        a["A"],
        mode="JAX",
    )
IndexError: Expected pd.DatetimeIndex or pd.RangeIndex on data, found <class 'pandas.core.indexes.base.Index'>
```
Maybe the check can be relaxed to non-consecutive integer index?
Also note that even if casting the index above to RangeIndex, the model will throw a warning about not finding a TimeIndex in the data, which I think is expected if you have a RandeIndex already.

jessegrabowski:
Closes #384 

I was assuming that any pandas index of integers was a `RangeIndex`; this will now accept anything. 

One issue is that if the index has holes (like the index goes 0, 1, 3, 4, ...), it will just be totally ignored, and internally we'll treat it *as though* it was continuous. I might be able to check for this, but it might also be the type of thing that you have to trust users to do the right thing with. Open to suggestions.

### DIFF
--- a/pymc_extras/statespace/utils/data_tools.py
+++ b/pymc_extras/statespace/utils/data_tools.py
@@ -114,9 +114,6 @@ def preprocess_pandas_data(data, n_obs, obs_coords=None, check_column_names=Fals
         if not np.issubdtype(index.dtype, np.integer):
             raise IndexError("Provided index is not an integer index.")
 
-        if not index.is_monotonic_increasing:
-            raise IndexError("Provided index is not monotonic increasing.")
-
         index_diff = index.to_series().diff().dropna().values
         if not (index_diff == 1).all():
             raise IndexError("Provided index is not monotonic increasing.")
--- a/pymc_extras/statespace/utils/data_tools.py
+++ b/pymc_extras/statespace/utils/data_tools.py
@@ -87,18 +87,18 @@ def preprocess_pandas_data(data, n_obs, obs_coords=None, check_column_names=Fals
     col_names = data.columns
     _validate_data_shape(data.shape, n_obs, obs_coords, check_column_names, col_names)
 
-    if isinstance(data.index, pd.RangeIndex):
-        if obs_coords is not None:
-            warnings.warn(NO_TIME_INDEX_WARNING)
-        return preprocess_numpy_data(data.values, n_obs, obs_coords)
-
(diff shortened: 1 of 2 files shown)


ricardoV94:
For the holes can you check if `len = (max - min + 1)`?

AlexAndorra (changes requested):
Thanks @jessegrabowski ! Looks good, modulo Ricardo's comment.

> One issue is that if the index has holes (like the index goes 0, 1, 3, 4, ...), it will just be totally ignored, and internally we'll treat it as though it was continuous. I might be able to check for this, but it might also be the type of thing that you have to trust users to do the right thing with. Open to suggestions.

Should we error out and tell users they should pass in the missing data points so that they are inferred? Or at least display a warning that highlights that with this behavior, all data points are assumed to be linearly spaced?

jessegrabowski:
The thing already warns pretty aggressively, so I'm not really inclined to go that route (if anything I think we should go the other way and remove some of the useless warnings).

I don't have a good sense of whether an index `[1, 2, 3, 5]` is a valid time series index. What to do hinges on the answer to that question.

ricardoV94 (approved):

PR #437, opened 2025-03-21
Fix bug in fit_laplace when model has exactly one variable

jessegrabowski:
When a mean vector has exactly one element, `scipy.stats.multivariate_normal.rvs` squeezes the leading dimension, which was causing errors in `fit_laplace` for models with exactly one parameter.

This PR patches the behavior by adding a check for this corner case.


fonnesbeck (approved):
LGTM

ricardoV94:
Why not use the numpy one that doesn't have that stupid squeeze behavior?

Issue #402, opened 2024-12-10
Split CI in more jobs


ricardoV94:
CI runtime is now approaching 45 minutes. Would be nice to split it so we get feedback faster

williambdean:
Splitting by file was an easy win

williambdean:
Yes, here is the first commit to do it

https://github.com/pymc-labs/pymc-marketing/commit/181b34885c86cee1902d375c04904dec77583f90

ricardoV94:
I thouht you tried something on this repo already

williambdean:
This repo meaning pymc-extras? No

hash-data:
Hey @ricardoV94  can I take it ? 
Thanks

Hashcode-Ankit:
Hey @ricardoV94  @williambdean  is someone working on it?

williambdean:
Good question. Any more iterations to do? @ricardoV94

ricardoV94:
Isn't this done?

PR #439, opened 2025-03-25
Split test suite by folder

williambdean:
Related to #402

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,3 +104,12 @@ jobs:
           env_vars: TEST_SUBSET
           name: ${{ matrix.os }}
           fail_ci_if_error: false
+
+  all_tests:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    needs: [ ubuntu, windows ]
+    steps:
+      - name: Check build matrix status
+        if: ${{ needs.ubuntu.result != 'success' || needs.windows.result != 'success' }}
+        run: exit 1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
           - tests/model
           - tests/statespace
           - tests/distributions
-          - tests --ignore tests/model --ignore tests/statespace --ignore tests/distributions 
+          - tests --ignore tests/model --ignore tests/statespace --ignore tests/distributions
       fail-fast: false
     runs-on: ${{ matrix.os }}
     env:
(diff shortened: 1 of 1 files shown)


williambdean:
Is half the run time ok?

ricardoV94 (approved):

ricardoV94:
Ah we need an all_tests job or something to make required on github

.github/workflows/test.yml
```diff
@@ -21,7 +21,12 @@ jobs:
...
         python-version: ["3.10"]
         test-subset:
-          - tests
+          - tests/model
+          - tests/statespace/test_statespace.py
+          - tests/statespace/test_kalman_filter.py
+          - tests/statespace --ignore tests/statespace/test_statespace.py --ignore tests/statespace/test_kalman_filter.py
+          - tests/distributions
+          - tests --ignore tests/model --ignore tests/statespace --ignore tests/distributions
       fail-fast: false
```
williambdean, line 30: Doesn't this make all run to completion? 

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#handling-failures

@ricardoV94

ricardoV94: I need a single job that I can add in the github UI as `required` for merging. You see the old ones that were in the system: 
image

I don't want to have to type all the test names that now exist, and even that would be outdated when we change the config.

ricardoV94: This is how it looks in PyMC: https://github.com/pymc-devs/pymc/blob/5db3779b11a1e1d663447a3a78a6f1213b740409/.github/workflows/tests.yml#L382-L394

williambdean: done

.github/workflows/test.yml
```diff
@@ -87,10 +97,19 @@ jobs:
...
           env_vars: TEST_SUBSET
           name: ${{ matrix.os }}
           fail_ci_if_error: false
+
+  all_tests:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    needs: [ ubuntu, windows ]
+    steps:
+      - name: Check build matrix status
```
ricardoV94, line 113: I don't know why we use this name elsewhere doesn't make any sense

williambdean: Wht do you recommend. Make a suggestion

ricardoV94: Check all tests pass

williambdean: Done

williambdean:
Was this good to go?

PR #441, opened 2025-03-27
Allow closed form marginalization

ricardoV94:
This PR adds base functionality to expand `marginalize` to handle closed form cases.

It refactored the workflow so that when you call `pm.marginalize` we partition the graph into abstract `MarginalRVs` that group the marginalized and dependent RVs together. This is a mere (hierarchical) graph partitioning. Rewrites are then needed to try to find suitable options for the marginalized subgraphs. The cases that were supported before with enumeration still are, and I added an example of a Normal(Normal) with the inner one being marginalized.

I am not completly sure I want to proceed with the current approach. Delaying the rewrites until the logp is called doesn't give immediate feedback to the user. Perhaps better not to mix this with the logp derivation rewrites?

It's also cumbersome to work with nested marginalized RVs, because we don't have a great mechanism to trigger rewrites inside inner graphs, although we could implement such a graph rewriter.

I tried one another appproach that worked with tagging Ops that did the same task of partitioning the graph, but it became a mess when mixed with the other logprob rewrites. Maybe another hint that it's best to do these two things separately?

### DIFF
--- a/pymc_extras/model/marginal/__init__.py
+++ b/pymc_extras/model/marginal/__init__.py
@@ -1,0 +1,1 @@
+import pymc_extras.model.marginal.rewrites  # Need import to register rewrites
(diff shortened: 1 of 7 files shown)


ricardoV94:
Closed in favor of #688

PR #443, opened 2025-03-29
Minor fix of blackjax import in fit_pathfinder function

aphc14:
Hi all, just a minor change for users who don't have JAX/BlackJAX installed.

Summary of changes:
* Moved the import statement for blackjax to ensure it is only imported when needed.
* Moved blackjax import statement to prevent import errors for users wanting to run pathfinder without JAX/BlackJAX installed.
* Updated the fit function to specify the return type as az.InferenceData.


zaxtax (approved):
LGTM!

PR #444, opened 2025-03-31
Added a PartialOrder transform

velochy:
This is an implementation of a partial order transform as discussed in https://discourse.pymc.io/t/feature-proposal-poset-order-transform/16753/7

I have tried to make it as generic as possible:
* It takes only the generating DAG and computes the transitive reduction
* It does not assume the DAG is given in topological order but instead handles sorting internally
* It converts the matrix to an adjacency list of incoming nodes, reducing computation for low-degree orders
* It can either take one single DAG for all data or allows specifying a separate DAG for each row

As I am new to contributing to both pymc and pymc-extras, I will likely need some help to get it acceptable.
So far, I have added the code and tests for it, but I don't know how to handle documentation yet. 
Also anything else that might need to happen before this can be merged.

### DIFF
--- a/pymc_extras/distributions/transforms/partial_order.py
+++ b/pymc_extras/distributions/transforms/partial_order.py
@@ -1,0 +1,148 @@
+#   Copyright 2025 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+import numpy as np
+import pytensor.tensor as pt
+
+from pymc.logprob.transforms import Transform
+
+__all__ = ["PartialOrder"]
+
+
+# Find the minimum value for a given dtype
+def dtype_minval(dtype):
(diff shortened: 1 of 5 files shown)


velochy:
Added to documentation too now. Please let me know if anything needs doing/improving

velochy:
@zaxtax Yes, let's try to get this one done next :)

I spent some time improving it further, adding a nice example to the documentation. 
I can't seem to figure out why the docustring for 'initvals' is not incorporated into the documentation, however. Maybe you can help with that?

zaxtax:
I think this relates to a larger refactor of making initvals live more on
the sampler than distribution and transform objects

https://github.com/pymc-devs/pymc/issues/7528

velochy:
> I think this relates to a larger refactor of making initvals live more on the sampler than distribution and transform objects [pymc-devs/pymc#7528](https://github.com/pymc-devs/pymc/issues/7528)

Interesting development. But I don't think its easy to incorporate transforms into it, as they can interact with the distribution they are transforming in weird ways. Case in point is the one here. For instance, if the underlying distribution has to be positive, we would need initvals that are both positive and respect the partial order provided. 

Current solution here is to allow .initvals fn to take lower and upper as parameters, so it can spit out values that are consistent with both. But the user still needs to be able to tell it that this is the case :)

pymc_extras/distributions/transforms/partial_order.py
```diff
@@ -0,0 +1,228 @@
...
+#   limitations under the License.
+import numpy as np
+import pytensor.tensor as pt
+
+from pymc.logprob.transforms import Transform
+
+__all__ = ["PartialOrder"]
+
+
+# Find the minimum value for a given dtype
```
zaxtax, line 22: These should probably just be docstrings

zaxtax (approved):
LGTM!

zaxtax:
I can't seem to get this merged. Can you try rebasing from `main` and hopefully that will unblock this?

PR #446, opened 2025-03-31
Dynamical Factor Models (DFM) Implementation (GSOC 2025)

andreacate:
# Dynamical Factor Models (DFM) Implementation

This PR provides a first draft implementation of Dynamical Factor Models as part of my application proposal for the [PyMC GSoC 2025 project](https://github.com/pymc-devs/pymc/wiki/GSoC-2025-projects#implement-new-statespace-models). A draft of my application report can be found at this [link](https://drive.google.com/drive/folders/1RVavnZ2q6V4OyS16F--0QuEmnFh-0K4b?usp=drive_link). 

## Overview
- Added `DFM.py` with initial functionality

## Current Status
This implementation is a work in progress and I welcome any feedback

## Next Steps
- Vectorize the construction of the transition and selection matrices (possibly by reordering state variables).
- Add support for measurement error.


zaxtax:
Looks interesting! Just say when you think it's ready for review

fonnesbeck:
cc @jessegrabowski

andreacate:
Thanks for the feedback!

I'm still exploring the best approach for implementing Dynamic Factor Models.
I've added a simple custom DFM model in a Jupyter notebook, which I plan to use as a prototype and testing tool while developing the main BayesianDynamicFactor class.

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,388 @@
...
+        self.ssm["initial_state", :] = x0
+
+        # initial covariance
+        P0 = self.make_and_register_variable(
+            "P0", shape=(self.k_states, self.k_states), dtype=floatX
+        )
+
+        self.ssm["initial_state_cov", :, :] = P0
+
+        # TODO vectorize the design matrix
```
jessegrabowski, line 299: You're going to have to double-check all of these matrix constructions if you re-ordered the states.

jessegrabowski (changes requested):

pymc_extras/statespace/models/DFM.py
```diff
@@ -274,6 +281,10 @@ def param_info(self) -> dict[str, dict[str, Any]]:
                 "shape": (self.k_endog, self.k_endog),
                 "constraints": "Positive Semi-definite",
             },
+            "sigma_obs": {
+                "shape": (self.k_endog,),
+                "constraints": "Positive Semi-definite",
```
jessegrabowski, line 286: Should just be "Positive" since it's not a matrix

pymc_extras/statespace/models/DFM.py
```diff
@@ -456,4 +468,11 @@ def build_ar_block_matrix(ar_coeffs):
...
         # Observation covariance matrix
-        self.ssm["obs_cov", :, :] = 0.0
+        if self.measurement_error:
+            sigma_obs = self.make_and_register_variable(
+                "sigma_obs", shape=(self.k_endog,), dtype=floatX
+            )
+            self.ssm["obs_cov", :, :] = pt.diag(sigma_obs)
+        else:
+            # If measurement error is not used, set obs_cov to zero
+            self.ssm["obs_cov", :, :] = 0.0
```
jessegrabowski, line 478: This isn't needed, all matrices are initialized to zero by default

jessegrabowski (changes requested):
Left  some comments. I didn't look over the tests because they still seem like WIP, but seem to be on the right track!

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,510 @@
...
+        self.exog = exog
+        # TODO add exogenous variables support?
+
+        # Determine the dimension for the latent factor states.
+        # For static factors, one might use k_factors.
+        # For dynamic factors with lags, the state might include current factors and past lags.
+        # If factor_order is 0, we treat the factor as static (no dynamics),
+        # but it is still included in the state vector with one state per factor.
+        # Factor_ar paramter will not exist in this case.
+        k_factor_states = sum(max(order, 1) for order in self.factor_order)
```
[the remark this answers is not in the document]
andreacate: If factor_order = 1, we would have just 1 state.
The idea was that if factor_order = p, we would have p-1 state related to this factor (to avoid to have a larger matrix, which since the p_th state will be not useful). 
Just in the case of factor_order = 0, we need a state since otherwise we will have trouble in the design matrix, but since there will be no ar_param for that factor, in the transition matrix there will be a 0.

So the formula for computing the number of states is: max(1,p-1) where p= factor_order.

Tell me if it sounds good

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+    def __init__(
+        self,
+        k_factors: int,
+        factor_order: int,
+        k_endog: int | None = None,
+        endog_names: Sequence[str] | None = None,
+        k_exog: int | None = None,
+        exog_names: Sequence[str] | None = None,
+        shared_exog_states: bool | None = None,
+        exog_innovations: bool | None = None,
```
jessegrabowski, line 279: This can have a default value of False, as can the other boolean arguments.

You only need to do the `type | None = None` thing if `type` is a mutable type, like a string or list. The issue is that these defaults can be mutated in-place, leading to unexpected side-effects. With non-mutable types like tuple, float, int, bool etc, you don't have to worry about this.

Here's an example of why it matters:

```py
def x(l: list[str] = ['a']):
    return l
res = x() 
res # ['a'], unsurprising
res.append('b') # Innocently keep working with the output
x() # ['a', 'b'], default argument modified in-place! 
```

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+        self.endog_names = endog_names
+        self.k_endog = k_endog
+        self.k_factors = k_factors
+        self.factor_order = factor_order
+        self.error_order = error_order
+        self.error_var = error_var
+        self.error_cov_type = error_cov_type
+
+        if k_exog is None and exog_names is None:
+            self._exog = False  # flag for presence of exogeneous variables
```
jessegrabowski, line 302: you can do `self._exog = self.k_exog > 0` once, after all the checks

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+            self.k_exog_states = k_exog * k_endog if not shared_exog_states else k_exog
+
+        self.exog_names = exog_names
+        self.k_exog = k_exog
+
+        # TODO add exogenous variables support
+        # I start implementing a version of exog support based on pymc_extras/statespace/models/structural/components/regression.py
+        # exog_innovations control if the beta coefficient follows a random walk
+        # shared_exog_states control if the exogenous states are shared across equations
+        # I tested the case of shared_exog_states=True and exog_innovations=False vs the stats case by looking at trajectory and everything works well
```
jessegrabowski, line 323: Excellent! But no need to keep the comments. Make a unittest to reproduce your test.

Keep a comment noting the difference between the statsmodels implementation (in the obs vector) and ours (in the state vector)

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+        # but it is still included in the state vector with one state per factor.
+        # Factor_ar paramter will not exist in this case.
+        k_factor_states = max(self.factor_order, 1) * k_factors
+
+        # Determine the dimension for the error component.
+        # If error_order > 0 then we add additional states for error dynamics, otherwise white noise error.
+        k_error_states = k_endog * error_order if error_order > 0 else 0
+
+        # Total state dimension
+        k_states = k_factor_states + k_error_states + (self.k_exog_states if self._exog else 0)
```
jessegrabowski, line 338:
```suggestion
        k_states = k_factor_states + k_error_states + self.k_exog_states
```

`k_exog_states` is already zero if self._exog is False right?

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+
+        # Number of independent shocks.
+        # Typically, the latent factors introduce k_factors shocks.
+        # If error_order > 0 and errors are modeled jointly or separately, add appropriate count.
+        k_posdef = (
+            k_factors
+            + (k_endog if error_order > 0 else 0)
+            + (self.k_exog_states if self._exog else 0)
```
jessegrabowski, line 346: see above

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+            names.remove("error_ar")
+        if self.error_cov_type == "unstructured":
+            names.remove("error_sigma")
+            names.append("error_cov")
+        if not self.measurement_error:
+            names.remove("sigma_obs")
+        if self._exog:
+            names.append("beta")
+            if self.exog_innovations:
+                names.append("beta_sigma")
```
jessegrabowski, line 383: Either do everything with remove or append, it's too confusing to mix them

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+                "constraints": "Positive",
+            },
+            "beta": {
+                "shape": (self.k_exog_states if self.k_exog is not None else 0,),
+                "constraints": None,
+            },
+            "beta_sigma": {
+                "shape": (self.k_exog_states if self.k_exog is not None else 0,),
+                "constraints": "Positive",
+            },
```
jessegrabowski, line 432: These shapes should also change base on `shared_exog_state` I think

jessegrabowski: also k_exog should never be `None`, you should check that `self.k_exog > 0`, or you use `_exog` flag

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+            for i in range(self.k_endog):
+                for lag in range(self.error_order):
+                    names.append(f"L{lag}.error_{i}")
+
+        if self._exog:
+            if self.shared_exog_states:
+                # Shared exogenous states
+                for i in range(self.k_exog):
+                    names.append(
+                        f"exog_{i}.shared"
```
jessegrabowski, line 463:
```suggestion
                        f"exog_{i}[shared]"
```

We're using this square bracket notation for now

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+                    )  # better to call beta? But could make confusion between parameter of the model and name of the state
+            else:
+                # Exogenous states
+                for i in range(self.k_exog):
+                    for j in range(self.k_endog):
```
jessegrabowski, line 468:
```suggestion
                    for endog_name in self.endog_names:
```

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+        # Idiosyncratic error states
+        if self.error_order > 0:
+            for i in range(self.k_endog):
+                for lag in range(self.error_order):
+                    names.append(f"L{lag}.error_{i}")
+
+        if self._exog:
+            if self.shared_exog_states:
+                # Shared exogenous states
+                for i in range(self.k_exog):
```
jessegrabowski, line 461: Replace this loop with `names.extend([exog_{i}[shared] for i in range(self.k_exog)])`

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+                    names.append(
+                        f"exog_{i}.shared"
+                    )  # better to call beta? But could make confusion between parameter of the model and name of the state
+            else:
+                # Exogenous states
+                for i in range(self.k_exog):
```
jessegrabowski, line 467:
```suggestion
for exog_name in self.exog_names:
```

jessegrabowski: Also see above, replace the loops with a list comprehension + extend

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+                    for j in range(self.k_endog):
+                        names.append(
+                            f"exog_{i}.endog_{j}"
```
jessegrabowski, line 470:
```suggestion
                            f"beta_exog_name[endog_name]"
```

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+                        )  # better to call beta? But could make confusion between parameter of the model and name of the state
```
jessegrabowski, line 471: yes, I prefer `beta_{exog_name}` or `coef_{exog_name}`.

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+    def coords(self) -> dict[str, Sequence]:
+        coords = make_default_coords(self)
+        # Add factor dimensions
+        coords[FACTOR_DIM] = [f"factor_{i+1}" for i in range(self.k_factors)]
+
+        # AR parameter dimensions - add if needed
+        if self.factor_order > 0:
+            coords[AR_PARAM_DIM] = list(range(1, (self.factor_order * self.k_factors) + 1))
+
+        # If error_order > 0
```
jessegrabowski, line 492: delete

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+    def observed_states(self) -> list[str]:
+        """
+        Returns the names of the observed states (i.e., the endogenous variables).
+        """
+        return self.endog_names
+
+    @property
+    def coords(self) -> dict[str, Sequence]:
+        coords = make_default_coords(self)
+        # Add factor dimensions
```
jessegrabowski, line 485: Remove superfluous comments

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+        # If error_order > 0
+        if self.error_order > 0:
+            if self.error_var:
+                coords[ERROR_AR_PARAM_DIM] = list(range(1, (self.error_order * self.k_endog) + 1))
+            else:
+                coords[ERROR_AR_PARAM_DIM] = list(range(1, self.error_order + 1))
+
+        if self._exog:
+            # Exogenous states
+            coords[EXOG_STATE_DIM] = list(range(1, self.k_exog_states + 1))
```
jessegrabowski, line 501: Use the state names

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+            return ["exog_data"]
+        return []
+
+    def make_symbolic_graph(self):
+        # Initial states
+
+        if not self._exog:
+            x0 = self.make_and_register_variable("x0", shape=(self.k_states,), dtype=floatX)
+        else:
+            x0_1 = self.make_and_register_variable(
```
jessegrabowski, line 584:
```suggestion
            initial_factor_loadings = self.make_and_register_variable(
```

jessegrabowski: Use a more descriptive variable name

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+                "x0", shape=(self.k_states - self.k_exog_states,), dtype=floatX
+            )
+            x0_2 = self.make_and_register_variable(
```
jessegrabowski, line 587:
```suggestion
            initial_betas = self.make_and_register_variable(
```

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+                "beta", shape=(self.k_exog_states,), dtype=floatX
+            )
+            x0 = pt.concatenate([x0_1, x0_2], axis=0)
```
jessegrabowski, line 590:
```suggestion
            x0 = pt.concatenate([initial_factor_loadings, initial_betas] axis=0)
```

jessegrabowski (changes requested):
I did a deeper pass on everything except the `build_symbolic_graph` method. I need to spend more time on that because it's gotten quite complex. 

I'll finish ASAP.

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+    Dynamic Factor Models
+
+    Parameters
+    ----------
+    k_factors : int
+        Number of latent factors.
+
+    factor_order : int
+        Order of the VAR process for the latent factors. If 0, the factors are treated as static (no dynamics).
+        Therefore, the state vector will include one state per factor and "factor_ar" will not exist.
```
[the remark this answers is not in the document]
andreacate: Yes sorry, maybe that was a bit misleading. The factor won't be autoregressive, but will still have stochastic innovation

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+        At least one of `k_endog` or `endog_names` must be provided.
+
+    k_exog : int, optional
+        Number of exogenous variables. If not provided, the model will not have exogenous variables.
+
+    exog_names : Sequence[str], optional
+        Names of the exogenous variables. If not provided, but `k_exog` is specified, default names will be generated as `exog_1`, `exog_2`, ..., `exog_k`.
+
+    shared_exog_states: bool, optional
+        Whether exogenous latent states are shared across the observed states. If True, there will be only one set of exogenous latent
```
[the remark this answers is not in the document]
andreacate: Yes, they are the betas coefficient

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,828 @@
...
+
+    measurement_error: bool, default True
+        If true, a measurement error term is added to the model.
+
+    verbose: bool, default True
+        If true, a message will be logged to the terminal explaining the variable names, dimensions, and supports.
+
+    Notes
+    -----
+    TODO: adding to notes, how exog variables are handled and add them in the example?
```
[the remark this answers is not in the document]
jessegrabowski: Yes that could be nice, but consider it a stretch goal, not a must-have

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,811 @@
...
+                "constraints": None,
+            },
+            "beta_sigma": {
+                "shape": (self.k_exog_states,),
+                "constraints": "Positive",
+            },
+        }
+
+        for name in self.param_names:
+            info[name]["dims"] = self.param_dims[name]
```
jessegrabowski, line 435: This line is erroring in the test I added, when `test_dynamic_factor_ar2_error_var_unstructured="unstructred"`. Check that `error_cov` is present in param_dims

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,811 @@
...
+
+    @property
+    def state_names(self) -> list[str]:
+        """
+        Returns the names of the hidden states: first factor states (with lags),
+        idiosyncratic error states (with lags), then exogenous states.
+        """
+        names = []
+
+        for i in range(self.k_factors):
```
jessegrabowski, line 447: Refactor these loops to be list comprehensions. Start with:

```py
names = [f'L{lag}.factor_{i}' for lag in range(max(self.factor_order ,1)) for i in range(k_factors)]
```

Then on subsequent loops you can use `names.extend([...])`, like you did for exog_flag

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,811 @@
...
+        if self.exog_flag:
+            coords[EXOG_STATE_DIM] = list(range(1, self.k_exog_states + 1))
+
+        return coords
+
+    @property
+    def shock_names(self):
+        shock_names = []
+
+        for i in range(self.k_factors):
```
jessegrabowski, line 500: same thing, refactor loop + append into list comp or extend

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,811 @@
...
+            coord_map["error_ar"] = (OBS_STATE_DIM, ERROR_AR_PARAM_DIM)
+
+        if self.error_cov_type in ["scalar"]:
+            coord_map["error_sigma"] = ()
+
+        elif self.error_cov_type in ["diagonal"]:
+            coord_map["error_sigma"] = (OBS_STATE_DIM,)
+
+        if self.error_cov_type == "unstructured":
+            coord_map["error_sigma"] = (OBS_STATE_DIM, OBS_STATE_AUX_DIM)
```
jessegrabowski, line 538: add case for error_cov

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,811 @@
...
+            )
+
+        if self.error_order > 0:
+            error_matrix = pt.eye(self.k_endog, dtype=floatX)
+            matrix_parts.append(error_matrix)
+            matrix_parts.append(
+                pt.zeros((self.k_endog, self.k_endog * (self.error_order - 1)), dtype=floatX)
+            )
+        if len(matrix_parts) == 1:
+            design_matrix = factor_loadings * 1.0
```
jessegrabowski, line 607: What does this line do?

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,811 @@
...
+        self.ssm["initial_state_cov", :, :] = P0
+
+        # Design matrix
+        factor_loadings = self.make_and_register_variable(
+            "factor_loadings", shape=(self.k_endog, self.k_factors), dtype=floatX
+        )
+
+        matrix_parts = [factor_loadings]
+
+        # Leaving space for higher-order factors
```
jessegrabowski, line 594: Can you make a more detailed comment here about the structure of the matrix? I guess it's ultimately going to have some block structure, like:

```bash
[ A  |  0  ]
[--------]
[ 0  |  B  ]
```
Just explain that A is the factor loadings with shape whatever, and B is whatever with shape whatever. Then you can say in the `factor_order > 1` case "Add the upper-right block of zeros"

jessegrabowski (changes requested):

tests/statespace/models/test_DFM.py
```diff
@@ -0,0 +1,682 @@
...
+
+        assert mod.param_names == expected_param_names
+        assert mod.param_dims == expected_param_dims
+        for k, v in expected_coords.items():
+            assert mod.coords[k] == v
+        assert len(mod.state_names) == 7
+        assert mod.observed_states == ["y0", "y1", "y2"]
+        assert len(mod.shock_names) == 5
+
+    def test_dynamic_factor_ar2_error_var_unstructured(self):
```
jessegrabowski, line 479: This test is currently failing

pymc_extras/statespace/models/DFM.py
```diff
@@ -0,0 +1,811 @@
...
+        # else: obs_cov remains zero (no measurement noise and idiosyncratic noise captured in state)
+        else:
+            if self.measurement_error:
+                # TODO: check this decision
+                # in this case error_order = 0, so there is no error term in the state, so the sigma error could not be added there
+                # Idiosyncratic + measurement error
+                sigma_obs = self.make_and_register_variable(
+                    "sigma_obs", shape=(self.k_endog,), dtype=floatX
+                )
+                total_obs_var = error_sigma**2 + sigma_obs**2
```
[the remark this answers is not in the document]
andreacate: Your last 2 comments I think are related to this doubt: how should we handle the error_sigma/error_cov when the error_order=0. 
In our implementation—which replicates what is done in Stats—when error_order = 0 there is no error term that is included in the state vector. This means I cannot add the error_sigma term to the state_cov, since its shape does not account for error terms. My first thought was to add it to the observation equation instead, but I don’t think that’s the right approach.

Conceptually, when error_order = 0, we should interpret it as having only the standard innovations (i.e., normal errors) on each endogenous variable, without introducing an explicit term in the state.

I have seen that in Stats, when error_order = 0, no additional states or innovations are introduced for the error term.

```python
       # Calculate the number of states
        k_states = self._factor_order
        k_posdef = self.k_factors
        if self.error_order > 0:
            k_states += self._error_order
            k_posdef += k_endog
```

jessegrabowski:
The notebook also looks great! 

Could you add some more headings, and motivate all the analysis that you're doing with some commentary? Make sure the piece connect together clearly. I'd move the Bayesian latent factor graph above the Coincident index. 

I would also suggest you comment on the fact that the optimizer doesn't converge in the statsmodels model. It ends up being "close enough", but make some hay out of the fact that MCMC doesn't "converge", it explores all equiprobable solutions, which in this case is important because the model is only weakly identified.

andreacate:
<p>Yes sure, you are right. No I have not made decision about parameters, since I wanted just to replicate what was done in Stats. Cointegration was not present in Stats, I was just curious at the beginning. I can just delete the cells about cointegration</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-extras/pull/446/discussion/#comment-3239156418'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jessegrabowski (approved):
This was amazing work, congratulations on a very successful GSoC @andreacate 🥳 🎉

PR #447, opened 2025-04-01
Make fit_pathfinder more similar to fit_laplace and pm.sample

velochy:
* Add initvals to parameters
* Add observations and constants to return value (idata)
* Add both fit_laplace and fit_pathfinder to docs and make fit link to both
* Clean up __init__.py include structure

This arose out of https://discourse.pymc.io/t/pymc-extras-fit-compatibility-with-pm-sample/16763/5

### DIFF
--- a/pymc_extras/__init__.py
+++ b/pymc_extras/__init__.py
@@ -15,9 +15,7 @@ import logging
 
 from pymc_extras import gp, statespace, utils
 from pymc_extras.distributions import *
-from pymc_extras.inference.find_map import find_MAP
-from pymc_extras.inference.fit import fit
-from pymc_extras.inference.laplace import fit_laplace
+from pymc_extras.inference import find_MAP, fit, fit_laplace, fit_pathfinder
 from pymc_extras.model.marginal.marginal_model import (
     MarginalModel,
     marginalize,
--- a/pymc_extras/inference/pathfinder/pathfinder.py
+++ b/pymc_extras/inference/pathfinder/pathfinder.py
@@ -12,6 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+
 import collections
 import logging
 import time
@@ -24,7 +25,6 @@ from typing import Literal, TypeAlias
 
 import arviz as az
 import filelock
-import jax
 import numpy as np
 import pymc as pm
 import pytensor
@@ -43,7 +43,6 @@ from pymc.pytensorf import (
     find_rng_nodes,
     reseed_rngs,
 )
-from pymc.sampling.jax import get_jaxified_graph
 from pymc.util import (
     CustomProgress,
     RandomSeed,
@@ -64,6 +63,7 @@ from rich.text import Text
 # TODO: change to typing.Self after Python versions greater than 3.10
 from typing_extensions import Self
 
+from pymc_extras.inference.laplace import add_data_to_inferencedata
 from pymc_extras.inference.pathfinder.importance_sampling import (
     importance_sampling as _importance_sampling,
 )
@@ -98,6 +98,8 @@ def get_jaxified_logp_of_ravel_inputs(model: Model, jacobian: bool = True) -> Ca
     Function
         A JAX function that computes the log-probability of a PyMC model with ravelled inputs.
     """
+
+    from pymc.sampling.jax import get_jaxified_graph
 
     # TODO: JAX: test if we should get jaxified graph of dlogp as well
     new_logprob, new_input = pm.pytensorf.join_nonshared_inputs(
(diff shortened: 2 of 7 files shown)


zaxtax:
This looks alright to me. Thoughts @ricardoV94 ?

fonnesbeck:
CC @aphc14

pymc_extras/__init__.py
```diff
@@ -15,9 +15,7 @@
 
 from pymc_extras import gp, statespace, utils
 from pymc_extras.distributions import *
-from pymc_extras.inference.find_map import find_MAP
-from pymc_extras.inference.fit import fit
-from pymc_extras.inference.laplace import fit_laplace
+from pymc_extras.inference import *
```
ricardoV94, line 18: Can we use explicit imports?

pymc_extras/inference/pathfinder/pathfinder.py
```diff
@@ -1685,17 +1688,24 @@ def fit_pathfinder(
...
-    arviz.InferenceData
+    :class:`~arviz.InferenceData`
         The inference data containing the results of the Pathfinder algorithm.
 
     References
     ----------
     Zhang, L., Carpenter, B., Gelman, A., & Vehtari, A. (2022). Pathfinder: Parallel quasi-Newton variational inference. Journal of Machine Learning Research, 23(306), 1-49.
     """
 
+    if initvals is not None:
```
ricardoV94, line 1705: Shouldn't mutate the model, make a copy perhaps if there's no better way to just forward the initvals

velochy: I would normally agree. However, I tried it, but model.copy() does not produce a working model sometimes - most notably when any transformations are used. 

Should I use some other copy function?

aphc14:
> @ricardoV94 it currently also fails a test on including blackjax, which is currently included in pathfinder.py How would you like to handle it? Do I bring the include lower down so it is only included when used, or do I add it to the requirements of the package?

Yup, that is a good idea. closes #443

pymc_extras/inference/pathfinder/pathfinder.py
```diff
@@ -1685,17 +1688,24 @@ def fit_pathfinder(
...
+        for rv_name, ivals in initvals.items():
+            model.set_initval(model.named_vars[rv_name], ivals)
```
aphc14, line 1707: Might need to ensure that ivals is a support point for the RV. For example, x ~ Uniform(-1, 1) would have `nan` initial values with `model.set_initval(model.named_vars["x"], 2)`

zaxtax:
Maybe we make a requirements-test.txt file like arviz has?

https://github.com/arviz-devs/arviz/blob/main/requirements-test.txt

pymc_extras/inference/pathfinder/pathfinder.py
velochy: While in the ideal world, I would agree, in practice
a) It is very nontrivial to do as I understand, as the limits are not specified anywhere where they are easy to take
b) pm.sample does no such checks, and the goal of this PR is to be compatible with that

velochy:
> Yup, that is a good idea. closes #443

Considering you have a standing PR on this, does it not make more sense to just have your quickfix merged separately? 
I don't think it conflicts with anything I've done here

zaxtax:
Definitely!

pymc_extras/__init__.py
velochy: done

zaxtax:
Try to rebase your PR. If it still fails, that's not on you to resolve

pymc_extras/inference/pathfinder/pathfinder.py
zaxtax: @velochy you want the `clone_model` function https://github.com/pymc-devs/pymc/blob/2842401f95de74ab37b7750cff455af28cddaffa/pymc/model/fgraph.py#L374

velochy: Thank you for that @zaxtax 
Updated the code. Also added a test to check if initvals are used correctly.

pymc_extras/inference/pathfinder/pathfinder.py
```diff
@@ -1662,12 +1665,12 @@ def fit_pathfinder(
...
         Method to apply sampling based on log importance weights (logP - logQ).
         Options are:
-        "psis" : Pareto Smoothed Importance Sampling (default)
-                Recommended for more stable results.
-        "psir" : Pareto Smoothed Importance Resampling
-                Less stable than PSIS.
-        "identity" : Applies log importance weights directly without resampling.
-        None : No importance sampling weights. Returns raw samples of size (num_paths, num_draws_per_path, N) where N is number of model parameters. Other methods return samples of size (num_draws, N).
+
+        - **"psis"** : Pareto Smoothed Importance Sampling (default). Usually most stable.
```
zaxtax, line 1669: Can you remove these ** characters? They aren't consistent the style used elsewhere.

zaxtax:
Can you also move the import of `jax` into the relevant functions in the pathfinder file?

pymc_extras/inference/pathfinder/pathfinder.py
velochy: Done

zaxtax:
I'm not sure why the build source distribution test is still unhappy. Can you investigate why importing jax still happens?

zaxtax (approved):

zaxtax:
Just giving others one last chance to chime in before I merge it 😀

PR #449, opened 2025-04-04
Add local level example

victorgarcia98:
To get familiar with the state space description, I wanted to implement the most simple SS model: the local level. Moreover, I show the StatsModels equivalent model next to it.


AlexAndorra (changes requested):
Thanks @victorgarcia98 ! Just left a few comments in Review NB. A broader question is: what do you envision this example for? Which use-cases?
Also, be careful because you also edited `pymc_extras/statespace/core/statespace.py`, and it seems to be accidental as it's adding a fixed shape to parameters

PR #450, opened 2025-04-04
Batched Kalman Filters

AlexAndorra:
Add batch dimension feature to Kalman filters, to enable fitting multiple time series at the same time


AlexAndorra:
Pushed latest progress: working with Filter, not yet with Smoother

PR #451, opened 2025-04-07
Use `set_data` in forecast method

### LINKED ISSUE #424: State Space: Exogenous Vars - Conflicting dimensions for time (bug?)
Hey PyMC & @jessegrabowski 

I’m trying to use state space’s forecast method with exogenous variables, but I keep getting this error:

>     ValueError: conflicting sizes for dimension ‘time’: length 1107 on the data but length 8 on coordinate ‘time’

My scenario: I’m passing exogenous indicator variables totaling 18 columns and 8 rows long, for an 8 day forecast. The period I’m forecasting is not at the -very- end of my time series. I’m testing my exogenous variable’s ability to reduce variance on events and holidays.

Time series info:

    Length: 1107 data points
    Start: 2022-01-19
    End: 2025-01-29

Relevant code:

```
// Define start date and forecast period
start_date, n_periods = pd.to_datetime("2024-1-25"), 8

// Extract exogenous indicator variables for the forecast period
scenario = {'data_exog': pd.DataFrame(df_holidays.loc[start_date:].iloc[:n_periods].to_numpy(dtype=float), 
                                      columns=df_holidays.columns)}

// Generate the forecast
forecasts = ss_mod.forecast(trace, start=start_date, periods=n_periods, scenario=scenario)
```
My second line of code is extracting the relevant portion of my exogenous variables and assigning it to the scenario.

Am I getting this error because I’m not forecasting at the very end of my time series? (edit: dumb question, I’ll test this later.)

Conda Environment-
PyMC Version: 5.20.1
PyMC Extras Version: 0.2.3
Python Version: Python 3.11.11

Thanks,
Roy

jessegrabowski:
Closes #424 

There was some really complicated logic for clone-replacing data variables in the `forecast` method for absolutely no reason. For certain models, this meant that the data was not being correctly updated, and coordinates didn't match after sampling.

Just doing a basic `pm.set_data` does the trick just fine.

### DIFF
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -1027,6 +1027,9 @@ class PyMCStateSpace:
             provided when the model was built.
         data_dims: str or tuple of str, optional
             Dimension names associated with the model data. If None, defaults to ("time", "obs_state")
+        scenario: dict[str, pd.DataFrame], optional
+            Dictionary of out-of-sample scenario dataframes. If provided, it must have values for all data variables
+            in the model. pm.set_data is used to replace training data with new values.
 
         Returns
         -------
@@ -2060,6 +2063,7 @@ class PyMCStateSpace:
 
         with pm.Model(coords=temp_coords) as forecast_model:
             (_, _, *matrices), grouped_outputs = self._kalman_filter_outputs_from_dummy_graph(
+                scenario=scenario,
                 data_dims=["data_time", OBS_STATE_DIM],
             )
 
@@ -2072,17 +2076,6 @@ class PyMCStateSpace:
             P0 = pm.Deterministic(
                 "P0_slice", cov[t0_idx], dims=cov_dims[1:] if cov_dims is not None else None
             )
-
-            if scenario is not None:
-                sub_dict = {
-                    forecast_model[data_name]: pt.as_tensor_variable(
-                        scenario.get(data_name), name=data_name
-                    )
-                    for data_name in self.data_names
-                }
-
-                matrices = graph_replace(matrices, replace=sub_dict, strict=True)
-                [setattr(matrix, "name", name) for name, matrix in zip(MATRIX_NAMES[2:], matrices)]
 
             _ = LinearGaussianStateSpace(
                 "forecast",
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -1570,8 +1570,10 @@ class PyMCStateSpace:
                 raise ValueError(
                     "Integer start must be within the range of the data index used to fit the model."
                 )
-        if periods is None and end is None:
-            raise ValueError("Must specify one of either periods or end")
+        if periods is None and end is None and not use_scenario_index:
+            raise ValueError(
+                "Must specify one of either periods or end unless use_scenario_index=True"
+            )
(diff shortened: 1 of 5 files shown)


Dekermanjian:
I believe I have figured out what is happening. In both Roy's example and in my own, we are naming our observation dimension on the exogenous variables 'time' which is what is also coded into the `TIME_DIM`. This is making 1 coord be used for 2 different things. First for the observations of the in-sample data and then again for the out of sample data. When the forecast method is being called it will sample appropriately, however, at the end when the trace is being converted into an InferenceData object the time coord is trying to describe two different things. The first being the in-sample time data and then the same coord is trying to be assigned to the out of sample forecasts. If the forecasts and the in-sample fit time data are of equal length, then there is no exception and everything works fine. However, when that is not the case you will get the dimension mismatch exception because the length of the time dimension on the in-sample data is not equal to that on the out of sample data. 

A very simple fix is to have a check to not allow the exogenous data to use the same name for the coord that describes the number of observations when fitting the model. I can confirm that if I change the name from 'time' to 'n_obs' the forecast method will work for different lengths of forecasts.

Just to be specific changing this:
```
@property
    def data_info(self):
        """
        Information about Data variables that need to be declared in the PyMC model block.

        Returns a dictionary of data_name: dictionary of property-name:property description pairs. The return value is
        used by the ``_print_data_requirements`` method, to print a message telling users how to define the necessary
        data for the model. Each dictionary should have the following key-value pairs:
            * key: "shape", value: a tuple of integers
            * key: "dims", value: tuple of strings
        """
        return {"exogenous_data": {"shape": (None, self.k_exog), "dims": (TIME_DIM, "exog_dims")}}
```
to this
```
@property
    def data_info(self):
        """
        Information about Data variables that need to be declared in the PyMC model block.

        Returns a dictionary of data_name: dictionary of property-name:property description pairs. The return value is
        used by the ``_print_data_requirements`` method, to print a message telling users how to define the necessary
        data for the model. Each dictionary should have the following key-value pairs:
            * key: "shape", value: a tuple of integers
            * key: "dims", value: tuple of strings
        """
        return {"exogenous_data": {"shape": (None, self.k_exog), "dims": ('n_obs', "exog_dims")}}
```

Fixes the issue.

Is it acceptable to not allow a user to use the same coordinates for the in-sample exogenous number of observations and the in-sample endogenous number of observations?

pymc_extras/statespace/core/statespace.py
```diff
@@ -2060,6 +2065,7 @@ def forecast(
 
         with pm.Model(coords=temp_coords) as forecast_model:
             (_, _, *matrices), grouped_outputs = self._kalman_filter_outputs_from_dummy_graph(
+                scenario=scenario,
```
Dekermanjian, line 2068: Hey @jessegrabowski, sorry about the constant back and forth,  but I think this addition here fixes the bug. The tests still fail but the error message is different than the expected dimension mismatch it is an assertion error:

```
FAILED tests/statespace/test_statespace.py::test_foreacast_valid_index - UserWarning: Skipping `CheckAndRaise` Op (assertion: The first dimension of a time varying matrix (the time dimension) must be equal to the first dimension of the data (the time dimension).) as JAX tracing would remove it.
```

 Running the test `test_foreacast_valid_index` in a notebook works without any problems.

jessegrabowski:
Closed in favor of #466

PR #452, opened 2025-04-09
Minor fixes to modelbuilder class

butterman0:
Made following fixes:

`_validate_data' called for `predict_posterior` method but not `predict` method.  Should be called in both.

`self.is_fitted_ = False` set in init, but not changed when fitted. Decided to keep as potentially useful therefore set to True after 1) fitting 2) loading in new class with fitted data

### DIFF
--- a/pymc_extras/model_builder.py
+++ b/pymc_extras/model_builder.py
@@ -444,6 +444,7 @@ class ModelBuilder:
             sampler_config=json.loads(idata.attrs["sampler_config"]),
         )
         model.idata = idata
+        model.is_fitted_ = True
         dataset = idata.fit_data.to_dataframe()
         X = dataset.drop(columns=[model.output_var])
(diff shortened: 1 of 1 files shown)


zaxtax:
Looks decent! Make the lint errors go away and the unit tests in model builder pass, and I think we can merge this

butterman0:
pre-commit.ci autofix

butterman0:
Thanks @zaxtax, got autofix to do it! Let me know if anything else.

butterman0:
Hi @zaxtax, I rebased as there seems to be some problems with pytest. Anything else I need to do here?

butterman0:
Hi @zaxtax, I removed the commit to add the call to `_validate_data` as I needed it in #426 to fix testing errors. There would have been some hacking going on otherwise. Not sure if this is good practice, but there won't be any conflicts now!

PR #453, opened 2025-04-10
Fix failed test related to PR #443 by replacing `pt.batched_dot` with `pt.vectorize(pt.dot,...)`

aphc14:
Fix failed test from PR #443 and replace `pt.batched_dot` with `pt.einsum`

* Modified the calculation of `mu` in `bfgs_sample_dense` and `bfgs_sample_sparse` to use `einsum` instead of the deprecated `batched_dot` function.
* Added calculations for `H_inv` in `bfgs_sample_sparse` to improve readability.
* Renamed `compile_pymc` to `compile` in pathfinder functions to reflect API changes.
* Updated warning filters to ignore UserWarnings related to einsum subscripts.

### DIFF
--- a/pymc_extras/inference/fit.py
+++ b/pymc_extras/inference/fit.py
@@ -11,9 +11,10 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+import arviz as az
 
 
-def fit(method, **kwargs):
+def fit(method: str, **kwargs) -> az.InferenceData:
     """
     Fit a model with an inference algorithm
 
--- a/pymc_extras/inference/pathfinder/pathfinder.py
+++ b/pymc_extras/inference/pathfinder/pathfinder.py
@@ -40,7 +40,7 @@ from pymc.initial_point import make_initial_point_fn
 from pymc.model import modelcontext
 from pymc.model.core import Point
 from pymc.pytensorf import (
-    compile_pymc,
+    compile,
     find_rng_nodes,
     reseed_rngs,
 )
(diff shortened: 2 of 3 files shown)


zaxtax:
It looks like it still has some errors. Once this passes CI we can merge

aphc14:
Ahh damn I spoke too soon. The failures are mostly due to `pytensor.graph.utils.InconsistencyError: Multiple destroyers of g`, which I will look into sometime tomorrow.

I don't get this error when I run pytest locally (with the same package dependencies). If there's a better way to run tests locally that'll pick up on CI errors, please do share, as I'm keen to know if I'm missing anything!

zaxtax:
Make sure you are using the latest pytensor or whichever version CI is on.
If CI is using an old version, we can update it

zaxtax:
@aphc14 it might make sense while we figure out what caused this to have a more modest change where we filter the warning for `batched_dot`? 

The regression appears to have been introduced between pytensor-2.30.2 pytensor-2.30.3. So it might make sense to try to make a minimal failing example and open an issue there too!

ricardoV94:
The warning is there for performance reasons it means you're using the same letter for something that has a different shape (but broadcasts fine) which is less efficient than using a different letter. Also users will still see the warning even if you ignore in the tests which is not a great experience.

aphc14:
Thanks @zaxtax @ricardoV94 , that narrows it down quite a fair bit. Will attempt to resolve this within the next couple of hours 🤞

aphc14:
@zaxtax @ricardoV94 done :)

replaced:
```
mu = x - pt.einsum("ijk,ik->ij", H_inv, g) # causes error: Multiple destroyers of g
```

with:
```
batched_dot = pt.vectorize(pt.dot, signature="(ijk),(ilk)->(ij)")
mu = x - batched_dot(H_inv, pt.matrix_transpose(g[..., None]))
```

`pt.einsum` is fussier for some reason, even though `pt.vectorize(pt.dot,...)` and the previous `pt.batched_dot` work

zaxtax:
Good stuff! I think we should still try to find a minimal example that
triggers the previous error to file an issue on the pytensor repo

zaxtax (approved):

PR #455, opened 2025-04-12
Filter warning for batched_dot until we change it

zaxtax:
While we figure out how to remove deprecated functionality in #453  this PR should get `main` building properly again!

### DIFF
--- a/pymc_extras/inference/pathfinder/pathfinder.py
+++ b/pymc_extras/inference/pathfinder/pathfinder.py
@@ -485,6 +485,7 @@ def bfgs_sample_dense(
     shapes: L=batch_size, N=num_params, J=history_size, M=num_samples
     """
 
+    _warnings.simplefilter("ignore", category=FutureWarning)
     N = x.shape[-1]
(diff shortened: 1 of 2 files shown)


fonnesbeck (approved):
LGTM

ricardoV94 (changes requested):
Don't filter ignore on the tests but when calling the function internally. Users shouldn't see warnigs they have no control over (that's why we fail on the tests)

ricardoV94 (approved):
looks good if tests pass. Why are we calling compile_pymv though? That's just a name change... People seem to be getting the point of deprecation warning wrong :)

zaxtax:
Does a name change not warrant a DeprecationWarning? What else would we use
if we intend to remove `compile_pymc` eventually?

ricardoV94:
Yes it warrants, but here in extras we should just have updated, not ignored it. That's a silly reaction to the warning

zaxtax:
They are updated in https://github.com/pymc-devs/pymc-extras/pull/453

PR #457, opened 2025-04-12
Progress bar fix for PF

### LINKED ISSUE #456: Pathfinder progress bar inaccurate final state
When running pathfinder, the progress bar the final update appears to be receiving incorrect information. For example, the following run:

Image

ends with the following state:

Image

Notice that the time is inaccurate and it ends at 97% instead of 100%.

It also occurs for single-path pathfinder:

Image

pymc       : 5.20.1
pytensor   : 2.27.1
pymc_extras: 0.2.3

fonnesbeck:
The final state is now updated correctly for the progress bar when using Pathfinder.

Closes #456


zaxtax (approved):
LGTM!

Issue #445, opened 2025-03-31
pathfinder missing lbfgs_status


velochy:
In running pathfinder with a more complex generated model, I get the following error:

```
File ~/miniconda3/envs/salk/lib/python3.12/site-packages/pymc_extras/inference/pathfinder/lbfgs.py:190, in LBFGS.minimize(self, x0)
    [188](https://file+.vscode-resource.vscode-cdn.net/home/velochy/salk/sandbox/debug/kov1-full_fix_package/~/miniconda3/envs/salk/lib/python3.12/site-packages/pymc_extras/inference/pathfinder/lbfgs.py:188)     lbfgs_status = LBFGSStatus.CONVERGED
--> [190](https://file+.vscode-resource.vscode-cdn.net/home/velochy/salk/sandbox/debug/kov1-full_fix_package/~/miniconda3/envs/salk/lib/python3.12/site-packages/pymc_extras/inference/pathfinder/lbfgs.py:190) return history.x, history.g, history.count, lbfgs_status

UnboundLocalError: cannot access local variable 'lbfgs_status' where it is not associated with a value
```

Looking at the code in lbfgs.py
```
        # warnings and suggestions for LBFGSStatus are displayed at the end
        if result.status == 1:
            lbfgs_status = LBFGSStatus.MAX_ITER_REACHED
        elif (result.status == 2) or (history.count <= 1):
            if result.nit <= 1:
                lbfgs_status = LBFGSStatus.INIT_FAILED
            elif result.fun == np.inf:
                lbfgs_status = LBFGSStatus.DIVERGED
        else:
            lbfgs_status = LBFGSStatus.CONVERGED

        return history.x, history.g, history.count, lbfgs_status
```
There is a clear gap in the status==2 block that is causing this, but I don't understand the lbfgs algorithm nearly well enough to know what status that gap should be filled with.

aphc14:
Hi @velochy, thanks for raising this issue. I'll try to address this one quickly. Can you set a breakpoint on this line:

`elif (result.status == 2) or (history.count <= 1):`

and share what output you get for `result` and `history.count`?

velochy:
I'm sorry, but I have managed to misplace the model I actually got this error on :(

I do have cases of "All paths failed." and "BUG: Failed to iterate!" if you are interested.
Although they are generated models so it is a bit of work to get them working as stand-alone models, I'd be happy to try to get minimal examples if you (or someone else) are interested in taking them on

[...]

PR #461, opened 2025-04-18
Fix LBFGS iteration conditions and status handling

aphc14:
hey all,

This PR fixes https://github.com/pymc-devs/pymc-extras/issues/445.

During LBFGS iteration, filtering conditions are applied to exclude non-finite values and sharp gradient updates. This PR ensures that when the majority of iterations are excluded, and not enough iterations are stored to run a single PF, that the overall  algorithm continues to the next single PF.

Summary:
- Added entry_condition_met method that combines the logic for determining if LBFGS iterations should be stored.
- Added LBFGS statuses: LOW_UPDATE_MASK_RATIO to inform user when the majority of LBFGS iters aren't being used.
- Added LBFGS statuses: INIT_FAILED_LOW_UPDATE_MASK to inform user when LBFGS failed to initialised due to failing to meet the update conditions.
- Added status messages for LOW_UPDATE_MASK_RATIO and INIT_FAILED_LOW_UPDATE_MASK.
- Renamed LBFGSStatus.DIVERGED to LBFGSStatus.NON_FINITE.
- Implemented a test for unstable LBFGS update mask scenarios, adding robustness against rejected iterations.

### DIFF
--- a/pymc_extras/inference/pathfinder/lbfgs.py
+++ b/pymc_extras/inference/pathfinder/lbfgs.py
@@ -52,7 +52,7 @@ class LBFGSHistoryManager:
         self.count = 0
 
         value, grad = self.value_grad_fn(self.x0)
-        if np.all(np.isfinite(grad)) and np.isfinite(value):
+        if self.entry_condition_met(self.x0, value, grad):
             self.add_entry(self.x0, grad)
 
     def add_entry(self, x: NDArray[np.float64], g: NDArray[np.float64]) -> None:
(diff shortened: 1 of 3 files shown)


fonnesbeck (approved):
LGTM

aphc14:
Demo of the fix, including cases when the number of LBFGS iterations is between 10 and 40 or more, but the stored iterations are 1 or 0.

Example after the fix:
[test_lbfgs_missing_gap_handle_nan.ipynb](https://gist.github.com/aphc14/4081019ac4a192a629053124da2df49b#file-test_lbfgs_missing_gap_handle_nan-ipynb)

Whereas previously, all scenarios would've raised an Exception before running PF to completion.

PR #467, opened 2025-05-02
Allow `method="basinhopping"` in `find_MAP` and `fit_laplace`

### LINKED ISSUE #454: Add `basinhopping` to allowed methods in `find_MAP` and `fit_laplace`
I added a pretty front-end to `scipy.optimize.bashinhopping` in better-optimize that should make it just easy to just drop into find_MAP. This would be nice to have for really fussy models with rugged likelihood landscapes.

jessegrabowski:
Closes #454 

This PR lets users use the `basinhopping` algorithm to fit PyMC models. This is in the class of simulated annealing algorithms, except that it also runs an inner optimization loop to find the nearest minimum after proposing a new starting position. 

API is pretty clean, if I do say so myself:

```py
    with pm.Model() as m:
        mu = pm.Normal("mu")
        sigma = pm.Exponential("sigma", 1)
        pm.Normal("y_hat", mu=mu, sigma=sigma, observed=rng.normal(loc=3, scale=1.5, size=100))

        optimized_point = find_MAP(
            method="basinhopping",
            use_grad=True,
            use_hess=False,
            use_hessp=True,
            progressbar=True,
            gradient_backend="pytensor",
            compile_kwargs={"mode": "JAX"},
            minimizer_kwargs=dict(method="trust-ncg"),
        )
```

Gives:
```
  Process                            Elapsed   Iteration   Step    Target   Objective     ||grad|| 
 ──────────────────────────────────────────────────────────────────────────────────────────────────
  Basinhopping   ━━━━━━━━━━━━━━━━━   0:00:02   100/100     0.617   0.500    186.98814     0.00000
  Minimize       ━━━━━━━━━━━━━━━━━   0:00:00   8/8         0.000   0.000    34964.56397   0.00000
```

The top progress bar shows the overall algorithm, the current step, the target step size, current best objective function, and the gradient at that current best. The bottom shows progress on the inner optimization loop, so it gets updated a lot. You can see here that the final inner optimization went somewhere stupid, so the objective value is very high.

### DIFF
--- a/pymc_extras/inference/find_map.py
+++ b/pymc_extras/inference/find_map.py
@@ -9,7 +9,7 @@ import pymc as pm
 import pytensor
 import pytensor.tensor as pt
 
-from better_optimize import minimize
+from better_optimize import basinhopping, minimize
 from better_optimize.constants import MINIMIZE_MODE_KWARGS, minimize_method
 from pymc.blocking import DictToArrayBijection, RaveledVars
 from pymc.initial_point import make_initial_point_fn
@@ -335,7 +335,7 @@ def scipy_optimize_funcs_from_loss(
 
 
 def find_MAP(
-    method: minimize_method,
+    method: minimize_method | Literal["basinhopping"],
     *,
     model: pm.Model | None = None,
     use_grad: bool | None = None,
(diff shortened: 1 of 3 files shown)


jessegrabowski:
I don't think the CI failure is our fault here, seems like a dask bug. I [opened an issue ](https://github.com/dask/dask/issues/11928) on their repo.

AlexAndorra (approved):
Super neat, thanks @jessegrabowski ! Should I merge despite the CI failure then?

jessegrabowski:
I'm going to try version pinning dask first

PR #468, opened 2025-05-02
Fix bug in `fit_MAP` when shared variables are used in graph

jessegrabowski:
Currently, `fit_MAP` fails on graphs with shared variables when `gradient_backend='jax'`. This is because we are discarding the pytensor function wrapper that helps work with them. This PR rewrites shared variables into constants using `_replace_shared_variables` in this case. This is the same approach used in `pymc.sampling.jax.get_jaxified_graph`

### DIFF
--- a/pymc_extras/inference/find_map.py
+++ b/pymc_extras/inference/find_map.py
@@ -15,6 +15,7 @@ from pymc.blocking import DictToArrayBijection, RaveledVars
 from pymc.initial_point import make_initial_point_fn
 from pymc.model.transform.optimization import freeze_dims_and_data
 from pymc.pytensorf import join_nonshared_inputs
+from pymc.sampling.jax import _replace_shared_variables
 from pymc.util import get_default_varnames
 from pytensor.compile import Function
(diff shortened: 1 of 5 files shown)


AlexAndorra (approved):

pymc_extras/inference/find_map.py
```diff
@@ -301,6 +301,14 @@ def scipy_optimize_funcs_from_loss(
         point=initial_point_dict, outputs=[loss], inputs=inputs
     )
 
+    # If we use pytensor gradients, we will use the pytensor function wrapper that handles shared variables. When
+    # computing jax gradients, we discard the function wrapper, so we can't handle shared variables --> rewrite them
+    # away.
+    if use_jax_gradients:
+        from pymc.sampling.jax import _replace_shared_variables
+
+        [loss] = _replace_shared_variables([loss])
```
ricardoV94, line 310: btw this is the sort of stuff that can lead to big constant foldings in JAX

ricardoV94 (approved):

PR #484, opened 2025-05-28
Expose multivariate normal `method` argument in post-estimation tasks

jessegrabowski:
Post-estimation sampling tasks, especially `sample_conditional_posterior`, can be really slow for long time series. One reason for this is that I had set the `method` argument of all the multivariate normal distributions internally to be `svd`, which is the most robust option. There are cases, however, where one *knows* the model is well behaved, so you can drop back to something faster (like `cholesky`) without any problem. This PR allows this choice.

The default is still `svd`. Ideally we'd check a couple covariance matrices from the provided `idata` and try to give the user a smart default, but I don't want to do that much work here.

### DIFF
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -1378,7 +1378,9 @@ class PyMCStateSpace:
              "predicted_prior", and "smoothed_prior".
         """
 
-        return self._sample_conditional(idata, "prior", random_seed, method, **kwargs)
+        return self._sample_conditional(
+            idata=idata, group="prior", random_seed=random_seed, method=method, **kwargs
+        )
 
     def sample_conditional_posterior(
         self, idata: InferenceData, random_seed: RandomState | None = None, method="svd", **kwargs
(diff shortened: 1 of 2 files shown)


pymc_extras/statespace/core/statespace.py
```diff
@@ -1109,6 +1109,7 @@ def _sample_conditional(
         group: str,
         random_seed: RandomState | None = None,
         data: pt.TensorLike | None = None,
+        method: str = "svd",
```
ricardoV94, line 1112: do you want to type hint the options and/or mention them on the docstrings? Potentially mention the reason for choosing some over the other?

pymc_extras/statespace/core/statespace.py
```diff
@@ -1130,6 +1131,11 @@ def _sample_conditional(
             Observed data on which to condition the model. If not provided, the function will use the data that was
             provided when the model was built.
 
+        method: str
+            Method used to compute draws from multivariate normal. One of "cholesky", "eig", or "svd". "cholesky" is
+            fastest, but least robust to ill-conditioned matrices, while "svd" is slow but extermely robust. Default
+            is "svd".
```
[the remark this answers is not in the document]
jessegrabowski: This is actually an interesting pattern, but we don't use it anywhere else so I don't want to do it here.

pymc_extras/statespace/core/statespace.py
jessegrabowski: Just pushed this without having read this comment, so yes :)

PR #470, opened 2025-05-04
Port Prior class from PyMC-Marketing

williambdean:
This ports the `Prior` class from PyMC-Marketing to `pymc_extras`.

The `deserialize` module is coming along for the ride as well.

Closes https://github.com/pymc-devs/pymc-extras/issues/448

CC: @drbenvincent

TODO: 

- [x] Get rid of any PyMC-Marketing references

PR #489, opened 2025-05-29
Add deserialize module

williambdean:
Will rebase one #470 is merged

### DIFF
--- a/pymc_extras/prior.py
+++ b/pymc_extras/prior.py
@@ -706,7 +706,6 @@ class Prior:
             dist = Prior("Normal", mu=0, sigma=1)
 
             dist.to_dict()
-            # {"dist": "Normal", "kwargs": {"mu": 0, "sigma": 1}}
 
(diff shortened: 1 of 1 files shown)


zaxtax:
Ok #470 is merged so I think after this is rebased and passes CI, we can merge it too

williambdean:
Seem like unrelated fails...

zaxtax:
Try rebasing from main and it should work now

pymc_extras/prior.py
```diff
@@ -685,6 +687,143 @@ def preliz(self):
...
+
+            dist = Prior(
+                "Normal",
+                mu=Prior("Normal"),
+                sigma=Prior("HalfNormal"),
+                dims="channel",
+            )
+
+            dist.to_dict()
+            # {
```
zaxtax, line 723: Can this be removed?

zaxtax:
Yep and then let's get this merged 😃

PR #490, opened 2025-05-30
Removing some spurious checks with the observation intercept has a time-varying dimension.

sangrey:
This PR fixes two issues in the PytensorRepresentation class that affect how time-varying parameters are validated and reshaped. 

Changes

    1. Corrected a bug in shape validation:
    The check for mismatched static shape of time-varying inputs (when X.ndim == 2) incorrectly compared X.shape[1:] instead of shape[1:].  X.shape is a symbolic tensor. 

    2.  X has two dimensions if it is time-varying & vector valued or if is not-tie-varying and not vector valued.  I added a check to ensure that we were in case two before adding additional dimensions.

### DIFF
--- a/pymc_extras/statespace/core/representation.py
+++ b/pymc_extras/statespace/core/representation.py
@@ -298,7 +298,7 @@ class PytensorRepresentation:
                     )
 
                 # Time varying vector case, check only the static shapes
-                if X.ndim == 2 and X.shape[1:] != expected_shape:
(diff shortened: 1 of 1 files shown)


sangrey:
Apologies for not getting back to you sooner,. I've been traveling, and busier than I expected  beyond that. It could be a bit before I would get the chance to split this up.  Also, it looks like you just merged a multivariate state-space model, correct?

PR #493, opened 2025-05-31
Pin pytensor to < 2.31

zaxtax:
pytensor introduced some regressions since 2.31

This pins the version below that while the following gets resolved

https://github.com/pymc-devs/pytensor/issues/1425


zaxtax:
No, this is just for the existing marginalize tests that used to pass and don't after the 2.31 release

PR #497, opened 2025-06-04
Drop Python 3.10 support

maresb:
In accordance with https://scientific-python.org/specs/spec-0000/ modern versions of preliz require Python 3.11. Since we depend on preliz, we are forcing users to install something ancient. In fact, this combination was only installable due to a [bug](https://github.com/conda-forge/better-optimize-feedstock/pull/10) in the better-optimize feedstock.


maresb:
@williambdean for some reason i can't tag you for review

PR #510, opened 2025-06-09
Forecast exogenous vars bug fix

Dekermanjian:
This PR addresses a bug in the Statespace module where erroneous forecasts were being produced when exogenous variables were present in the model. This PR addresses issue #491.

The issue stemmed from updating the model's data prior to running the model forward up to the forecast's initial time index as specified by the user. This caused the forecasts to be produced using incorrect x0 and P0 initializations at the forecast initial time index. 

The fix ensures that the model is run forward up to the forecast time index and the states (filtered, predicted or smoothed) at that point are frozen using all the observed data. After which new data replacements are made to generate forecasts.

### DIFF
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -2209,49 +2209,32 @@ class PyMCStateSpace:
             group_idx = FILTER_OUTPUT_TYPES.index(filter_output)
             mu, cov = grouped_outputs[group_idx]
 
-            if scenario is not None:
-                sub_dict = {
-                    forecast_model[data_name]: pt.as_tensor_variable(
-                        x=np.atleast_2d(self._exog_data_info[data_name]["value"].T).T,
-                        name=data_name,
-                    )
-                    for data_name in self.data_names
-                }
+            sub_dict = {
+                data_var: pt.as_tensor_variable(data_var.get_value(), name="data")
+                for data_var in forecast_model.data_vars
+            }
 
-                # Will this always be named "data"?
-                sub_dict[forecast_model["data"]] = pt.as_tensor_variable(
-                    np.atleast_2d(self._fit_data.T).T, name="data"
-                )
-            else:
-                # same here will it always be named data?
-                sub_dict = {
-                    forecast_model["data"]: pt.as_tensor_variable(
-                        np.atleast_2d(self._fit_data.T).T, name="data"
-                    )
-                }
+            replacements_diff = np.setdiff1d(
+                ar1=[*self.data_names, "data"], ar2=[k.name for k, _ in sub_dict.items()]
+            )
+            if replacements_diff.size > 0:
+                raise ValueError(f"{replacements_diff} data used for fitting not found!")
 
-            mu, cov = graph_replace([mu, cov], replace=sub_dict, strict=True)
+            mu_frozen, cov_frozen = graph_replace([mu, cov], replace=sub_dict, strict=True)
 
             x0 = pm.Deterministic(
-                "x0_slice", mu[t0_idx], dims=mu_dims[1:] if mu_dims is not None else None
+                "x0_slice", mu_frozen[t0_idx], dims=mu_dims[1:] if mu_dims is not None else None
             )
             P0 = pm.Deterministic(
-                "P0_slice", cov[t0_idx], dims=cov_dims[1:] if cov_dims is not None else None
+                "P0_slice", cov_frozen[t0_idx], dims=cov_dims[1:] if cov_dims is not None else None
             )
 
             if scenario is not None:
-                for name in self.data_names:
-                    if name in scenario.keys():
-                        pm.set_data({name: scenario[name]})
-
-                for name in self.data_names:
-                    if name in scenario.keys():
-                        # same here will it always be named data?
-                        pm.set_data(
-                            {"data": np.zeros((len(forecast_index), self.k_endog))},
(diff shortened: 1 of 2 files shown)


pymc_extras/statespace/core/statespace.py
```diff
@@ -2203,28 +2203,39 @@ def forecast(
...
-
             group_idx = FILTER_OUTPUT_TYPES.index(filter_output)
             mu, cov = grouped_outputs[group_idx]
 
+            sub_dict = {
+                data_var: pt.as_tensor_variable(data_var.get_value(), name="data")
+                for data_var in forecast_model.data_vars
+            }
+
+            replacements_diff = np.setdiff1d(
```
jessegrabowski, line 2217:
```suggestion
            missing_data_vars = np.setdiff1d(
```

Slightly more obvious name

jessegrabowski (changes requested):
This is really great!

I guess the last thing I want to ask for is a test that makes sure this works. Basically we want to make sure that if we:

1. Make a really simple model (LevelTrend + Regression)
2. Fit it 
3. Run a forecast with two scenarios (it doesn't matter what the are) 
4. Check that the the mean of all the data *before* the forecast starts is all the same  (you can seed the forecast rng to avoid the issue of random draws)

I think this will test the issue we've been having?

If you have a more clear idea, please go for it. I just want some kind of regression test to make sure this issue is dead once and for all.

Dekermanjian:
Yes, I will update the test `test_forecast_with_exog_data` in `test_statespace.py` to add a check on the exogenous variable. 

When I write the test, is it okay to use lower level components? Otherwise if I just use the `ss_mod.forecast()` method how would I get the data before the forecast starts?

jessegrabowski:
I did the refactor, could you see if it's enough to get a test going?

jessegrabowski:
The test is a good start. I think you're right that we probably only need one scenario, so let's do that just to simplify matters.

You're going to need to test a few extra things:

1. Make sure that there are no non-random generator `SharedVariable` among the inputs to node on the compute graphs of `x0_slice` and `P0_slice` of `test_forecast_model`. You can do this by using `graph_inputs`, which returns a generator of all, well, graph inputs:

```py
from pytensor.graph.basic import graph_inputs
from pytensor.compile import SharedVariable

frozen_shared_inputs = [inpt for inpt in
                        graph_inputs([test_forecast_model.x0_slice, test_forecast_model.P0_slice])
                        if isinstance(inpt, SharedVariable) and not isinstance(inpt.get_value(),
                                                                               np.random.Generator)]


assert len(frozen_shared_inputs) == 0
```

2. Make sure there *are* SharedVariables in the final forecast output, and make sure that it's the data_exog, and make sure the values are correctly set by `pm.set_data`:

```py
unfrozen_shared_inputs = [inpt for inpt in graph_inputs([test_forecast_model.forecast_combined])
                          if isinstance(inpt, SharedVariable) and not isinstance(inpt.get_value(),
                                                                                 np.random.Generator)]
assert len(unfrozen_shared_inputs) == 1
assert unfrozen_shared_inputs[0].name == 'data_exog'

with test_forecast_model:
    dummy_obs_data = np.zeros((len(forecast_index), exog_ss_mod.k_endog))
    pm.set_data(
        {'exog_data': scenario} | {"data": dummy_obs_data},
        coords={"data_time": np.arange(len(forecast_index))},
    )

np.testing.assert_allclose(unfrozen_shared_inputs[0].get_value(), scenario['x1'].values.reshape((-1, 1)))
```

Another thought I had with multiple scenarios was to vary the starting value of the scenario, and make sure that `x0_slice` and `P0_slice` end up being what they're supposed to be. That is, if you call `pm.sample_posterior_predictive(idata_exog, var_names=['x0_slice', 'P0_slice'])` in the test, you can check that the "sampled" values of x0_slice and P0_slice in the posterior predictive are just copies of the values of `smoothed_state` and `smoothed_covariance` in `idata_exog` at `time=t`

Some other notes:

1. To make the tests less miserably slow, use the new pymc sampling mocker. Add this to the top of the file:

```py
from pymc.testing import mock_sample_setup_and_teardown
mock_pymc_sample = pytest.fixture(scope="session")(mock_sample_setup_and_teardown)
```

Then to every fixture or test that calls `pm.sample` add `mock_pymc_sample` to the function signature. For example:

```py
@pytest.fixture(scope="session")
def idata(pymc_mod, rng, mock_pymc_sample):
    with pymc_mod:
        idata = pm.sample(draws=10, tune=0, chains=1, random_seed=rng)
        idata_prior = pm.sample_prior_predictive(draws=10, random_seed=rng)

    idata.extend(idata_prior)
    return idata
```

This will monkeypatch pm.sample with pm.sample_prior_predictive, and rejigger the outputs to look like real sample outputs. This should increase the speed of the tests by a factor of roughly infinity.

2. A super definitive test of our forecast would be to test it against statsmodels. I'm not 100% how their API works for setting exogenous variable, nor do I necessarily think we should do it in this PR. But it's something to keep in mind. Maybe i'll make a separate issue for it.

Dekermanjian:
Hey Jesse, I tried adding `from pymc.testing import mock_sample_setup_and_teardown` but I was getting import errors. I tried syncing up with main but still got the import error so I rolled back the sync because I don't think it's a great idea to merge main into a topic branch.

jessegrabowski (approved):
Looks great! I think we've finally squashed this bug. As one last test, can you confirm that your hurricane notebook works on this PR without the coord hack, and that the forecasts are correct?

PR #511, opened 2025-06-09
Support getattr for prior module

williambdean:
This ports this PR: https://github.com/pymc-labs/pymc-marketing/pull/1534

It was removed from PyMC-Marketing due to the documentation build failing. However, since the 
build for PyMC-extras requires specified classes and functions, it should work.

The prior class now has the convenience imports:

```python
# Direct import of a distribution
from pymc_extras.prior import Normal, HalfNormal

distribution = Normal(mu=Normal(), sigma=HalfNormal(), dims="covariate")

# dot notation for convenience
import pymc_extras.prior as pr

distribution = pr.Normal(mu=pr.Normal(), sigma=pr.HalfNormal(), dims="covariate")
```

Then use like normal:

```python
assert isinstance(distribution, pr.Prior)
assert distribution == Prior("Normal", mu=Prior("Normal"), sigma=Prior("HalfNormal"), dims="covariate")
```

No functional change. Just a convenience import

### DIFF
--- a/pymc_extras/prior.py
+++ b/pymc_extras/prior.py
@@ -84,6 +84,7 @@ from __future__ import annotations
 import copy
 
 from collections.abc import Callable
+from functools import partial
 from inspect import signature
 from typing import Any, Protocol, runtime_checkable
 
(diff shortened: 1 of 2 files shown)


williambdean:
Closes#500

Issue #342, opened 2024-05-15
Newton solver to find the mode of the Gaussian (for INLA)


theorashid:
Part of INLA roadmap https://github.com/pymc-devs/pymc-experimental/issues/340.

We want to do the Laplace approximation around $\mu$. We need to find it first.

## Notes

- jax [implementation](https://github.com/geraschenko/gmrfs/blob/main/gmrfs/inla.py#L108-L112)
- GPy implementation (maybe [somewhere in here](https://github.com/SheffieldML/GPy/blob/devel/GPy/inference/latent_function_inference/laplace.py))

jessegrabowski:
You shouldn't be calling `newton` directly, use the `root_scalar(method='newton')` api. All fixed point iteration problems can be re-cast as root finding problems, I show this in a new pytensor example notebook here. So if you have one variable, you need `root_scalar`, if you need many variables, you need `root`. I still haven't actually seen the math of what you're trying to do, so it's hard for me to give more concrete advice.

Also you don't need to approximate hessians in pytensor, you should just be able to compute it, unless the thing is *really* massive. I've done models with ~100k variables and used symbolic inverse Jacobians successfully.

Michal-Novomestsky:
Thanks for the advice! I'm essentially trying to find the mode of a function (that is, maximise log likelihood). The INLA notes all point to using a fixed-point iterator with Newton's method to find the root of the derivative (approximated up to second order), but given that we need to calculate jac and hess for this at each step anyway, I don't see why we shouldn't simply call ``minimize`` on the NLL directly? If anything, this gives us the flexibility to choose the optimisation scheme we prefer (I believe find_MAP does exactly this, and indeed it uses BFGS as Colt mentions rather than Newton).

jessegrabowski:
As you point out, any minimization problem can be transformed into a root-finding problem by using the system of equations defined by the gradients. If you have access to a contraction operator $T(x_i)$ such that repeated application $x_{i+1} = T(x_i)$ converges to $x^\star$, you can also transform that into a root-finding problem by directly solving for the roots of $T(x_0) - x_0$. And if you have a system of equations $f(x, \theta) =0$, you can always define $\epsilon = f(x_0, \theta)$ as a vector of errors given solution guess $x_0$ and minimize sum of squares $<\epsilon, \epsilon>$. 

Which method to use is entirely up to you.

Michal-Novomestsky:
[...]

PR #513, opened 2025-06-10
Implement a minimizer for INLA

Michal-Novomestsky:
Addresses #342.

This PR should add:
-  [x] ```get_conditional_gaussian_approximation```

To get the mode and the laplace approximation at that point.

Contingent on pymc-devs/pytensor#1182, as it uses ```pytensor.tensor.optimize.minimize``` to find the mode (and hessian at that point).

### DIFF
--- a/pymc_extras/inference/laplace.py
+++ b/pymc_extras/inference/laplace.py
@@ -98,17 +98,20 @@ def get_conditional_gaussian_approximation(
     """
     model = pm.modelcontext(model)
 
+    if args is None:
+        args = model.continuous_value_vars + model.discrete_value_vars
+
     # f = log(p(y | x, params))
-    f = model.logp()
-    jac = pytensor.gradient.grad(f, x)
+    f_x = model.logp()
+    jac = pytensor.gradient.grad(f_x, x)
     hess = pytensor.gradient.jacobian(jac.flatten(), x)
 
-    # Component of log(p(x | y, params)) which depends on x (for rootfinding)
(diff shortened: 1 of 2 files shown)


pymc_extras/inference/laplace.py
```diff
@@ -415,6 +417,128 @@ def sample_laplace_posterior(
...
+def find_mode(
+    x: TensorVariable,
+    args: dict,
+    x0: TensorLike | None = None,
+    model: pm.Model | None = None,
+    method: minimize_method = "BFGS",
+    use_jac: bool = True,
+    use_hess: bool = False,  # TODO Tbh we can probably just remove this arg and pass True to the minimizer all the time, but if this is the case, it will throw a warning when the hessian doesn't need to be computed for a particular optimisation routine.
+    optimizer_kwargs: dict | None = None,
+) -> list[TensorLike]:
```
ricardoV94, line 429: Signature is wrong, this function returns numpy arrays. Why are you compiling a pytensor function and evaluating it? This seems to just be doing `find_map`? I imagine you wanted a symbolic `mode` not the numerical (evaluated) one?

Michal-Novomestsky: I've refactored it to only return the compiled function.

Michal-Novomestsky:
@ricardoV94 @jessegrabowski The unittests currently seem to be failing because the current release of pytensor doesn't have optimize in it yet. Would it be possible to make a point release to so we can merge this?

zaxtax:
@Michal-Novomestsky whenever you are ready for review, remove the Draft status. It looks like there is a small bug in the unit test, but good otherwise!

Issue #514, opened 2025-06-11
Set `hermitian = True` in (Moore-Penrose) pseudo-inverse in kalman smoother - StateSpace


Dekermanjian:
In certain conditions the SVD may fail to converge. The covariance matrix is by definition symmetric setting hermitian to True is more efficient and robust

jessegrabowski:
Closed by #515

PR #515, opened 2025-06-11
Use `hermitian=True` in `pt.linalg.pinv` in Kalman smoother

Dekermanjian:
PR addresses #514 by setting `pt.linalg.pinv(hermitian=True)` in Kaman smoother.


jessegrabowski:
Great catch -- thanks!

PR #517, opened 2025-06-12
removed pinned version of libgcc from environment-test.yml

Dekermanjian:
This resolves #516

PR #519, opened 2025-06-12
Add `BaseEstimator` mixin to `LinearModel`

jessegrabowski:
Scikit-learn 1.7.0 wants all `LinearModel` estimators to also have a `BaseEstimator` mixin to the far right. This is currently raising a DepreciationWarning, but will switch to an error at some point.


jessegrabowski:
No, we don't pin it currently. I think that's why tests were failing in #517 

As far as I can tell, it shouldn't cause any backward compat problems

PR #524, opened 2025-06-16
Restore Python 3.10 support

maresb:
With #522 we no longer need to restrict to Python >=3.11.


ricardoV94:
CI is failing for 3.10

williambdean:
Dont test use dev from pyproject?

maresb:
Ugh, ya, we're pulling in an ancient preliz version. We need to work on the dependency specification to prevent this.

maresb:
Supserseded by #527

Issue #521, opened 2025-06-14
Add `sample_filter_outputs` helper method to `StateSpaceModel`


jessegrabowski:
Copied from @Dekermanjian 's comment [here](https://github.com/pymc-devs/pymc-extras/issues/230#issuecomment-2973024545): 

> Getting the variance covariance matrix for the predicted|filtered|smoothed states using the conditional posterior is not trivial because there is no `idata.cov()` method and `xarray.cov()` only computes the covariances. This means you will need to compute the covariances then the variances and then build the variance covariance matrix manually. It would be nice to have a utility function in state space that would handle this for the user.

Here is a thumbnail sketch of what we're looking for:

```py
with pm.Model(coords=ss_mod.coords) as m:
    ss_mod._build_dummy_graph()
    ss_mod._insert_random_variables()
    
    x0, P0, c, d, T, Z, R, H, Q = ss_matrices = ss_mod.unpack_statespace()
    filter_outputs = ss_mod.kalman_filter.build_graph(pt.as_tensor(ss_mod._fit_data), 
                                                      *ss_matrices)
    smoother_outputs = ss_mod.kalman_smoother.build_graph(T, R, Q, filter_outputs[0], filter_outputs[3])
    for output in filter_outputs[:-1] + list(smoother_outputs):
        pm.Deterministic(output.name, output)
        
    idata_filter = pm.sample_posterior_predictive(idata, 
                                                  var_names=[x.name for x in m.deterministics],
                                                  compile_kwargs={'mode':'JAX'})
```

Obviously it needs a bit more to support exogenous data and dims and stuff. But this is enough to get someone started on a PR :)

PR #526, opened 2025-06-17
added sample_filter_outputs utility and accompanying simple tests

Dekermanjian:
This addresses #521, allows state space users to sample filtered|predicted|smoothed|observed states|covariances using a utility that is consistent with PyMC workflow of sample -> sample_posterior.

### DIFF
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -805,16 +805,16 @@ class PyMCStateSpace:
         states, covs = outputs[:4], outputs[4:]
 
         state_names = [
-            "filtered_state",
-            "predicted_state",
-            "predicted_observed_state",
-            "smoothed_state",
+            "filtered_states",
+            "predicted_states",
+            "predicted_observed_states",
+            "smoothed_states",
         ]
         cov_names = [
-            "filtered_covariance",
-            "predicted_covariance",
-            "predicted_observed_covariance",
-            "smoothed_covariance",
+            "filtered_covariances",
+            "predicted_covariances",
+            "predicted_observed_covariances",
+            "smoothed_covariances",
         ]
 
         with mod:
@@ -939,7 +939,7 @@ class PyMCStateSpace:
             all_kf_outputs = [*states, smooth_states, *covs, smooth_covariances]
             self._register_kalman_filter_outputs_with_pymc_model(all_kf_outputs)
 
-        obs_dims = FILTER_OUTPUT_DIMS["predicted_observed_state"]
+        obs_dims = FILTER_OUTPUT_DIMS["predicted_observed_states"]
         obs_dims = obs_dims if all([dim in pm_mod.coords.keys() for dim in obs_dims]) else None
 
         SequenceMvNormal(
@@ -1727,14 +1727,14 @@ class PyMCStateSpace:
             )
 
             # Filter output names are singular in constants.py but are returned as plural from kalman_.build_graph()
-            filter_output_dims_mapping = {}
-            for k in FILTER_OUTPUT_DIMS.keys():
-                filter_output_dims_mapping[k + "s"] = FILTER_OUTPUT_DIMS[k]
+            # filter_output_dims_mapping = {}
+            # for k in FILTER_OUTPUT_DIMS.keys():
+            #     filter_output_dims_mapping[k + "s"] = FILTER_OUTPUT_DIMS[k]
 
             all_filter_outputs = filter_outputs[:-1] + list(smoother_outputs)
             # This excludes observed states and observed covariances from the filter outputs
             all_filter_outputs = [
-                output for output in all_filter_outputs if output.name in filter_output_dims_mapping
+                output for output in all_filter_outputs if output.name in FILTER_OUTPUT_DIMS
             ]
 
             if filter_output_names is None:
@@ -1752,7 +1752,7 @@ class PyMCStateSpace:
                 ]
 
             for output in filter_output_names:
-                dims = filter_output_dims_mapping[output.name]
+                dims = FILTER_OUTPUT_DIMS[output.name]
                 pm.Deterministic(output.name, output, dims=dims)
 
         frozen_model = freeze_dims_and_data(m)
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -1678,6 +1678,91 @@ class PyMCStateSpace:
 
         return matrix_idata
 
+    def sample_filter_outputs(
+        self, idata, filter_output_names: str | list[str] | None, group: str = "posterior", **kwargs
+    ):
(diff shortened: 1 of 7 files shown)


pymc_extras/statespace/core/statespace.py
```diff
@@ -1678,6 +1678,91 @@ def sample_statespace_matrices(
...
         return matrix_idata
 
+    def sample_filter_outputs(
+        self, idata, filter_output_names: str | list[str] | None, group: str = "posterior", **kwargs
+    ):
+        compile_kwargs = kwargs.pop("compile_kwargs", {})
+        compile_kwargs.setdefault("mode", self.mode)
+
+        with pm.Model(coords=self.coords) as m:
+            pm_mod = modelcontext(None)
```
jessegrabowski, line 1688: This will always return `m` (the current model context). What is `pm_mod` used for?

jessegrabowski (changes requested):
This looks great to me, just a small question about the use of `modecontext` here

jessegrabowski:
Also you need to rebase :)

pymc_extras/statespace/core/statespace.py
```diff
@@ -1678,6 +1678,91 @@ def sample_statespace_matrices(
...
+
+            for output in filter_output_names:
+                match output.name:
+                    case "filtered_states" | "predicted_states" | "smoothed_states":
+                        dims = [TIME_DIM, "state"]
+                    case "filtered_covariances" | "predicted_covariances" | "smoothed_covariances":
+                        dims = [TIME_DIM, "state", "state_aux"]
+                    case "observed_states":
+                        dims = [TIME_DIM, "observed_state"]
+                    case "observed_covariances":
```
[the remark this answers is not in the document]
jessegrabowski: A dictionary mapping cases to dims might be better than the casework here. There isn't already one in constants?

pymc_extras/statespace/core/statespace.py
Dekermanjian: Yeah, this was wrong. I removed it.

pymc_extras/statespace/core/statespace.py
```diff
@@ -1678,6 +1678,91 @@ def sample_statespace_matrices(
 
         return matrix_idata
 
+    def sample_filter_outputs(
+        self, idata, filter_output_names: str | list[str] | None, group: str = "posterior", **kwargs
+    ):
```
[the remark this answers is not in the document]
Dekermanjian: I added this check in.

pymc_extras/statespace/core/statespace.py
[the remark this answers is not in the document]
Dekermanjian: There was a dictionary in `constats.py` variable name `FILTER_OUTPUT_DIMS` but the key names are singular whereas the returned names from both `kalman_(filter|smoother).build_graph()` are plural. I handled this inside the method but I am wondering if you want to consolidate this either in `constants.py` of inside the `kalman_(filter|smoother).build_graph()` methods?

Dekermanjian:
Hey @jessegrabowski, I updated some of the constant in `constants.py` to be plural and updated the tests for any mismatches. There were a few constants that I was not sure about so I wanted to ask you first before I change them. These are:

```
ALL_STATE_DIM = "state(?s)"
ALL_STATE_AUX_DIM = "state(?s)_aux"
OBS_STATE_DIM = "observed_state(?s)"
OBS_STATE_AUX_DIM = "observed_state(?s)_aux"

NEVER_TIME_VARYING = ["initial_state(?s)", "initial_state(?s)_cov(?s)"]
VECTOR_VALUED = ["initial_state(?s)", "state(?s)_intercept(?s)", "obs_intercept(?s)"]

LONG_MATRIX_NAMES = [
    "initial_state(?s)",
    "initial_state(?s)_cov(?s)",
    "state(?s)_intercept(?s)",
    "obs_intercept(?s)",
    "obs_cov(?s)",
    "state_cov(?s)",
]
```

jessegrabowski:
Dims should be singular, I have strong feelings on that.

For the matrix names, I have less of a strong preference. On one hand, `x0` is a vector of states, but on the other hand, it is *the* state vector. So it could go either way. I guess I lean to not changing things in that case?

If you agree, that would mean all of the things you identified there would stay as-is I think.

Dekermanjian:
I agree that the singular dim names sound better. No objections from me about keeping the rest as-is! I think the main issue of having the same thing named differently is now resolved 🤞

jessegrabowski (changes requested):
A few final nitpicks then let's merge this! It's looking really great.

pymc_extras/statespace/core/statespace.py
```diff
@@ -1678,6 +1678,93 @@ def sample_statespace_matrices(
...
+            )
+
+            smoother_outputs = self.kalman_smoother.build_graph(
+                T, R, Q, filter_outputs[0], filter_outputs[3]
+            )
+
+            # Filter output names are singular in constants.py but are returned as plural from kalman_.build_graph()
+            # filter_output_dims_mapping = {}
+            # for k in FILTER_OUTPUT_DIMS.keys():
+            #     filter_output_dims_mapping[k + "s"] = FILTER_OUTPUT_DIMS[k]
```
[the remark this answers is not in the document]
Dekermanjian: oops! Sorry about this. That was a careless oversight. I will clean that up right away!

pymc_extras/statespace/core/statespace.py
```diff
@@ -1684,6 +1684,21 @@ def sample_filter_outputs(
         if isinstance(filter_output_names, str):
             filter_output_names = [filter_output_names]
 
+        drop_keys = {"predicted_observed_states", "predicted_observed_covariances"}
```
jessegrabowski, line 1687: I think we shouldn't treat these as special (even though I agree it's silly to ask for them). I'd be confused if I tried to ask for them and it said it's not a valid filter output name.

Having everything in one place is convenient, even if it's duplicative.

Dekermanjian: @jessegrabowski, I believe this happens in `_postprocess_scan_results()` in `kalman_filter.py`. It looks like the names of the filter outputs are hardcoded in there.

jessegrabowski: If you want to make this consistent in this PR I have no objection. I don't have a good sense if you should change FILTER_OUTPUT_DIMS to match the output names, or change the output names to match the FILTER_OUTPUT_DIMS. I'll defer to you if you have a sense of which one is better.

Dekermanjian:
> I made one last nitpick, but it's not a blocker. Feel free to address or not, then merge :)

Hey @jessegrabowski, I don't believe I have permissions to merge.

jessegrabowski:
Great work as always :D

Dekermanjian:
Thank you, Jesse! Always happy to help!

PR #538, opened 2025-07-16
Pass model to `add_fit_to_inference_data`

OriolAbril:
Without this find_map only works when called inside a model context, otherwise this step has no access to the model object

### DIFF
--- a/pymc_extras/inference/laplace_approx/find_map.py
+++ b/pymc_extras/inference/laplace_approx/find_map.py
@@ -336,7 +336,7 @@ def find_MAP(
     }
 
     idata = map_results_to_inference_data(optimized_point, frozen_model, include_transformed)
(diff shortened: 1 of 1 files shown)


pymc_extras/inference/laplace_approx/find_map.py
```diff
@@ -336,7 +336,7 @@ def find_MAP(
     }
 
     idata = map_results_to_inference_data(optimized_point, frozen_model, include_transformed)
-    idata = add_fit_to_inference_data(idata, raveled_optimized, H_inv)
+    idata = add_fit_to_inference_data(idata, raveled_optimized, H_inv, model)
```
ricardoV94, line 339: To avoid name confusion?

```suggestion
    idata = add_fit_to_inference_data(idata, raveled_optimized, H_inv, model=model)
```

ricardoV94:
Fixed in #543

Issue #541, opened 2025-07-17
pymc-extras can't be loaded with pymc version 5.24.0


Dekermanjian:
Pathfinder is trying to import `CustomProgress` from `pymc.util` however in pymc version 5.24.0 `CustomProgress` was moved from `pymc.util` to `pymc.progress_bar`.

This leads to an import error when trying to load the latest version of pymc-extras with the latest version of pymc.

jessegrabowski:
Closed by #542

PR #542, opened 2025-07-17
fix import error due to moving progress bar functions in latest pymc

Dekermanjian:
import progress bar functions from `pymc.progress_bar` instead of `pymc.util`

This addresses the issue in #541 

What I am not sure of though is what happens if a user wants to use the latest `pymc-extras` version with `pymc<5.24.0`?

### DIFF
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -3,7 +3,7 @@ channels:
 - conda-forge
 - nodefaults
 dependencies:
-  - pymc>=5.24.0
+  - pymc>=5.24.1
   - pytensor>=2.31.4
   - scikit-learn
   - better-optimize>=0.1.5
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -4,8 +4,9 @@ channels:
 - nodefaults
 dependencies:
   - pymc>=5.24.0
+  - pytensor>=2.31.4
   - scikit-learn
-  - better-optimize>=0.1.2
+  - better-optimize>=0.1.5
   - dask<2025.1.1
   - xhistogram
   - statsmodels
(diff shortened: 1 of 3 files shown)


jessegrabowski:
> What I am not sure of though is what happens if a user wants to use the latest pymc-extras version with pymc<5.24.0?

@maresb

ricardoV94:
Preliz needs to change it's import from compile_pymc to compile

ricardoV94:
https://github.com/arviz-devs/preliz/issues/682

jessegrabowski:
preliz already fixed it and cut a release, but it's not out of conda-forge yet

ricardoV94:
I see, then just wait 1/2 days? If you're confident everything is fixed feel free to merge

jessegrabowski:
@ricardoV94 I think something about the progress bar update is breaking pickling on windows.

ricardoV94:
Maybe the custom sampler is broken? You can change the test to run with cores=1 to see the error message

ricardoV94:
Odd can you reproduce locally?

jessegrabowski:
Yeah. It happens on windows because it defaults to `mp_ctx='spawn'`. It will happen on linux too if you set this manually.

ricardoV94:
And didn't happen before our changes? So odd

jessegrabowski:
The error appears to be an infinite recursion when trying to unpickle the inner function in `PointFunc`. That was changed in [this commit](https://github.com/pymc-devs/pymc/commit/3b8500bf3009b44b70371e11304bf7db3e8be805), which might be the actual source of the problems?

jessegrabowski (approved):

jessegrabowski:
Thank you @Dekermanjian !

Dekermanjian:
Happy to help! I am glad this is resolved now!

PR #543, opened 2025-07-19
Forward model argument to  add_fit_to_inference_data in find_MAP

jessegrabowski:
Bugfix for find_MAP. The following code currently errors:

```py
    with pm.Model() as m:
        mu = pm.Normal("mu", 0, 1)
        sigma = pm.Exponential("sigma", 1)
        y_hat = pm.Normal("y_hat", mu=mu, sigma=sigma, observed=np.random.normal(size=10))

    idata = find_MAP(
        model=m,
        method="L-BFGS-B",
        use_grad=True,
        use_hess=False,
        use_hessp=False,
        progressbar=False,
        compile_kwargs={"mode": "JAX"},
    )
```

Because the model argument is not correctly forwarded everywhere it is needed inside find_MAP. This patches the bug and adds a regression test.

### DIFF
--- a/tests/inference/laplace_approx/test_find_map.py
+++ b/tests/inference/laplace_approx/test_find_map.py
@@ -185,6 +185,22 @@ def test_find_MAP(
         assert not hasattr(idata, "unconstrained_posterior")
 
 
+def test_find_map_outside_model_context():
+    """
+    Test that find_MAP can be called outside of a model context.
+    """
+    with pm.Model() as m:
+        mu = pm.Normal("mu", 0, 1)
+        sigma = pm.Exponential("sigma", 1)
+        y_hat = pm.Normal("y_hat", mu=mu, sigma=sigma, observed=np.random.normal(size=10))
+
+    idata = find_MAP(model=m, method="L-BFGS-B", use_grad=True, progressbar=False)
+
+    assert hasattr(idata, "posterior")
+    assert hasattr(idata, "fit")
+    assert hasattr(idata, "optimizer_result")
+
+
 @pytest.mark.parametrize(
     "backend, gradient_backend",
     [("jax", "jax")],
--- a/pymc_extras/inference/laplace_approx/find_map.py
+++ b/pymc_extras/inference/laplace_approx/find_map.py
@@ -335,13 +335,20 @@ def find_MAP(
         var.name: value for var, value in zip(unobserved_vars, unobserved_vars_values)
     }
 
-    idata = map_results_to_inference_data(optimized_point, frozen_model, include_transformed)
-    idata = add_fit_to_inference_data(idata, raveled_optimized, H_inv)
-    idata = add_optimizer_result_to_inference_data(
-        idata, optimizer_result, method, raveled_optimized, model
+    idata = map_results_to_inference_data(
+        map_point=optimized_point, model=frozen_model, include_transformed=include_transformed
     )
+
+    idata = add_fit_to_inference_data(
+        idata=idata, mu=raveled_optimized, H_inv=H_inv, model=frozen_model
+    )
+
+    idata = add_optimizer_result_to_inference_data(
+        idata=idata, result=optimizer_result, method=method, mu=raveled_optimized, model=model
+    )
+
     idata = add_data_to_inference_data(
-        idata, progressbar=False, model=model, compile_kwargs=compile_kwargs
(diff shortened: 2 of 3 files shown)


ricardoV94:
Does this superseed: https://github.com/pymc-devs/pymc-extras/pull/538 ?

ricardoV94 (approved):

jessegrabowski:
> Does this superseed: https://github.com/pymc-devs/pymc-extras/pull/538 ?

Yes. I didn't check if anything else was open before I did this. Sorry @OriolAbril 

> How comes PyPI CI is failing with import?

I think this will be fixed by #541 ?

OriolAbril:
Can we also add the correct pin for pytensor? The map and laplace functionality needs newer pytensor than what is currently allowed. Don't worry about my PR

jessegrabowski:
It's actually not map and laplace needing the new pytensor, it's the INLA function.

I think we could move that specific import inside this function? It shouldn't affect 99% of users.

tomicapretto:
@jessegrabowski this is the dependency problem I encountered before

```shell
conda create --name test-pymc-extras python=3.12
conda activate test-pymc-extras
conda install -c conda-forge pymc-extras
python
```

```python
import pymc_extras
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tomas/miniconda3/envs/test-pymc-extras/lib/python3.12/site-packages/pymc_extras/__init__.py", line 20, in <module>
    from pymc_extras.inference import find_MAP, fit, fit_laplace, fit_pathfinder
  File "/home/tomas/miniconda3/envs/test-pymc-extras/lib/python3.12/site-packages/pymc_extras/inference/__init__.py", line 17, in <module>
  File "/home/tomas/miniconda3/envs/test-pymc-extras/lib/python3.12/site-packages/pymc_extras/inference/laplace_approx/laplace.py", line 38, in <module>
    from pytensor.tensor.optimize import minimize
ModuleNotFoundError: No module named 'pytensor.tensor.optimize'
```

```python
pymc.__version__, pytensor.__version__
# ('5.23.0', '2.31.3')
```

and the pymc_extras I got installed was 0.3.1

If I do

```shell
conda install -c conda-forge "pytensor=2.31.7"
python
```

```python
import pymc_extras
pymc_extras.__version__
# '0.3.1'
```

jessegrabowski:
Thanks Tomi. I updated the pins. I didn't add an upper pin -- I don't really like them, and @maresb said he doesn't either, so that's confirmation bias :D

jessegrabowski:
I moved the version pin and dependency stuff to #542 , they didn't seem appropriate in this PR

tomicapretto:
> I moved the version pin and dependency stuff to #542 , they didn't seem appropriate in this PR

Makes sense!

tests/inference/laplace_approx/test_find_map.py
```diff
@@ -185,6 +185,22 @@ def test_find_MAP(
...
+    with pm.Model() as m:
+        mu = pm.Normal("mu", 0, 1)
+        sigma = pm.Exponential("sigma", 1)
+        y_hat = pm.Normal("y_hat", mu=mu, sigma=sigma, observed=np.random.normal(size=10))
+
+    idata = find_MAP(model=m, method="L-BFGS-B", use_grad=True, progressbar=False)
+
+    assert hasattr(idata, "posterior")
+    assert hasattr(idata, "fit")
+    assert hasattr(idata, "optimizer_result")
```
tomicapretto, line 201: Is it relevant / worth checking the presence of any data group? I see above a call like the following within `find_MAP`

```python
    idata = add_data_to_inference_data(
        idata=idata, progressbar=False, model=model, compile_kwargs=compile_kwargs
    )
```

tomicapretto: haha, well another thing I just saw:

Is it necessary to compute deterministics within `add_data_to_inference_data`? See

https://github.com/pymc-devs/pymc-extras/blob/fd2933f2c787755f03e1cfe106be37e9eb508fb1/pymc_extras/inference/laplace_approx/idata.py#L211-L224

I guess that function is not exposed to the user, but I just wanted to raise that potentially silent side effect.

jessegrabowski: it's not necessary, but that's what pm.sample does right?

tomicapretto: Ahh I see now why it's included

Issue #553, opened 2025-07-31
maintenance of statespace notebooks


Dekermanjian:
We have had a few rounds of making a bit of changes to the statespace module. Adding the multivariate support to structural, updating naming conventions, and changing how we specify different backends to be used. These changes are not reflected in the example notebooks and in certain cases introduce errors in certain cells. These notebooks are possibly the primary means in which new users learn how to use the state space module and would be beneficial to keep them maintained.

Dekermanjian:
Hey @theorashid, I am checking this now. I am on pymc 5.25.1 and pytensor 2.31.7. I am going to try to run through the ETS notebook and then upgrade pytensor and run through again. Give me a few minutes and I will let you know if I am also seeing the same error on my end.

theorashid:
Also, a standard question that I've never understood. We have jax flags everywhere including with the nutpie sampler. Why do I still get this warning if I'm saying not to use the C backend

```sh
src/pymc_ss/.venv/lib/python3.12/site-packages/pytensor/link/c/cmodule.py:2968: UserWarning: PyTensor could not link to a BLAS installation. Operations that might benefit from BLAS will be severely degraded.
This usually happens when PyTensor is installed via pip. We recommend it be installed via conda/mamba/pixi instead.
Alternatively, you can use an experimental backend such as Numba or JAX that perform their own BLAS optimizations, by setting `pytensor.config.mode == 'NUMBA'` or passing `mode='NUMBA'` when compiling a PyTensor function.
For more options and details see https://pytensor.readthedocs.io/en/latest/troubleshooting.html#how-do-i-configure-test-my-blas-library

Dekermanjian:
Hey @theorashid, I was able to run the ETS notebook with no issues using pymc 5.25.1 and pytensor 2.31.7. I am trying to install pytensor 2.32.0 but it just isn't wanting to install. Did you use coda-forge to install it?

AlexAndorra:
Yeah, that's the same error I flagged the other day @theorashid . @ricardoV94 sent me the line I should add a `dprint` to in the pytensor code base, but haven't gotten the time to do it yet. Will be soon, but feel free to beat me to it if you have bandwidth now! Ping me on Discord if you don't find the public thread.

@Dekermanjian install the `main` pytensor branch, and you will be able to reproduce the error

ricardoV94:
> We have jax flags everywhere including with the nutpie sampler. Why do I still get this warning if I'm saying not to use the C backend

[...]

PR #554, opened 2025-07-31
maintenance of statespace notebooks 

Dekermanjian:
I went through the statespace notebooks:
- ETS
- VARMAX
- SARMA
- Custom statespace

and made sure that first there are no breakages that lead to error outputs but also updated specific parts that are going to be deprecated in the future like how/when `mode="JAX"` is being used. 

I also updated the shape of the sigma parameter specified by the SARIMAX implementation where it was returning a shape of `None` instead of `()` for a scalar

This is related to issue #553


AlexAndorra (approved):
Looks great @Dekermanjian , thanks a lot! I just left a comment in the ETS NB. Please make sure the tests pass, and then I'll merge. LMK if you have any questions!

PR #555, opened 2025-07-31
Improve `xarray` display of structural components for multivariate time series

### LINKED ISSUE #545: Better dimensions & coordinates for multivariate time series components
Currently, `sample_conditional_posterior` and `extract_components_from_idata` return square brackets notations for the `state` coord, making it hard to select components when working with multivariate series, as showed in #544.

As discussed with @jessegrabowski and @OriolAbril, we should return an `xarray.Dataset` with one entry per observed state (e.g `['chirac2', 'sarkozy', 'hollande', 'macron', 'macron2']`), thus enabling users to select the corresponding structural components with `.sel`. This also allows each observed state to have arbitrary model structure inside, which the current multivariate setup allows.

AlexAndorra:
Closes #545 

Currently `extract_components_from_idata` returns square brackets notations for the `state` coord, making it hard to select components when working with multivariate series. Things like:
```python
state = [
            "trend[level[gdp]]",
            "trend[trend[gdp]]",
            "trend[level[unemployment]]",
            "trend[trend[unemployment]]",
            "ar[gdp]",
            "ar[unemployment]",
]
```
This PR adds a `restructure` argument to `extract_components_from_idata` (default `False` for backwards compatibility). When `True`, it will restructure the `state` coordinates as a multi-index for easier component selection, thus enabling selections like `idata.sel(component='level')`, `idata.sel(observed='gdp')`, or even `idata.sel(component='level', observed='gdp')`.
Again, this is especially useful for multivariate models with _multiple_ observed states.

More precisely, the `state` dimension is broken down into two new ones, `component` and `observed`, whose coordinates will be `[('level', 'gdp'), ('trend', 'gdp'), ('ar', 'gdp')], [('level', 'unemployment'), ('trend', 'unemployment'), ('ar', 'unemployment')] `.

This also allows each observed state to have arbitrary model structure inside, which the current multivariate setup allows.

_NB: This PR is a first pass, and in no way exhaustive -- we probably need to expand to more complex cases, that users will surface up. But at least it gets the ball rolling and should be self-sufficient to already merge_.

### DIFF
--- a/pymc_extras/statespace/models/structural/core.py
+++ b/pymc_extras/statespace/models/structural/core.py
@@ -18,6 +18,7 @@ from pymc_extras.statespace.models.utilities import (
     join_tensors_by_dim_labels,
     make_default_coords,
 )
+from pymc_extras.statespace.utils.component_parsing import restructure_components_idata
 from pymc_extras.statespace.utils.constants import (
     ALL_STATE_AUX_DIM,
     ALL_STATE_DIM,
@@ -208,7 +209,7 @@ class StructuralTimeSeries(PyMCStateSpace):
         self._component_info = component_info.copy()
 
         self._name_to_variable = name_to_variable.copy()
-        self._name_to_data = name_to_data.copy()
+        self._name_to_data = name_to_data.copy() if name_to_data is not None else {}
 
         self._exog_names = exog_names.copy()
         self._needs_exog_data = len(exog_names) > 0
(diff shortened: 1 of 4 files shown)


jessegrabowski:
I need to think a bit about this. As a v0 I guess it's fine, but I have the feeling that the whole extract_components should be refactored to just do the right thing from jump. Probably it could make better use of arviz in the first place. For example here we're casting everything to numpy then working with that. Seems dumb?

Some general comments:

- I'm not wild about regex when dealing with nested structures, it has a lot of sharp edges, and the patterns are quite arcane.
- I don't think we need to be backwards compatible. We're doing API breaks with every PR these days.

Issue #332, opened 2024-04-16
Implement optimized closed-form gradients for Kalman Filter


jessegrabowski:
I'm quite interested in the results of [this paper](https://arxiv.org/pdf/2303.16846.pdf). The authors  derive closed-form gradients for backprop through Kalman Filters. Specifically equations 28-31.

They report a 38x speedup over autodiff gradients from PyTorch. I suspect (with no evidence) that the gradient computations are where the default PyMC sampler really fall down, so this might even make non-JAX sampling of SS models palatable.

victorgarcia98:
Hey @jessegrabowski, 

Cool paper!
I would like to contribute towards implementing (or at least trying) this gradients. To start, I can try to reproduce the paper and then I'll need some help to know where to integrate these gradients.

What do you think? 😄

jessegrabowski:
Yeah that's how I'd proceed. You have closed form gradients from the current implementation, so basically you can just try to implement their equations 28-31. For a test model use a local level model. I recommend this because it doesn't have any parameters (except for the innovation covariance, and measurement error if included)

A big contribution to just getting the ball rolling would also just be to work out the notation differences between what this paper uses and what we use (we follow Durbin and Koopman Time Series Analysis by Statespace Methods).

victorgarcia98:
Awesome! Sounds like I plan :D

I'll create a table comparing the notations!

JeanVanDyk:
# Hi all

This post is a preliminary draft. I will likely revise and edit it further as the notations and concepts become clearer and better defined.

Here, I will try to synthesize the article and explain what these closed-form expressions are, which supposedly lead to a significant speed-up in the gradient computation time. I will use the notations from *"Durbin and Koopman, Time Series Analysis by State Space Methods"*. (To be fully honest, I mostly relied on the notations from *"Statistical algorithms for models in state space using SsfPack 2.2"*, which I hope are equivalent.)

First, to establish a common ground, let’s define the transition equations used in the Kalman filter:
 
$$
\begin{align}
a_{t+1} &= c_t + T_t a_t + H_t \epsilon_t \qquad \text{where } a_1 \sim \mathcal{N}(a, P) \\
y_t &= d_t + Z_t a_t + G_t \epsilon_t \qquad \text{where } \epsilon_t \sim \mathcal{N}(0, I)
\end{align}
$$

### Parameter definitions

For clarity, here are some key parameters:

#### State equation parameters

- $a_t$ : state vector at time t.
[...]

PR #557, opened 2025-08-02
#332 Closed form gradients for Kalman filter

JeanVanDyk:
Here you'll find, in the notebook section, the notebook I've used to compare execution time between using autodiff and the analytic gradients.

### DIFF
--- a/notebooks/Kalman_Filter_Gradients.ipynb
+++ b/notebooks/Kalman_Filter_Gradients.ipynb
@@ -1,0 +1,743 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "69ae14a1",
+   "metadata": {},
+   "source": [
+    "### Imports"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 127,
+   "id": "90979a41",
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


ricardoV94:
Suggestion, run custom gradient vs default autodiff in pytensor and time it. Print the compiled graph (.dprint() on the function) to see if what pytensor autodiff is doing isn't already similar.

I see the original paper compares with PyTorch, which IMO isn't very clever AD, specially in terms of memory optimization. I'm not sure their 38x speedup / memory improvement also holds against pytensor (or jax).

JeanVanDyk:
Thanks @ricardoV94!

I’ve tried looking into it, but it quickly becomes messy… Do you have a particular method for comparing the dprint output? I saw that you can name operations to make things clearer, but that doesn’t seem very efficient given the hundreds of lines I get.

I also tried timing it, and the gradient with the closed-form expression actually performs worse. I’m inclined to think that autodiff is still being used under the hood, especially since the runtime scales in pretty much the same way for both forms, depending on the number of states.

ricardoV94:
Can you share the timing code, that's easier to give feedback over.

Re the dprint that was just curiosity just paste it after you compile (yeah it will be long)

PR #559, opened 2025-08-03
Fix seasonality dims, coord and params

AlexAndorra:
- TimeSeasonality:
    - Fixed dims and `params_info` for multivariate TS with innovations
    - Renamed parameter from `coefs_` to `params_` to be consistent with the other components
- FrequencySeasonality:
    - Same thing
    - In addition, I fixed the coords of the parameters: the names of the unsaturated states were concatenated with the observed states, which was giving the wrong coords and shape

All of this is now fixed and tested (I also refactored the tests to use pytest fixtures more consistently).
Ready for review 🍾

### DIFF
--- a/pymc_extras/statespace/models/structural/components/seasonality.py
+++ b/pymc_extras/statespace/models/structural/components/seasonality.py
@@ -212,7 +212,7 @@ class TimeSeasonality(Component):
             sigma_level_trend = pm.HalfNormal(
                 "sigma_level_trend", sigma=1e-6, dims=ss_mod.param_dims["sigma_level_trend"]
             )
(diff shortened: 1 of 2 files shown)


jessegrabowski:
I updated this PR. The issue that lead to the test failures was that frequency seasonality has a weird setup for innovations, that requires that *every* hidden state be present in the shocks. 

I also changed the name order for the states to be `fn_i_name` (e.g. Sin_0_annual[data_1]) instead of `fn_name_i` (e.g. Sin_annual_0[data_1]). I thought this better adhered to our something_name decision.

jessegrabowski (approved):
I fixed this up, can be merged after the CI passes

PR #564, opened 2025-08-14
Add deterministic advi

martiningram:
Hi everyone,

I'm one of the authors of the [paper on deterministic ADVI](https://jmlr.org/papers/v25/23-1015.html). There is an open feature request for this in PyMC [here](https://github.com/pymc-devs/pymc/issues/7374) so I thought I'd kick things off with this PR.

In simple terms, DADVI is like ADVI but rather than using a new draw to estimate its objective at each step, it uses a fixed set of draws during the optimisation. That means that (1) it can use regular off-the-shelf optimisers rather than stochastic optimisation, making convergence more reliable, and (2) it's possible to use techniques to improve the variance estimates. This is in the paper, as well as tools to assess how big the error is from using fixed draws.

This PR covers only the first part -- optimising ADVI with fixed draws. This is because I thought I'd start simple and because I'm hoping that it already addresses a real problem with ADVI, which is the difficulty in assessing convergence.

In addition to adding the code, there is an example notebook in `notebooks/deterministic_advi_example.ipynb`. It fits DADVI to the PyMC basic linear regression example. I can add more examples, but I thought I'd start simple.

I mostly lifted the code from my research repository, so there are probably some style differences. Let me know what would be important to change.

Note that JAX is needed, but there shouldn't be any other dependencies.

Very keen to hear what you all think! :)

All the best,
Martin


jessegrabowski:
This is super cool -- I'm very excited to look more closely over the next few days.

Since you're ultimately building a loss function and sending it to scipy.optimize, do you think we could re-use any of the machinery that exists for doing that in the laplace_approx module, for example [this](https://github.com/pymc-devs/pymc-extras/blob/6531ceacdf19201eefe5ff3cbdee21cd8b547b9f/pymc_extras/inference/laplace_approx/scipy_interface.py#L72) or [this](https://github.com/pymc-devs/pymc-extras/blob/6531ceacdf19201eefe5ff3cbdee21cd8b547b9f/pymc_extras/inference/laplace_approx/scipy_interface.py#L144)?

ricardoV94:
Happy to assist you. If you're vectorizing the Jacobian you probably want to build `jacobian(vectorize=True)` which can further be vectorized more nicely.

Everything is described here although a bit scattered: https://pytensor.readthedocs.io/en/latest/tutorial/gradients.html

martiningram:
Hey all, I think I made good progress with the pytensor version. A first version is here: https://github.com/pymc-devs/pymc-extras/pull/564/files#diff-1b6e7da940ec73fce49f5e13ae1db5369ec011cb0b55974ec04d81e519e923f6R55

I think the only major thing missing is to transform the draws back into the constrained space from the unconstrained space. Is there a code snippet anyone could point me to? Thanks for your help and for all the helpful advice you've already given!

zaxtax:
This looks great! Do you have any sense for the performance difference
between this and the original JAX implementation?

zaxtax:
I agree! Just curious from a optimize pytensor POV not as a blocker for
this PR!

pymc_extras/inference/deterministic_advi/dadvi.py
```diff
@@ -0,0 +1,228 @@
...
+
+    Returns
+    -------
+    Tuple[TensorVariable, TensorVariable]
+        A tuple whose first element contains the variational parameters,
+        and whose second contains the DADVI objective.
+    """
+
+    # Make the fixed draws
+    state = np.random.RandomState(random_seed)
```
ricardoV94, line 146: should use `np.random.default_rng`? RandomState is legacy numpy

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
   "better-optimize>=0.1.5",
   "pydantic>=2.0.0",
   "preliz>=0.20.0",
+  "jax>=0.7.0"
```
ricardoV94, line 43: jax is an optional dependency on the project

pymc_extras/inference/deterministic_advi/dadvi.py
```diff
@@ -0,0 +1,228 @@
...
+    sample = means + pt.exp(log_sds) * draw
+
+    # Graph in terms of a single sample
+    logp_draw = pytensor.clone_replace(logp, replace={flat_input: sample})
+    draw_matrix = pt.constant(draws)
+
+    # Vectorise
+    logp_vectorized_draws = pytensor.graph.vectorize_graph(
+        logp_draw, replace={draw: draw_matrix}
+    )
```
ricardoV94, line 173: I don't think you need the intermediate `clone_replace` step?

```suggestion
    draw_matrix = pt.constant(draws)
    samples = means + pt.exp(log_sds) * draw_matrix
    
    logp_vectorized_draws = pytensor.graph.vectorize_graph(
        logp, replace={flat_input: samples}
    )
```

pymc_extras/inference/__init__.py
```diff
@@ -12,9 +12,16 @@
...
 from pymc_extras.inference.laplace_approx.laplace import fit_laplace
 from pymc_extras.inference.pathfinder.pathfinder import fit_pathfinder
 
-__all__ = ["find_MAP", "fit", "fit_laplace", "fit_pathfinder"]
+__all__ = [
+    "find_MAP",
+    "fit",
+    "fit_laplace",
+    "fit_pathfinder",
+    "fit_deterministic_advi",
```
jessegrabowski, line 26:
```suggestion
    "fit_dadvi",
```
For brevity? Just a suggestion

pymc_extras/inference/deterministic_advi/dadvi.py
```diff
@@ -0,0 +1,216 @@
...
+
+    f_fused, f_hessp = _compile_functions_for_scipy_optimize(
+        objective,
+        [var_params],
+        compute_grad=True,
+        compute_hessp=True,
+        compute_hess=False,
+    )
+
+    result = minimize(f_fused, np.zeros(2 * n_params), method="trust-ncg", jac=True, hessp=f_hessp)
```
jessegrabowski, line 93: Allow the user to parameterize the `minimize` function? I know the `find_MAP` signature is a bit... much, but optimization is very problem specific, it's nice to be able to fiddle with it

pymc_extras/inference/deterministic_advi/dadvi.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Graph in terms of a flat input
+    [logp], flat_input = join_nonshared_inputs(
+        point=initial_point_dict, outputs=[logp], inputs=inputs
+    )
+
+    var_params = pt.vector(name="eta", shape=(2 * n_params,))
+
+    means = var_params[:n_params]
+    log_sds = var_params[n_params:]
```
jessegrabowski, line 156:
```suggestion
    means , log_sds= pt.split(var_params, 2)
```

jessegrabowski (approved):
Looks great to me! I left some minor suggestions, but nothing I'd block over. Allower minimizer_kwargs is the biggest issue IMO

martiningram:
Thanks a lot for the suggestions and comments @jessegrabowski ! I think I implemented them, please let me know.

The only one I didn't go for in the end was your `pt.split` suggestion -- it looked great but the suggestion didn't work as-is, and it looked like the signature of `pt.split` is actually quite complicated (see [here](https://github.com/pymc-devs/pytensor/blob/40ccab1aadf84e7b6f3dd2179765aafde86416c2/pytensor/tensor/basic.py#L2189)) and after some messing around I thought the original version might be simpler. But please let me know if I'm missing something.

pymc_extras/inference/dadvi/dadvi.py
```diff
@@ -0,0 +1,237 @@
...
+)
+
+
+def fit_dadvi(
+    model: Model | None = None,
+    n_fixed_draws: int = 30,
+    random_seed: RandomSeed = None,
+    n_draws: int = 1000,
+    keep_untransformed: bool = False,
+    opt_method: minimize_method = "trust-ncg",
```
jessegrabowski, line 31: We should agree on a unified name for this, because `find_MAP` is `method`, `fit_laplace` is `optimizer_method` and now this is `opt_method`.

My first choice is just `method` but it clashes with the `pmx.fit` API (which I don't really like anyway). Thoughts?

pymc_extras/inference/dadvi/dadvi.py
```diff
@@ -0,0 +1,237 @@
...
+        compute_grad=True,
+        compute_hessp=True,
+        compute_hess=False,
+    )
+
+    result = minimize(
+        f_fused,
+        np.zeros(2 * n_params),
+        method=opt_method,
+        jac=True,
```
jessegrabowski, line 113: You don't need to hard-code this to true, `minimize` will automatically figure it out if `f_fused` returns 2 values (and the method uses gradients).

jessegrabowski (approved):
Looks great!

Bummer about `pt.split`, maybe @ricardoV94 can chime in on usage. I agree it looks more complex than expected.

pymc_extras/inference/dadvi/dadvi.py
martiningram: Thanks @jessegrabowski , I agree. I first matched the syntax in `find_map` but then noticed the clash with the `fit` method. I'm also not a huge fan of the shared API tbh.

On the other hand, I personally think `method` is maybe a bit vague anyhow. I'd be OK to go with `optimizer_method`, then at least it's consistent with `fit_laplace`. What do you think?

pymc_extras/inference/dadvi/dadvi.py
martiningram: Thanks, I didn't know that! Will change it

martiningram: I've made using the Jacobian and hessp optional now, let me know what you think!

jessegrabowski:
Yes looks nice. Could you add a use_hess option as well, in case some psycho wants to use `trust-exact` or `dogleg`?

Otherwise I'm satisfied and already approved :)

jessegrabowski:
Oh, I didn't notice you were using `scipy.optimize.minimize`. You can use[ `better_optimize.minimize`](https://github.com/jessegrabowski/better_optimize), which allows a triple-fused function. [Here](https://github.com/pymc-devs/pymc-extras/blob/d5f8f763a55c6eb680bfffa56def636571312ac1/pymc_extras/inference/laplace_approx/find_map.py#L309) is where it's actually being called. Notice that you only pass `f_fused` (which is returning either `loss`, `(loss, grad)`, or `(loss, grad, hess)` ) and `f_hessp` (which can be `None`). 

This is also why `use_jac` didn't work automatically in our previous discussion thread. `better_optimize.minimize` automatically sets the `use_jac`, `use_hess`, and `use_hessp` flags based on the method and the fused function. 

I'm sorry for not looking more closely before, that was a source of confusion.

zaxtax:
@martiningram what's required now to get LR support into DADVI?

martiningram:
Thanks a lot guys!

@jessegrabowski the sparse Jacobians look pretty awesome, thanks! I've also been keen to get sparse matrix algebra going for a long time. But I'll also have to read a bit to understand the topic, and I'm also spread quite thin right now unfortunately. But I'll have a read and will try to understand what you've done so far!

@ricardoV94 Awesome, thanks!

I have a little blog myself (https://martiningram.github.io/), but if you'd like and there's interest, I'd also love to post it on some PyMC blog. Is there one? Only if you like the article of course (I can share it for comments before) and if there's one where it would make sense, let me know.

PR #565, opened 2025-08-20
Numba backend for Pathfinder

fonnesbeck:
First pass at using Numba with Pathfinder. It works, but is not yet performant (order of magnitude slower at the moment).


ricardoV94:
You're implementing way too much stuff. All the contiguous matmul stuff... why? Numba is capable of doing that much better and will dispatch to blas... which you want to do.

What was the old pathfinder doing and why can't you just compile PyTensor stuff to numba?

fonnesbeck:
OK, I figured we wanted to use PyTensor where possible. Back to the drawing board!

ricardoV94:
> OK, I figured we wanted to use PyTensor where possible. Back to the drawing board!

Yes we do, but you shouldn't have to implement any new Ops, like you were doing?

PR #567, opened 2025-08-25
Allow exogenous regressors in `BayesianVARMAX`

jessegrabowski:
This PR adds support for exogenous variables in `BayesianVARMAX` via the observation intercept. Same deal as SARIMAX -- it's implemented in a way that minimizes the number of states in the system.

Unlike Statsmodels, we have support for each observed state to have its own regressors by passing a dictionary. If you just pass a list, everyone shares the same data.

Example 1: Everyone shares data, data names are auto-assigned 

```py
mod = BayesianVARMAX(
        endog_names=["y1", "y2", "y3"],
        order=(1, 0),
        k_exog=2,
    )
```

Example 2: Everyone shares data, data names are provided 

```py
mod = BayesianVARMAX(
        endog_names=["y1", "y2", "y3"],
        order=(1, 0),
        exog_state_names=['x1', 'x2']
    )
```

Example 3: Each observes state (endogenous state) has its own exogenous regressors. y3 has no exogenous regressors at all, so it is omitted from the dictionary. It just so happens that each observed has its own regressors, but they could also share all or a subset.

```py
mod = BayesianVARMAX(
        endog_names=["y1", "y2", "y3"],
        order=(1, 0),
        exog_state_names={'y1':['x1', 'x2'], 'y2':['x3']}
    )
```

### DIFF
--- a/pymc_extras/statespace/models/VARMAX.py
+++ b/pymc_extras/statespace/models/VARMAX.py
@@ -28,60 +28,6 @@ class BayesianVARMAX(PyMCStateSpace):
     r"""
     Vector AutoRegressive Moving Average with eXogenous Regressors
 
-    Parameters
-    ----------
-    order: tuple of (int, int)
-        Number of autoregressive (AR) and moving average (MA) terms to include in the model. All terms up to the
-        specified order are included. For restricted models, set zeros directly on the priors.
-
-    endog_names: list of str, optional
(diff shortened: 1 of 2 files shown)


AlexAndorra (approved):
Quite the monster review! Thanks @jessegrabowski 🤩

Issue #573, opened 2025-08-30
Add `ShiftedBetaGeometric` Distribution


ColtAllen:
A discrete time-to-event distribution originally developed for [contract renewals](https://faculty.wharton.upenn.edu/wp-content/uploads/2012/04/Fader_hardie_jim_07.pdf), with extensions for static covariates and even [expected residual lifetimes and retention elasticities](https://faculty.wharton.upenn.edu/wp-content/uploads/2012/04/Fader_hardie_contractual_mksc_10.pdf).

This distribution would be very useful for CLV cohort studies as outlined in https://github.com/pymc-labs/pymc-marketing/issues/169

ricardoV94:
I don't think this makes sense in pymc-extras. It's too specific and fine in pymc-marketing? Any reason why you wanted to add it here?

PR #574, opened 2025-09-04
Add `ShiftedBetaGeometric` discrete distribution

ColtAllen:
Closes https://github.com/pymc-devs/pymc-extras/issues/573

`test_random_edge_cases` uses rather extreme values and may be unstable.

### DIFF
--- a/pymc_extras/distributions/discrete.py
+++ b/pymc_extras/distributions/discrete.py
@@ -17,9 +17,13 @@ import pymc as pm
 
 from pymc.distributions.dist_math import betaln, check_parameters, factln, logpow
 from pymc.distributions.distribution import Discrete
-from pymc.distributions.shape_utils import rv_size_is_none
+from pymc.distributions.shape_utils import implicit_size_from_params, rv_size_is_none
+from pymc.pytensorf import normalize_rng_param
 from pytensor import tensor as pt
+from pytensor.tensor.random.basic import beta as beta_rng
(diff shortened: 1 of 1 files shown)


pymc_extras/distributions/discrete.py
```diff
@@ -399,3 +400,127 @@ def dist(cls, mu1, mu2, **kwargs):
...
+            size = np.broadcast_shapes(alpha.shape, beta.shape)
+
+        alpha = np.broadcast_to(alpha, size)
+        beta = np.broadcast_to(beta, size)
+
+        p = rng.beta(a=alpha, b=beta, size=size)
+
+        samples = rng.geometric(p, size=size)
+
+        return samples
```
ricardoV94, line 424: Use a SymbolicRandomVariable that defines the random method in terms of pytensor operations. That way it will work automatically in all backends (jax/numba), not just C/python

ricardoV94: Here is an example: https://github.com/pymc-devs/pymc/blob/9f653a6ebcc4a205b1f506e2f3e513753d718b28/pymc/distributions/discrete.py#L393-L412

ColtAllen:
Closing because although this distribution is generalized in application, it has several post-inference derivations that make more sense to bundle into a dedicated model in `pymc-marketing`.

@ricardoV94  thanks for suggesting `SymbolicRandomVariable`; this may prove beneficial elsewhere in  PyMC-Marketing.

PR #582, opened 2025-09-23
Always return dictionary from data_info

jessegrabowski:
VARMAX forecasting is broken when there is no exogenous data because the return value of `data_info` is incorrect. I also added tests that would have caught this bug.


Dekermanjian:
> > I do wonder, though, since other state space models and if I'm not mistaken the structural api use a similar input schema if it would be a good idea to wrap that up in like a _validate_endog_exog_inputs() utility?
> 
> Yes I agree the validation functionality should be moved out to a helper. Can you open an issue? :)

Yes, I'd be happy to!

PR #584, opened 2025-09-28
Fix: autoregressive params info

AlexAndorra:
AutoRegressiveComponent doesn't handle the `params_info` correctly when there are `observed_state_names` and `share_states=True`. This PR fixes it.
Ready for review

### DIFF
--- a/pymc_extras/statespace/models/structural/components/autoregressive.py
+++ b/pymc_extras/statespace/models/structural/components/autoregressive.py
@@ -141,19 +141,19 @@ class AutoregressiveComponent(Component):
 
         self.param_info = {
             f"params_{self.name}": {
-                "shape": (k_states,) if self.k_endog == 1 else (self.k_endog, k_states),
+                "shape": (k_endog_effective, k_states) if k_endog_effective > 1 else (k_states,),
                 "constraints": None,
                 "dims": (AR_PARAM_DIM,)
-                if self.k_endog == 1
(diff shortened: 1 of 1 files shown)


jessegrabowski:
Yes it's a pytensor level issue with the BLAS linker on the new windows VMs that github deployed a few weeks ago. We can ignore it as long as all other tests pass.

jessegrabowski (changes requested):
Can you add a test that fails on main but passes on this PR?

AlexAndorra:
Done @jessegrabowski ! You can merge once tests pass

tests/statespace/models/structural/components/test_autoregressive.py
```diff
@@ -163,6 +163,12 @@ def test_autoregressive_shared_and_not_shared():
         observed_state_names=["data_1", "data_2", "data_3"],
         share_states=True,
     )
+
+    # make sure param_info is correct
+    # shound't have endog state when share_states is True
+    assert "endog_" not in shared.param_info["params_shared_ar"]["dims"][0]
```
jessegrabowski, line 169:
```suggestion
    assert not any(dim.startwith("endog_") for dim in shared.param_info["params_shared_ar"]["dims"])
```

Check everything, not the just (arbitrary) first one

jessegrabowski (approved):
Looks good, I have a small request to make the test a bit less arbitrary, then we're good

PR #588, opened 2025-09-30
Add pathfinder output to idata

### LINKED ISSUE #586: PathfinderResult not available to users
The pathfinder implementation dutifully collects the optimization output in a `PathfinderResult` or `MultiPathfinderResult` object during model fitting but then simply prints the contents to the console and does not return or store these results for users to access. This information should either be available as an attribute, or returned by`fit`, either as a `PathfinderResult` or as part of the `InferenceData` object that is currently returned.

fonnesbeck:
Following a similar approach to laplace, this adds the output from `PathfinderResult` and `MultiPathfinderResult` to the idata that is returned from model fitting.

image

Closes #586

### DIFF
--- a/tests/pathfinder/test_pathfinder.py
+++ b/tests/pathfinder/test_pathfinder.py
@@ -144,6 +144,7 @@ def test_pathfinder(inference_backend, reference_idata):
                 jitter=12.0,
                 random_seed=41,
                 inference_backend=inference_backend,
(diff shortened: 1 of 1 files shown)


fonnesbeck:
Anything is possible. Might be a little unwieldy but I can see what it looks like.

zaxtax (approved):

PR #592, opened 2025-10-08
add missing statespace models to docs

williambdean:
Noticed this one was missing.


williambdean:
Moving the docstring as well so that the details about the model show up in the docs too.

PR #595, opened 2025-10-14
Prefer `mT` over `T` in statespace equations

jessegrabowski:
This hasn't ever come up, but if we end up with more support for batch dims/time varying, we have to be more careful about shape assumptions inside the kalman filters. This PR switches things to batch-friendly operations, switching `x.dot(y)` to `x @ y` (the latter is blockwise) and `x.T` for `x.mT` (in case `x` is not 2d).

There are still a few `.T` sprinkled in there, mostly because there are some 1d objects. It might be preferable to switch everything to 2d (so that e.g. data will be represented as a column vector). This has it's own drawbacks, mostly related to unintentional broadcasting. I tried it in an early iteration and eventually got to where things are now.

### DIFF
--- a/pymc_extras/statespace/filters/utilities.py
+++ b/pymc_extras/statespace/filters/utilities.py
@@ -1,7 +1,5 @@
 import pytensor.tensor as pt
 
-from pytensor.tensor.nlinalg import matrix_dot
-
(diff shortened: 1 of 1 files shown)


AlexAndorra (approved):
I didn't know about that `mT` trick -- that's neat! All good to go @jessegrabowski !
Ruff is failing though; I'll merge once you push again

PR #597, opened 2025-10-17
Pytensor 2.35 Compatibility Fixes

jessegrabowski:
Update imports following pytensor 2.35 to clear noisy warnings.

### DIFF
--- a/tests/test_histogram_approximation.py
+++ b/tests/test_histogram_approximation.py
@@ -78,6 +78,7 @@ def test_histogram_init_discrete(use_dask, min_count, ndims):
 
 @pytest.mark.parametrize("use_dask", [True, False], ids="dask={}".format)
 @pytest.mark.parametrize("ndims", [1, 2], ids="ndims={}".format)
+@pytest.mark.skip(reason="Flakey test on Windows CI, needs investigation")
 def test_histogram_approx_cont(use_dask, ndims):
     data = np.random.randn(*(10000, *(2,) * (ndims - 1)))
     if use_dask:


ricardoV94:
Need to bump pymc dep. Also we should be able to take out that windows warning filter

jessegrabowski:
Preliz also needs to update u_u

tests/model/marginal/test_distributions.py
```diff
@@ -128,5 +128,5 @@ def test_marginalized_hmm_multiple_emissions(batch_chain, batch_emission1, batch
         test_value_emission2 = np.broadcast_to(-test_value, emission2_shape)
     test_point = {"emission_1": test_value_emission1, "emission_2": test_value_emission2}
     res_logp, dummy_logp = logp_fn(test_point)
-    assert res_logp.shape == ((1, 3) if batch_chain else ())
+    assert res_logp.shape == ((3, 1) if batch_chain else ())
```
ricardoV94: The "first dependent RV" that gets the full logp changed with our new toposort algo

ricardoV94:
```python
jax.scipy.linalg.qr(jax.numpy.eye(3), mode="r")  # (JitTracer<float64[3,3]>,)
```

We changed from using `jax.numpy.qr` to `jax.scipy.qr`, so that's why. Will need a fix upstream in PyTensor

jessegrabowski:
Yeah there's a lot of room for improvement in those test files. Thanks for cleaning a bit.

maresb:
Some stuff including JAX is coming from pypi :thinking:

ricardoV94:
CI is still failing?

Issue #598, opened 2025-10-17
Split `populate_component_properties` into separate properties/methods


jessegrabowski:
The `PyMCStateSpace` class asks the user to fill out an (admittedly) huge number of properties so the model can build the coords/dims:

- param_names
- data_names
- param_info
- data_info
- state_names
- observed_states
- shock_names
- coords
- param_dims

I get confused by these and I wrote the thing, so clearly others don't have much hope. So we should think about a better way to organize all this information.

This issue isn't about that, though. When I originally wrote the `Component` class, I thought "all these properties are horrible, let's gather everything into one place". Thus `populate_component_properties` was born. This class is somehow even more confusing to look at than the list of properties. Everything is mixed together in one place, and it's not obvious what is required. 

I think the individual properties is a better approach -- at minimum, it's consistent with the `PyMCStateSpace` class, and it's easier to look at and read.

sajjanaaru-ops:
Hi! I'd like to work on this issue.

I've set up `pymc-extras` locally and started going through the `Component` implementation in `statespace/models/structural/core.py`. I'll first map out the existing `populate_component_properties` / `_set_*` behavior and the relevant tests before making the refactor.

Please let me know if there's a specific API direction you'd prefer for the separate properties/methods.

sajjanaaru-ops:
Hi! I was looking into this issue as a potential first contribution.

After going through the current `Component` implementation in `statespace/models/structural/core.py`, I noticed that `populate_component_properties()` has already been split into separate `_set_states()`, `_set_parameters()`, `_set_shocks()`, `_set_data_info()`, and `_set_coords()` methods, which seems to match the direction proposed earlier in this thread.

Is there any remaining work intended for this issue, or can it be considered resolved?

PR #607, opened 2025-11-02
Represent statespace metadata with dataclasses 

Dekermanjian:
This is a draft proposal for #598 

The idea is to handle each component separately using `_set_{component}` methods and all information are stored using data classes for easy mapping. 

I believe this will simplify our tests of these components and will reduce redundancies where we have the same information spread across multiple sub-components like `data_names` and `data_info`. 

@jessegrabowski let me know what you think I put a little notebook together to showcase the changes.


pymc_extras/statespace/models/structural/components/regression_dataclass.py
```diff
@@ -0,0 +1,539 @@
...
+
+    def __str__(self):
+        base = f"name: {self.name}\nshape: {self.shape}\ndims: {self.dims}"
+        if self.constraints:
+            return base + f"\nconstraints: {self.constraints}"
+        return base
+
+
+@dataclass
+class ParameterProperties:
```
jessegrabowski:
```suggestion
class ParameterInfo(Info[StatespaceParameter]):
```

Same here. I'm also not against just calling it `StatespaceParameters` (plural), but that's a bit less obvious.

pymc_extras/statespace/models/structural/components/regression_dataclass.py
```diff
@@ -0,0 +1,539 @@
...
+
+@dataclass
+class DataProperties:
+    data: list[DataProperty]
+    needs_exogenous_data: bool = field(default=False, init=False)
+
+    def __post_init__(self):
+        for d in self.data:
+            if d.is_exogenous:
+                self.needs_exogenous_data = True
```
jessegrabowski:
```suggestion
    @property
    def needs_exogenous_data(self) -> bool:
        return any(d.is_exogenous for d in self.items)
```

Just compute this on demand with a property instead of going through post init hoops. In this case you can also set `frozen=True`.

pymc_extras/statespace/models/structural/components/regression_dataclass.py
```diff
@@ -0,0 +1,539 @@
...
+        regression_state_prop = CoordProperty(
+            dimension=f"state_{self.name}", labels=[state for state in self.base_names]
+        )
+        endogenous_state_prop = CoordProperty(
+            dimension=f"endog_{self.name}", labels=[state for state in self.observed_state_names]
+        )
+
+        self.coords = CoordProperties(coords=[regression_state_prop, endogenous_state_prop])
+
+    def populate_component_properties(self) -> None:
```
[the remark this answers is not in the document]
Dekermanjian: @jessegrabowski, I was able to update the regression component to work with the new stuff. I spent some time trying to write a `build_parameters` method that would work across all the STS components but they are quite different from one another. Are you sure you want to put the `_set_foo()` in the `Component` superclass? I think maybe we can strip out some of the most basic stuff like the `k_endog` and `k_exog` stuff but I think we will still need to extend the `_set_foo()` methods in the subclasses.

pymc_extras/statespace/models/structural/components/regression.py
```diff
@@ -194,64 +206,110 @@ def make_symbolic_graph(self) -> None:
...
         k_endog_effective = 1 if self.share_states else k_endog
+        k_states = self.k_states // k_endog_effective
+
+        beta_param_name = f"beta_{self.name}"
+        beta_param_shape = (k_endog_effective, k_states) if k_endog_effective > 1 else (k_states,)
+        beta_param_dims = (
+            (f"endog_{self.name}", f"state_{self.name}")
+            if k_endog_effective > 1
+            else (f"state_{self.name}",)
+        )
```
jessegrabowski: Let's eliminate the intermediate variables and just put these directly into the `Parameter` constructor

pymc_extras/statespace/models/structural/core.py
```diff
@@ -136,6 +141,7 @@ class StructuralTimeSeries(PyMCStateSpace):
            methods (2nd ed.). Oxford University Press.
     """
 
+    # TODO need to discuss cutting some of these args down. All the <foo>_names args are already inside of <foo>_info
```
jessegrabowski: Yep let's get rid of them. The `info` should be the single source of truth for all information about the model.

pymc_extras/statespace/models/structural/core.py
```diff
@@ -540,6 +565,8 @@ def __init__(
 
         self.param_info = {}
         self.data_info = {}
+        self.shock_info = {}
+        self.state_info = {}
```
jessegrabowski: Lets initialize these with empty versions of the relevant dataclasses, or `None` if you prefer. They should never be a dictionary.

pymc_extras/statespace/models/structural/core.py
```diff
@@ -764,15 +806,17 @@ def _combine_property(self, other, name, allow_duplicates=True):
...
             raise TypeError(
                 f"Property {name} of {self} and {other} are not the same and cannot be combined. Found "
                 f"{type(self_prop)} for {self} and {type(other_prop)} for {other}'"
             )
 
-        if not isinstance(self_prop, list | dict):
+        if not is_dataclass(self_prop) and not isinstance(self_prop, list | tuple | dict):
             raise TypeError(
-                f"All component properties are expected to be lists or dicts, but found {type(self_prop)}"
+                f"All component properties are expected to be dataclasses, but found {type(self_prop)}"
```
jessegrabowski: Would be better to have this error say the specific class it expects (ParameterInfo / DataInfo / etc) rather than just generic "dataclass"

pymc_extras/statespace/models/structural/core.py
```diff
@@ -648,6 +675,21 @@ def make_and_register_data(self, name, shape, dtype=floatX) -> Variable:
...
+        raise NotImplementedError
+
+    def _set_shocks(self) -> None:
+        raise NotImplementedError
+
+    def _set_states(self) -> None:
+        raise NotImplementedError
+
+    def _set_coords(self) -> None:
+        raise NotImplementedError
```
jessegrabowski: Not all of these should be mandatory. We should offer some defaults based on `k_endog`/`k_exog`/`k_posdef` if the user doesn't provide pretty names. Coords are always optional in pymc models. Most models don't use `data` at all

pymc_extras/statespace/models/utilities.py
```diff
@@ -19,14 +20,23 @@
...
 def make_default_coords(ss_mod):
-    coords = {
-        ALL_STATE_DIM: ss_mod.state_names,
-        ALL_STATE_AUX_DIM: ss_mod.state_names,
-        OBS_STATE_DIM: ss_mod.observed_states,
-        OBS_STATE_AUX_DIM: ss_mod.observed_states,
-        SHOCK_DIM: ss_mod.shock_names,
-        SHOCK_AUX_DIM: ss_mod.shock_names,
-    }
+    ALL_STATE_COORD = Coord(dimension=ALL_STATE_DIM, labels=ss_mod.state_names)
```
jessegrabowski: This function should be eliminated in favor of the method `CoordsInfo.defaults_from_model`

tests/statespace/models/structural/components/test_regression.py
```diff
@@ -66,7 +66,7 @@ def test_exogenous_component(self, rng, regression_data, innovations):
 
         mod = mod.build(verbose=False)
         _assert_basic_coords_correct(mod)
-        assert mod.coords["state_exog"] == ["feature_1", "feature_2"]
+        assert mod.coords["state_exog"].labels == ["feature_1", "feature_2"]
```
jessegrabowski: revert these changes after `Model.coords` already returns `self._coords_info.to_dict()`

jessegrabowski (changes requested):

pymc_extras/statespace/models/structural/core.py
Dekermanjian: Here are you suggesting I exclude _set_data and _set_coords? Then based on the k_<foo> make some default data classes for states and shocks? Leaving only parameter as required?

pymc_extras/statespace/core/properties.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+@dataclass(frozen=True)
+class ParameterInfo(Info[Parameter]):
+    def __init__(self, parameters: list[Parameter]):
+        super().__init__(items=tuple(parameters), key_field="name")
+
+    def add(self, parameter: Parameter) -> "ParameterInfo":
+        # return a new ParameterInfo with parameter appended
+        return ParameterInfo(parameters=[*list(self.items), parameter])
```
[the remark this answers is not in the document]
jessegrabowski: I might have missed your point, but all the classes should hold their information in the `items` field. Is that not already true?

tests/statespace/core/test_properties.py
```diff
@@ -0,0 +1,119 @@
+import pytest
```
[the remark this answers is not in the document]
jessegrabowski: You'll need the same tests either way :D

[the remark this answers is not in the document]
Dekermanjian: You are right! I am adding those now. I was able to add in the `add` & `merge` methods to the base class.

pymc_extras/statespace/models/structural/components/regression.py
```diff
@@ -194,64 +206,108 @@ def make_symbolic_graph(self) -> None:
...
+                shape=(k_states,),
+                dims=(f"state_{self.name}",),
+                constraints="Positive",
+            )
+
+            self.param_info = ParameterInfo(parameters=[beta_parameter, sigma_parameter])
+            self.param_names = self.param_info.names
+        else:
+            self.param_info = ParameterInfo(parameters=[beta_parameter])
+            self.param_names = self.param_info.names
```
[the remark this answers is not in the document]
Dekermanjian: My bad on the indentation. Will fix in next commit.

pymc_extras/statespace/models/structural/core.py
```diff
@@ -525,23 +537,43 @@ def __init__(
         self.k_posdef = k_posdef
         self.measurement_error = measurement_error
 
-        self.state_names = list(state_names) if state_names is not None else []
-        self.observed_state_names = (
-            list(observed_state_names) if observed_state_names is not None else []
+        self.param_info = ParameterInfo(
```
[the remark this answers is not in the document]
Dekermanjian: I think that would be nice. Right now, there is this intermediate conversion step that needs to take place.

pymc_extras/statespace/models/structural/core.py
```diff
@@ -648,6 +680,24 @@ def make_and_register_data(self, name, shape, dtype=floatX) -> Variable:
         self._name_to_data[name] = placeholder
         return placeholder
 
+    def _set_parameters(self) -> None:
+        raise NotImplementedError
+
+    def _set_shocks(self) -> None:
+        return ShockInfo(shocks=[Shock(name=f"shock_{n}") for n in range(self.k_posdef or 0)])
+
+    def _set_states(self) -> None:
```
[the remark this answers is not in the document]
Dekermanjian: Yes, correct. I will fix this error.

pymc_extras/statespace/core/properties.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @property
+    def observed_state_names(self) -> tuple[State, ...]:
+        return tuple(s.name for s in self.items if s.observed)
+
+    @property
+    def unobserved_state_names(self) -> tuple[State, ...]:
+        return tuple(s.name for s in self.items if not s.observed)
+
+    def merge(self, other: StateInfo, allow_duplicates: bool = False) -> StateInfo:
```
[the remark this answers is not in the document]
Dekermanjian: Yeah, I think we can port over the unique merge when `allow_duplicates` is False and duplicate merge otherwise to all the other components. Can you think of any case where this might break downstream SSM processes?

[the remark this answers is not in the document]
jessegrabowski: Only one way to find out :)

Serious answer: I don't have enough conviction in my mental model of how things currently work to say off hand

Dekermanjian:
> Yeah it looks really great! Go ahead and do the others. Excited to get this over the finish line

Yeah, this is going to be pretty cool! I am going to commit + push the components one by one as I complete them. I will also let you know here once I get all of them done so that if you prefer reviewing everything all at once.

jessegrabowski:
Amazing! How is your impression of it after refactoring everything? Does it seem a bit more readable? Any sharp edges that are still confusing?

Also it looks like you need to rebase and run pre-commit

Dekermanjian:
I think it is more readable. There are some repeated code that I am trying to figure out how to make less repeatable. These are the k\_\<foo\> variables that you will see at the start of many of the \_set\_\<foo\> methods. I am sure once you take a look in your review we can iron them out.

jessegrabowski:
> @jessegrabowski, I rebased and ran pre-commit again but the pre-commit says everything passed.

Looks like the CI is happy now :)

pymc_extras/statespace/core/properties.py
```diff
@@ -0,0 +1,257 @@
...
+    ALL_STATE_AUX_DIM,
+    ALL_STATE_DIM,
+    OBS_STATE_AUX_DIM,
+    OBS_STATE_DIM,
+    SHOCK_AUX_DIM,
+    SHOCK_DIM,
+)
+
+if TYPE_CHECKING:
+    from pytensor.tensor.variable import TensorVariable
```
jessegrabowski: You should be fine to import `TensorVariable` outside of the TYPE_CHECKING block. That is just for avoiding circular imports.

pymc_extras/statespace/core/properties.py
```diff
@@ -0,0 +1,241 @@
...
+                continue
+            key = getattr(item, self.key_field)
+            # if key in index:
+            #     raise ValueError(f"Duplicate {self.key_field} '{key}' detected.") # This needs to be possible for shared states
+            index[key] = item
+        if missing_attr:
+            raise AttributeError(f"Items missing attribute '{self.key_field}': {missing_attr}")
+        object.__setattr__(self, "_index", index)
+
+    def _key(self, item: T) -> str:
```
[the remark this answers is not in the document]
jessegrabowski: Let's add it later when/if its needed, then

pymc_extras/statespace/core/properties.py
```diff
@@ -0,0 +1,257 @@
...
+    def __str__(self) -> str:
+        base = "coordinates:"
+        for coord in self.items:
+            coord_str = str(coord)
+            indented = "\n".join("  " + line for line in coord_str.splitlines())
+            base += "\n" + indented + "\n"
+        return base
+
+    @classmethod
+    def default_coords_from_model(cls, model: PyMCStateSpace | Component) -> Self:
```
jessegrabowski:
```suggestion
    def default_coords_from_model(cls, model: StateSpaceLike) -> Self:
```

To simplify all the type checking, we can just put a protocol here that has the property we expect.

pymc_extras/statespace/core/properties.py
```diff
@@ -0,0 +1,257 @@
+from __future__ import annotations
+
+import warnings
+
+from collections.abc import Iterator
+from copy import deepcopy
+from dataclasses import dataclass, fields
+from typing import TYPE_CHECKING, Generic, Self, TypeVar
```
jessegrabowski:
```suggestion
from typing import TYPE_CHECKING, Generic, Self, TypeVar, Protocol
```

pymc_extras/statespace/core/statespace.py
```diff
@@ -236,8 +242,8 @@ def __init__(
         self._fit_exog_data: dict[str, dict] = {}
 
         self._needs_exog_data = None
-        self._name_to_variable = {}
-        self._name_to_data = {}
+        # self._name_to_variable = {}
+        # self._name_to_data = {}
```
jessegrabowski: Initialize an empty `_tensor_variable_info` here

pymc_extras/statespace/core/statespace.py
```diff
@@ -559,14 +567,17 @@ def make_and_register_data(
                 f"parameters."
             )
 
-        if name in self._name_to_data.keys():
+        if name in self._tensor_data_info:
             raise ValueError(
                 f"{name} is already a registered placeholder variable with shape "
-                f"{self._name_to_data[name].type.shape}"
+                f"{self._tensor_data_info[name].type.shape}"
```
jessegrabowski: Same as above, use `add` so the duplicate check is being done by the `Info`. The `StateSpace` class itself doesn't need to know about that logic.

pymc_extras/statespace/models/structural/components/cycle.py
```diff
@@ -250,76 +263,116 @@ def make_symbolic_graph(self) -> None:
...
+            state_names = [
                 f"{name}[{var_name}]" if k_endog_effective > 1 else name
-                for var_name in self.observed_state_names
+                for var_name in observed_state_names
                 for name in base_names
             ]
+            self.state_info = StateInfo(
+                states=[State(name=name, observed=False, shared=False) for name in state_names]
+            )
+            self.state_info = self.state_info.merge(
```
jessegrabowski: see above (or not!)

pymc_extras/statespace/models/structural/components/cycle.py
```diff
@@ -250,76 +263,116 @@ def make_symbolic_graph(self) -> None:
...
+        coord_container = [state_coords]
+
+        if k_endog_effective != 1:
+            endog_coords = Coord(dimension=f"endog_{self.name}", labels=observed_state_names)
+            coord_container.append(endog_coords)
+
+        self.coords_info = CoordInfo(coords=coord_container)
+
+    def populate_component_properties(self) -> None:
+        # The order of these are important because things like k_states are set in _set_states and needed in _set_shocks etc..
```
jessegrabowski: Actually this comment isn't true for this component -- is it true for any?

pymc_extras/statespace/models/structural/components/cycle.py
```diff
@@ -250,76 +263,116 @@ def make_symbolic_graph(self) -> None:
...
+        self.param_info = ParameterInfo(parameters=params_container)
+
+    def _set_shocks(self) -> None:
+        if self.innovations:
+            self.shock_info = ShockInfo(shocks=[Shock(name=name) for name in self.state_names])
+
+    def _set_coords(self) -> None:
+        k_endog = self.k_endog
+        k_endog_effective = 1 if self.share_states else k_endog
+        base_names = [f"{f}_{self.name}" for f in ["Cos", "Sin"]]
```
jessegrabowski: This is the same as `self.state_names` right?

pymc_extras/statespace/models/structural/components/measurement_error.py
```diff
@@ -122,25 +128,36 @@ def __init__(
...
-                "constraints": "Positive",
-                "dims": (f"endog_{self.name}",) if k_endog_effective > 1 else None,
-            }
-        }
+            endog_coord = Coord(dimension=f"endog_{self.name}", labels=observed_state_names)
+            coords_container.append(endog_coord)
+
+        self.coords_info = CoordInfo(coords=coords_container)
+
+    def populate_component_properties(self) -> None:
```
jessegrabowski: You should implement all the other mandatory setters too (the ones that raise `NotImplementedError` by default). Just have them do nothing or set an empty `Info`. Not sure which is easier downstream. 

If we always set an empty Info, the logic for joining components becomes a bit easier, because we never have to worry about the `None` case. If its a lot of work no need to worry about it in this PR.

pymc_extras/statespace/models/structural/components/measurement_error.py
```diff
@@ -122,25 +128,36 @@ def __init__(
...
+        # The order of these are important because things like k_states are set in _set_states and needed in _set_shocks etc..
```
jessegrabowski: definitely not true here!

pymc_extras/statespace/models/structural/components/regression.py
```diff
@@ -172,64 +183,103 @@ def make_symbolic_graph(self) -> None:
...
-        self.param_dims = {
-            f"beta_{self.name}": (f"endog_{self.name}", f"state_{self.name}")
-            if k_endog_effective > 1
-            else (f"state_{self.name}",)
-        }
+        self.shock_info = ShockInfo(shocks=[Shock(name=name) for name in shock_names])
 
-        base_names = self.state_names
+    def _set_states(self) -> None:
+        self.base_names = self.state_names
```
jessegrabowski: do this assignment in `__init__` if you want to store the base_names. You can do it after you call `super()`.

As a general rule, you shouldn't create a member variable outside of the initializer.

pymc_extras/statespace/models/structural/components/regression.py
```diff
@@ -307,7 +301,7 @@ def _set_coords(self) -> None:
 
     def populate_component_properties(self) -> None:
```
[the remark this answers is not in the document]
jessegrabowski: The generic `populate_component_propterties` should call everything. For any setter that raises by default, we should always implement it (even if we just have `def _set_foo(self): pass`). Then it's really obvious that this component doesn't have a specific property. Alternatively we can set an empty `Info`.

For optional properties, we should make sure it's harmless to call e.g. `_set_data_info()` for components that don't use it.

pymc_extras/statespace/models/structural/core.py
```diff
@@ -519,54 +542,116 @@ def __init__(
...
+        self._k_posdef = k_posdef
         self.measurement_error = measurement_error
 
-        self.state_names = list(state_names) if state_names is not None else []
-        self.observed_state_names = (
-            list(observed_state_names) if observed_state_names is not None else []
+        self.param_info = ParameterInfo(
+            parameters=[
+                Parameter(name=n, shape=(1,), dims=(f"{n}_placeholder"))
+                for n in (param_names or [])
```
jessegrabowski:
```suggestion
                for name in param_names
```

Have the class properties always return a list, so you don't have to check for `None` here.

pymc_extras/statespace/models/structural/core.py
```diff
@@ -519,54 +542,116 @@ def __init__(
...
-        self.param_dims = {}
-
-        self.param_info = {}
-        self.data_info = {}
-
-        self.param_counts = {}
+        self.data_info = DataInfo(
+            data=[
+                Data(name=n, shape=(None, 1), dims=(f"{n}_placeholder"), is_exogenous=False)
+                for n in (data_names or [])
```
jessegrabowski: see above

pymc_extras/statespace/models/structural/core.py
```diff
@@ -519,54 +542,116 @@ def __init__(
...
+            ]
+            + [
+                Data(name=n, shape=(None, 1), dims=(f"{n}_placeholder"), is_exogenous=True)
+                for n in (exog_names or [])
```
jessegrabowski: see above

pymc_extras/statespace/models/structural/core.py
```diff
@@ -519,54 +542,116 @@ def __init__(
...
+            ]
+        )
+        self.shock_info = ShockInfo(shocks=[Shock(name=n) for n in (shock_names or [])])
+        self.state_info = StateInfo(
+            states=[State(name=n, observed=False, shared=share_states) for n in (state_names or [])]
```
jessegrabowski: see above

pymc_extras/statespace/models/structural/core.py
```diff
@@ -632,22 +719,53 @@ def make_and_register_data(self, name, shape, dtype=floatX) -> Variable:
...
 
+    def _set_parameters(self) -> None:
+        """
+        Sets component parameter specifications. Since different component types will require different specifications this method must be over-ridden.
+        """
+        raise NotImplementedError
+
+    def _set_shocks(self) -> None:
+        """
+        Sets default shock specifications based on the number of sources of innovations in the component. It is encouraged to override this method.
```
jessegrabowski:
```suggestion
        Set default shock specifications based on the number of sources of innovations in the component. 
        
        It is encouraged to override this method.
```

pymc_extras/statespace/models/structural/core.py
```diff
@@ -632,22 +719,53 @@ def make_and_register_data(self, name, shape, dtype=floatX) -> Variable:
...
+        """
+        self.shock_info = ShockInfo(
+            shocks=[Shock(name=f"shock_{n}") for n in range(self.k_posdef or 0)]
+        )
+
+    def _set_states(self) -> None:
+        """
+        Sets default state specifications based on number of states and endogenous variables in the component. It is encouraged to override this method.
```
jessegrabowski:
```suggestion
        Set default state specifications based on number of states and endogenous variables in the component. 
        
        It is encouraged to override this method.
```

tests/statespace/models/structural/components/test_cycle.py
```diff
@@ -178,10 +183,14 @@ def test_cycle_multivariate_with_innovations_and_cycle_length(rng):
     x, y = simulate_from_numpy_model(cycle, rng, params)
 
     cycle.build(verbose=False)
-    _assert_basic_coords_correct(cycle)
+    # _assert_basic_coords_correct(cycle) # This assumes coords has keys such as state and state_aux. I think these were passing in the past because even if these don't exist we were supplying empty dims with those keys
```
jessegrabowski: We can remove this test then

tests/statespace/models/structural/components/test_level_trend.py
```diff
@@ -209,6 +217,7 @@ def test_add_level_trend_with_different_observed():
     )
 
 
+@pytest.mark.filterwarnings("ignore:Duplicate names found:UserWarning")
```
jessegrabowski: duplicates shouldn't ever warn, it should raise (if disallowed) or silently accept it.

tests/statespace/models/structural/components/test_measurement_error.py
```diff
@@ -18,27 +20,34 @@ def test_measurement_error(rng):
...
+        "data_1",
+        "data_2",
+    )
+    assert mod.param_info["sigma_obs"].dims == ("endog_obs",)
 
 
 def test_measurement_error_share_states():
     mod = st.MeasurementError("obs", observed_state_names=["data_1", "data_2"], share_states=True)
-    mod.build(verbose=False)
+    mod = mod.build(verbose=False)  # How was this passing?
```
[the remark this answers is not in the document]
Dekermanjian: Alright, cool! Just making sure this wasn't a silent bug we were missing.

pymc_extras/statespace/models/structural/components/cycle.py
Dekermanjian: This was true for setter methods that used inside the setter method one of the component class properties that defined by the <foo>_info objects. By not using these properties in the setter methods we can call the setter methods in any order.

pymc_extras/statespace/models/structural/components/level_trend.py
```diff
@@ -287,3 +233,124 @@ def make_symbolic_graph(self) -> None:
...
+        if self.k_posdef > 0:
+            shock_coord = Coord(
+                dimension=f"shock_{self.name}", labels=[state for state in base_shock_names]
+            )
+            coords_container.append(shock_coord)
+
+        self.coords_info = CoordInfo(coords=coords_container)
+
+    def populate_component_properties(self) -> None:
+        # The order of these are important because things like k_states are set in _set_states and needed in _set_shocks etc..
```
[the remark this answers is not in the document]
Dekermanjian: Yes, I agree with you. These should  work in any order. This can be done by avoiding using the component class properties inside the setter methods.

pymc_extras/statespace/models/structural/components/measurement_error.py
Dekermanjian: Yes, after I make some changes it won't be true anywhere.

pymc_extras/statespace/models/structural/components/seasonality.py
```diff
@@ -424,6 +375,100 @@ def make_symbolic_graph(self) -> None:
...
+        coords_container = [state_coord]
+
+        if k_endog_effective > 1:
+            endog_coord = Coord(dimension="endog_{self.name}", labels=observed_state_names)
+            coords_container.append(endog_coord)
+
+        self.coords_info = CoordInfo(coords=coords_container)
+
+    def populate_component_properties(self) -> None:
+        # The order of these are important because things like k_states are set in _set_states and needed in _set_shocks etc..
```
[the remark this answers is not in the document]
Dekermanjian: This won't be true after some fixes. I will remove the comment everywhere

jessegrabowski:
> @jessegrabowski, I had to rename the data classes TensorVariable and TensorVariableInfo to PyTensorVariable and PyTensorVariableInfo because there was a conflict with Pytensor's pt.TensorVariable class.

What about `SymbolicTensor` and `SymbolicTensorInfo` ?

pymc_extras/statespace/core/statespace.py
```diff
@@ -559,14 +567,17 @@ def make_and_register_data(
...
             )
 
         placeholder = pt.tensor(name, shape=shape, dtype=dtype)
         self._name_to_data[name] = placeholder
```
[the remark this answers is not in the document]
Dekermanjian: Oh! I missed that and misunderstood what you were referring to. Thank you, Jesse for pointing that out. I will fix it in the next commit.

Issue #602, opened 2025-10-24
Feature: allow multiple PymcStateSpace instances via name parameter


opherdonchin:
### Summary

Add a `name: str` parameter to `PymcStateSpace` so multiple independent state-space models can coexist in a single PyMC model without name collisions.

### Motivation

* Many real problems combine several latent processes (e.g., multiple subjects, multiple modalities, etc.).
* Currently, internal RV/graph names collide; a `name` kwarg scopes internal variables and allows clean composition.

### Proposed API

```python
ssm = PymcStateSpace(
    ...,
    name="vision",   # optional, defaults to existing behavior if omitted
)
```

* When `name` is provided, internal RVs/coordinates are prefixed/namespaced accordingly.
* When omitted, behavior is 100% backward compatible.

### Docs / Example

* Docstring update for `PymcStateSpace` parameters.

### Performance / Risks
* Negligible performance impact (string prefixing only).
* No breaking changes expected.

### Reference implementation

* Fork/branch: https://github.com/opherdonchin/pymc-extras
* Happy to open a PR if maintainers approve the direction.

### Questions for maintainers

* Any preferred naming convention for namespacing (e.g., `{name}::var` vs `{name}_var`)?

PR #611, opened 2025-11-14
Add optional name parameter to PyMCStateSpace for multiple models

opherdonchin:
### Summary

As discussed in [[#602](https://github.com/pymc-devs/pymc-extras/issues/602)](https://github.com/pymc-devs/pymc-extras/issues/602), this PR adds an optional `name` parameter to `PyMCStateSpace`. When provided, it is used to prefix all variables and registered data created by that state-space model. This makes it possible to use multiple state-space models inside one PyMC model without name collisions.

### What this changes

* Add `name: str | None = None` to `PyMCStateSpace.__init__`.
* Add `_prefix_name(self, var_name: str) -> str`, which:

  * returns `var_name` unchanged if `self.name is None`
  * otherwise returns `"{name}_{var_name}"`.
* Apply `_prefix_name` when:

  * creating `pm.Deterministic` variables,
  * registering Kalman filter and likelihood outputs,
  * constructing names in `SequenceMvNormal`,
  * naming `H_jittered`, `x0_slice`, `P0_slice`, `forecast_*`, `irf`, `initial_shock`,
  * building names for posterior predictive variables.
* Update data helpers (`add_data_to_active_model`, `register_data_with_pymc`) so that state-space models can optionally pass prefixed data names.

### Backwards compatibility

* If `name` is not provided, everything behaves exactly as before.
* Existing code that uses a single `PyMCStateSpace` should work without any changes.

### Example

```python
ssm1 = PyMCStateSpace(..., name="vision")
ssm2 = PyMCStateSpace(..., name="motor")
```

This results in variables such as `"vision_obs"`, `"motor_obs"`, `"vision_data"`, etc., allowing both models to coexist cleanly in the same PyMC model.

### Testing

* Added `tests/statespace/test_register_variable.py` to check the basic naming and variable-registration behavior.
* Attempted to run `pytest tests/statespace` in a clean `pip install -e ".[dev]"` environment.
* Test collection currently stops with `FutureWarning` errors from `pytensor.graph.basic.*` imports (these have moved to `pytensor.graph.traversal.*`). This appears unrelated to this PR and originates from existing imports in the codebase and test suite.

### DIFF
--- a/pymc_extras/prior.py
+++ b/pymc_extras/prior.py
@@ -187,10 +187,7 @@ def handle_dims(x: pt.TensorLike, dims: Dims, desired_dims: Dims) -> pt.TensorVa
     missing_dims = aligned_dims.sum(axis=0) == 0
     new_idx = aligned_dims.argmax(axis=0)
 
-    args = [
-        "x" if missing else idx
-        for (idx, missing) in zip(new_idx, missing_dims, strict=False)
-    ]
+    args = ["x" if missing else idx for (idx, missing) in zip(new_idx, missing_dims, strict=False)]
     args = _remove_leading_xs(args)
     return x.dimshuffle(*args)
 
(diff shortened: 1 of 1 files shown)


jessegrabowski:
Thanks for the poke, this one slipped past me. Will review asap.

Looks like pre-commit is failing though, start by clearing that.

pymc_extras/statespace/core/statespace.py
```diff
@@ -484,6 +509,53 @@ def add_default_priors(self) -> None:
         """
         raise NotImplementedError("The add_default_priors property has not been implemented!")
 
+    def register_variable(self, name: str, variable: pt.TensorVariable) -> None:
```
jessegrabowski: Can you give an example of a workflow where you want to use this? I'm hesitant because it breaks the pattern assumed by statespace, in which the statespace model and the pymc model are totally independent.

jessegrabowski:
We merged #607 and this feature was requested by another user [here](https://discourse.pymc.io/t/is-it-possible-to-simultaneously-calibrate-multiple-statespace-models/17570/2), so I definitely think getting this in is a priority.

PR #615, opened 2025-12-14
Run whole suite with C and Numba linkers

ricardoV94:
Relies on next PyTensor/PyMC release

Testing compatibility in anticipation of default switch in PyTensor

### DIFF
--- a/pymc_extras/statespace/filters/kalman_filter.py
+++ b/pymc_extras/statespace/filters/kalman_filter.py
@@ -648,7 +648,9 @@ class SquareRootFilter(BaseFilter):
         y_hat = Z.dot(a) + d
         v = y - y_hat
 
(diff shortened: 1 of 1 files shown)


jessegrabowski (approved):
lgtm!

jessegrabowski:
I already merged, but we're probably testing too much now. We should do cvm and numba on ubuntu and just one or the other on windows? Seems like overkill to do the whole matrix.

ricardoV94:
Oh yeah absolutely. Also I would like to test on higher python versions (still keep on of the jobs on the oldest ofc), 3.14? It's also faster

jessegrabowski:
whoops.

I guess that undoes one of the times I accidentally did a rebase merge

jessegrabowski:
We should have a tag for no-squash-merge to make that easier to communicate

PR #619, opened 2025-12-16
recover_marginals infers model from model context

### LINKED ISSUE #610: `recover_marginals` doesn't work with implicit model from modelcontext
```python
with marginal_model:
  recover_marginals(idata)  # missing argument idata (because it assumes idata is model)
```

zaxtax:
This enables using `recover_marginals` in a model context. Just required making the model argument optional

Closes #610

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,10 +105,13 @@ filterwarnings =[
     'ignore:os\.fork\(\) was called\.:RuntimeWarning',
 
     # Preliz needs to update for pytensor > 2.35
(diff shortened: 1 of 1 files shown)


tests/model/marginal/test_marginal_model.py
```diff
@@ -933,7 +936,7 @@ def test_nested(self):
             )
             idata = InferenceData(posterior=dict_to_dataset(prior))
 
-        idata = recover_marginals(marginal_m, idata, return_samples=True)
+            idata = recover_marginals(idata, return_samples=True)
```
ricardoV94: Are we testing the explicit model kwarg anywhere now?

ricardoV94 (commented):
Just some minor suggestions, let me know if too much of a drag

ricardoV94:
We have two RTD jobs? Anyway failure unlikely to be related

PR #627, opened 2026-01-09
Allow non-vectorized laplace fit draws

ricardoV94:


pymc_extras/inference/laplace_approx/laplace.py
```diff
@@ -200,134 +147,178 @@ def get_conditional_gaussian_approximation(
...
-    unconstrained_names = [name for name in names if name not in constrained_names]
-
-    new_outputs, unconstrained_vector = join_nonshared_inputs(
-        model.initial_point(),
-        inputs=model.value_vars,
-        outputs=outputs,
-    )
+def unpack_last_axis(packed_input, packed_shapes):
+    if len(packed_shapes) == 1:
+        # Single case currently fails in unpack
```
jessegrabowski: We need an issue for that

pymc_extras/inference/laplace_approx/laplace.py
```diff
@@ -200,134 +147,178 @@ def get_conditional_gaussian_approximation(
...
-        (rv, name) for rv, name in zip(value_rvs, names) for name in unconstrained_names
-    ]
-    # names = [*constrained_names, *unconstrained_names]
+def draws_from_laplace_approx(
+    *,
+    mean,
+    covariance=None,
+    standard_deviation=None,
+    chains: int,
+    draws: int,
```
jessegrabowski: Since there's already a ton of arguments, we could just remove `chains` and always set it to `1`? 

For the record I did it this way because sometimes you need two idata to have the same shape (if you're combining results from two models), so it's handy. But a user can always do `draws = chains * draws` then reshape it.

pymc_extras/inference/laplace_approx/laplace.py
```diff
@@ -230,104 +178,125 @@ def _unconstrained_vector_to_constrained_rvs(model):
...
-                dims = (*batch_dims, *model.named_vars_to_dims[name])
-            else:
-                n_dim = batched_rv.ndim - 2  # (temp_chain, temp_draw) are always first 2 dims
-                dims = (*batch_dims,) + (None,) * n_dim
-
-            pm.Deterministic(name, batched_rv, dims=dims)
-
-    return laplace_model
 
+def draws_from_laplace_approx(
```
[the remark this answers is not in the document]
ricardoV94: I don't mean the current function is more general, but that code we have in it can be refactored out into perhaps 2 more general functions, leaving only the part where we define the latent vector to be a gaussian unique to the "laplace_approx".

pymc_extras/inference/laplace_approx/laplace.py
ricardoV94: Very happy to remove chains!

PR #630, opened 2026-01-25
Remove "Component" suffix from StateSpace components

### LINKED ISSUE #534: Drop "Component` from component names
Currently the components in `statespace.structural` have inconsistent names. On one hand, there's a component called `LevelTrendComponent`, but then there's also `MeasurementError` and `TimeSeaonality`.

I think the word "component" here is just extra keystrokes, so we can rename everything to drop it. So `LevelTrendComponent` becomes `LevelTrend`, `RegressionComponent` becomes `Regression`, etc.

jessegrabowski:
Rename:

- `LevelTrendComponent` -> `LevelTrend`
- `CycleComponent` -> `Cycle`
- `RegressionComponent` -> `Regression`

Closes #534

### DIFF
--- a/pymc_extras/statespace/models/structural/__init__.py
+++ b/pymc_extras/statespace/models/structural/__init__.py
@@ -1,3 +1,5 @@
+import warnings
+
 from pymc_extras.statespace.models.structural.components.autoregressive import (
     Autoregressive,
 )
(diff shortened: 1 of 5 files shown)


jessegrabowski:
@ricardoV94 ping

PR #635, opened 2026-02-02
New ADVI API 

jessegrabowski:
This PR moves the work from https://github.com/pymc-devs/pymc/pull/7799 over here to extras. The key idea is to copy the numpyro guide model API, but with our own PyMC flair. 

I also added:

- stick the landing estimator
- forward sampling helper
- LLM sketch for a training API patterned after pytorch-lightning

I updated the notebook to use the proposed training API. It obviously needs a lot of work (being that it's llm trash) but hopefully it can get some ideas flowing.


zaxtax (commented):
I think this is pretty neat! After some rebasing and small API changes we should definitely try to get this merged!

pymc_extras/inference/advi/fit.py
```diff
@@ -0,0 +1,186 @@
...
+        recent = np.mean(history[-window:])
+        previous = np.mean(history[-2 * window : -window])
+        return bool(abs(recent - previous) < self.relative_tolerance * (abs(previous) + 1e-8))
+
+
+def fit_advi(
+    model: Model | None = None,
+    *,
+    n_steps: int = 10_000,
+    draws_per_step: int = 8,
```
ricardoV94:
```suggestion
    draws_per_step: int = 1,
```

fonnesbeck:
Is the plan to deprecate the OPVI `fit` function when this merged? (or matriculates to the main project?)

pymc_extras/inference/advi/autoguide.py
```diff
@@ -180,7 +180,7 @@ def AutoDiagonalNormal(model: Model, random_seed=None) -> AutoGuideModel:
             )
             Deterministic(
                 rv.name,
-                loc + pt.softplus(scale) * z,
+                loc + pt.exp(scale) * z,
```
zaxtax, line 192: We had it with softplus since that's what numpyro does? Why change?

ricardoV94: 2x faster, logdet becomes -scale, instead of log(softplus)

ricardoV94: It's not a sure choice, just something I'm exploring

ricardoV94:
> Is the plan to deprecate the OPVI fit function when this merged? (or matriculates to the main project?)

There's no big plan on my end, still very much WIP. Maybe if it proves itself. The VI part of this is def smaller in scope (but also complexity)

jessegrabowski:
> Is the plan to deprecate the OPVI fit function when this merged? (or matriculates to the main project?)

From my part, this refactor was an effort to explore a different design. Could maybe lead to a full replacement of pm.fit, but the scope so far is just ADVI. I'm not sure the particle methods would be in scope for what we have in this PR.

But I think the guide model setup is really interesting, and works really well with the pymc Model API + pytensor graph mutations. So its appealing in and of itself.

ricardoV94:
@zaxtax let's leave the minibatch out of this. I already had quite some discussion with @jessegrabowski I don't think PyMC models have the right API for us to eagerly go and guess what's batched data and what's not, and assume it's over the 0-th axis and so forth. Check the over-enginereered #700 I started for what I think is the minimum work a library built on top of PyMC would have to do to "auto minibatch". I would not put any minibatch stuff on this PR. For now if users want to minibatch they'd use the lover-or-hate it pm.Minibatch path

zaxtax:
But the proposed API here offers a high degree of control, so I'm not sure
what's ambiguous here

ricardoV94:
> But the proposed API here offers a high degree of control, so I'm not sure what's ambiguous here

That's exactly the sort of discussion I didn't want to have here... Anyway I see no place to control which X variables you stream and how. Your proposal is a bit more low-level and less debatable than an "auto minibatch", but I still would prefer to leave it to its own PR. The scaling thing is also a bit naive (in terms of calling model.logp() per var which can be costly). I didn't check how it handles univariate vs multivariates. Also why merging observed transform into the fitting routine?

zaxtax:
I'm happy to have that be part of a separate PR. I'll try to on that future
PR really go into detail about the approach.

jessegrabowski:
@ricardoV94 can I close this PR as superseded by the other ones?

zaxtax:
Did my Trainer class commit end up in another PR?

ricardoV94:
> @ricardoV94 can I close this PR as superseded by the other ones?

no, but this needs rebasing of things that went elsewhere

ricardoV94:
> Did my Trainer class commit end up in another PR?

I reverted the minibatch commit, the idea was to discuss in another PR yes. I didn't open yet

zaxtax:
There was another Trainer commit in addition to that one around minibatch

ricardoV94:
I tried to be careful to only revert, not delete stuff that wasn't mine. But may have failed

zaxtax:
I mean this one https://github.com/pymc-devs/pymc-extras/pull/635/changes/c18622923c24e43bcfc425a160e8f89297b515db

zaxtax:
But the Trainer class I don't think is a major design change in my opinion
at least

ricardoV94:
I'll take a look, but I suspect we may have 3 chefs who want 3 slight different things here. 

Personally, I'd distinguish fit machinery state and the fit result. For the fit machinery state we should try to reuse the pymc step stampler state api/machinery (not the step samplers, those are different things). https://github.com/pymc-devs/pymc/blob/df1a369d5fbdc0cc49082b0c0ef7cbd511cf76e9/pymc/step_methods/state.py#L54

We can reject if we find a good reason why it doesn't fit the mold. It doesn't need to happen in this PR either, but I'd like to plan on checking it.

Or data structures / classes, I don't want to incuur overhead of storing each step result in a fancy dataclass or promise a state that was stale (i.e. read the shared variables content on the first iteration and pretend those are up to date in the pure pytensor path). I don't know if I pushed the latest state of the api notebook, but most my decisions were based on bringing inference speed down, and they checked out there (in the last example brought from @juanitorduz blogpost). When the inner step function gets cheap I don't want to be losing to jax/numpyro.

My one criteria is I don't want @juanitorduz to have an excuse to use numpyro for SVI. Or at least not a good one.

On the control over the fit loop, and what we expose vs not I don't have much opinion, but both you @jessegrabowski and @zaxtax need to check if you're on the same page. I didn't look into that at all but may have accidentally reverted or erased something (sorry if erased, it wasn't supposed to if it happened)

zaxtax:
> I'll take a look, but I suspect we may have 3 chefs who want 3 slight different things here.
> 
> Personally, I'd distinguish fit machinery state and the fit result. For the fit machinery state we should try to reuse the pymc step stampler state api/machinery (not the step samplers, those are different things). https://github.com/pymc-devs/pymc/blob/df1a369d5fbdc0cc49082b0c0ef7cbd511cf76e9/pymc/step_methods/state.py#L54
> 
> We can reject if we find a good reason why it doesn't fit the mold. It doesn't need to happen in this PR either, but I'd like to plan on checking it.
> 
> Or data structures / classes, I don't want to incuur overhead of storing each step result in a fancy dataclass or promise a state that was stale (i.e. read the shared variables content on the first iteration and pretend those are up to date in the pure pytensor path). I don't know if I pushed the latest state of the api notebook, but most my decisions were based on bringing inference speed down, and they checked out there (in the last example brought from @juanitorduz blogpost). When the inner step function gets cheap I don't want to be losing to jax/numpyro.
> 
> My one criteria is I don't want @juanitorduz to have an excuse to use numpyro for SVI. Or at least not a good one.
> 
> On the control over the fit loop, and what we expose vs not I don't have much opinion, but both you @jessegrabowski and @zaxtax need to check if you're on the same page. I didn't look into that at all but may have accidentally reverted or erased something (sorry if erased, it wasn't supposed to if it happened)

I don't have a strong opinion here but I worry sampler state is a poor fit for the different state representations ADVI methods can have. As for the fit loop, all I ask is @jessegrabowski have a peek about what I made in the existing commit (possibly the reverted one) and we confirm if we are on the same page. If we are not, I'm happy for us to just pick something and go with it

juanitorduz:
I will be happy to test this against a real-life example https://github.com/pymc-labs/pymc-forecast/blob/main/docs/examples/retail_stockouts.ipynb :D.

In the meantime,I wanna try to run this (as in, the current api, not this one) on GPU + Modal (I made the numpyro version fit this model in less than 10 min with SVI and 50K time series), https://github.com/pymc-labs/pymc-forecast/blob/main/docs/examples/retail_stockouts.ipynb is the same example, and we tried to port the batching machinery https://github.com/pymc-labs/pymc-forecast/pull/48

zaxtax:
And yes I want @juanitorduz to love whatever we make

juanitorduz:
> And yes I want @juanitorduz to love whatever we make

I AM loving it ... that is why we are pushing with @twiecki a testing field : https://github.com/pymc-labs/pymc-forecast :D

ricardoV94:
> I will be happy to test this against a real-life example https://github.com/pymc-labs/pymc-forecast/blob/main/docs/examples/retail_stockouts.ipynb :D.
> 
> In the meantime,I wanna try to run this (as in, the current api, not this one) on GPU + Modal (I made the numpyro version fit this model in less than 10 min with SVI and 50K time series), https://github.com/pymc-labs/pymc-forecast/blob/main/docs/examples/retail_stockouts.ipynb is the same example, and we tried to port the batching machinery https://github.com/pymc-labs/pymc-forecast/pull/48

what batching machinery? minibatch or vectorization -adjacent?

juanitorduz:
> > I will be happy to test this against a real-life example https://github.com/pymc-labs/pymc-forecast/blob/main/docs/examples/retail_stockouts.ipynb :D.
> > In the meantime,I wanna try to run this (as in, the current api, not this one) on GPU + Modal (I made the numpyro version fit this model in less than 10 min with SVI and 50K time series), https://github.com/pymc-labs/pymc-forecast/blob/main/docs/examples/retail_stockouts.ipynb is the same example, and we tried to port the batching machinery [pymc-labs/pymc-forecast#48](https://github.com/pymc-labs/pymc-forecast/pull/48)
> 
> what batching machinery? minibatch or vectorization -adjacent?

For posterior predictive sampling. In the NumPyro function `Predictive()` this was killing the memory, and the "trick" that worked was to move this computation to the CPU and do it by batches.

zaxtax:
Oh this is a good insight. Can PyMC PPC be run in batches over the data?

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,512 @@
...
+        backend: str | None = None,
+        compile_kwargs: dict | None = None,
+        random_seed=None,
+    ):
+        # Per-run policy: read afresh by every fit call
+        self.learning_rate = learning_rate
+        self.convergence_window = convergence_window
+        self.relative_tolerance = relative_tolerance
+
+        self.model = model
```
zaxtax, line 186: Should we make Trainer object `modelcontext` aware?

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,512 @@
...
+        self._loss_history: list[float] = []
+        self._step = 0
+
+    @property
+    def guide(self) -> AutoGuideModel | None:
+        """The guide being fit, once resolved. Compiled in, so it cannot be replaced."""
+        return self._guide
+
+    @property
+    def clip_norm(self) -> float | None:
```
zaxtax: Do we need this?

tests/inference/advi/test_schedules.py
zaxtax, line 1: Do we need learning rate schedulers in this PR?

tests/inference/advi/test_fit.py
```diff
@@ -0,0 +1,195 @@
...
+    kwargs = dict(model=model, n_steps=50, draws=50, convergence_window=None, backend="jax")
+    draws_a = fit_advi(random_seed=42, **kwargs)["posterior"].dataset["theta"].values
+    draws_b = fit_advi(random_seed=42, **kwargs)["posterior"].dataset["theta"].values
+    draws_c = fit_advi(random_seed=13, **kwargs)["posterior"].dataset["theta"].values
+
+    np.testing.assert_array_equal(draws_a, draws_b)
+    assert not np.array_equal(draws_a, draws_c)
+
+
+def test_fit_advi_early_stopping(conjugate_model):
```
zaxtax: Do we need early stopping support in this PR?

zaxtax (approved):

zaxtax:
@ricardoV94 what's left to do on this one other than maybe remove learning rate and clip norm from the Trainer object?

PR #637, opened 2026-02-03
Rerun statespace notebooks

jessegrabowski:
We've had a lot of changes lately, so it's a good time to re-run everything.

All sampling is now done with numba -- no more jax. Otherwise things are largely unchanged.


jessegrabowski:
@Dekermanjian this is updated to include the recent changes, would appreciate a look

Dekermanjian:
@jessegrabowski, went through the notebooks and they look great. I ran into some rendering issues that I assume are local to reviewNB, and some oddities that I spotted in the SARIMAX notebook but could be irrelevant.

ricardoV94 (approved):

PR #641, opened 2026-02-08
Do not include jacobian adjustment in find_MAP

littlewu2508:
jacobian=False is the behavior of pymc.find_MAP

### DIFF
--- a/pymc_extras/inference/laplace_approx/find_map.py
+++ b/pymc_extras/inference/laplace_approx/find_map.py
@@ -165,7 +165,6 @@ def find_MAP(
     use_hess: bool | None = None,
     initvals: dict | None = None,
     random_seed: int | np.random.Generator | None = None,
-    jacobian: bool = True,
     jitter_rvs: list[TensorVariable] | None = None,
     progressbar: bool = True,
     include_transformed: bool = True,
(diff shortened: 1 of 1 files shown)


jessegrabowski:
No because you're still transforming the variables. Maybe I misunderstand something.

littlewu2508:
> Yes but you get the result you would by optimizing the constrained model. The change of variables becomes simply a trick for scipy to optimize without finding boundary issues?

Yes, that's why I'd like to use find_MAP with `jacobian=False`, because it finds the true maximum point in sample space defined by users, not the transformed space, because I use it to do frequentist analysis.

The only thing I'm not certain is whether the MAP of transformed space is useful or not. Can it serve as a better starting points for NUTS? If it has some value, then maybe we should keep a flag.

pymc_extras/inference/laplace_approx/find_map.py
```diff
@@ -165,6 +165,7 @@ def find_MAP(
     use_hess: bool | None = None,
     initvals: dict | None = None,
     random_seed: int | np.random.Generator | None = None,
+    jacobian: bool = True,
```
jessegrabowski:
```suggestion

```

jessegrabowski:
Thanks for spotting and fixing this @littlewu2508

PR #642, opened 2026-02-08
REF: Migrate to ArviZ 1.0 (InferenceData -> DataTree)

AAlexxis222:
Migrates pymc-extras to support ArviZ 1.0, replacing InferenceData with xarray.DataTree.

Changes:
* Imports: Replaced arviz.InferenceData with xarray.DataTree.
* Attributes: Refactored .posterior access to dictionary style (["posterior"]).
* API: Updated constructors (DataTree.from_dict) and replaced .extend() with .update().

Verification:
* ~325 tests passing (verified against PyMC PR #8019 rebase).
* Validated prior and pathfinder modules.

### DIFF
--- a/pymc_extras/model/marginal/marginal_model.py
+++ b/pymc_extras/model/marginal/marginal_model.py
@@ -532,7 +532,7 @@ def recover_marginals(
     dims.update(rv_dims)
     rv_dataset = dict_to_dataset(
         rv_dict,
-        library=pymc,
+        inference_library=pymc,
         dims=dims,
         coords=coords,
         skip_event_dims=True,


AAlexxis222:
Hi! Yes, I applied the formatting tools (`black`, `isort`, `ruff`) because the `pre-commit` CI job was failing with linting errors on the modified files.

I wanted to ensure the CI was green. However, if you prefer to keep the diff minimal, I can try to revert the style changes and only keep the logic migration (though `pre-commit` might fail again)

jessegrabowski:
> I wanted to ensure the CI was green. However, if you prefer to keep the diff minimal, I can try to revert the style changes and only keep the logic migration (though pre-commit might fail again)

The CI needs to be green, but this PR shouldn't touch 87 files. You should be running the pre-commit hooks we have defined locally that should be sufficient.

jessegrabowski:
Thanks for this @AAlexxis222 , I pulled your changes from this PR into #668 and will continue work from there. We need to coordinate these changes with the next major pymc and pytensor releases because they will break user code.

Issue #612, opened 2025-11-17
Shape error in UnivariateFilter for time-varying matrices


sebcroft:
Hi!

I seem to be running into an error when using the univariate filter with time-varying state-space matrices. However, the same model set-up works fine when using the standard filter with time-varying matrices, and it also works for the univariate filter if all the matrices are time-invariant.

Below is a minimal-ish example that should hopefully reproduce it:
- pymc : 5.26.1
- pytensor : 2.35.1
- pymc_extras : 0.5.0
```python
import pytensor.tensor as pt
import pymc as pm
from pymc_extras.statespace.core import PyMCStateSpace

class StateSpace(PyMCStateSpace):
    def __init__(self, k_endog, k_states, k_posdef, n, filter_type):
        self.n = n
        super().__init__(k_endog=k_endog, k_states=k_states, k_posdef=k_posdef, filter_type=filter_type)

    @property
    def param_names(self):
        return ['x0', 'P0', 'sigma_epses', 'sigma_etas']


    def make_symbolic_graph(self):
        x0 = self.make_and_register_variable('x0', shape=(self.k_states,))
        P0 = self.make_and_register_variable('P0', shape=(self.k_states,self.k_states))
        sigma_epses = self.make_and_register_variable('sigma_epses', shape=(self.k_endog,))
        sigma_etas = self.make_and_register_variable('sigma_etas', shape=(self.k_posdef,))        

        self.ssm['initial_state', :] = x0
        self.ssm['initial_state_cov', :, :] = P0
        self.ssm['obs_cov', *np.diag_indices(self.k_endog)] = sigma_epses**2
        self.ssm['state_cov', *np.diag_indices(self.k_posdef)] = sigma_etas**2

            
        self.ssm['transition', :, :] = pt.zeros((self.k_states, self.k_states))
        self.ssm['selection'] = pt.zeros((self.n, self.k_states, self.k_posdef)) # think this is enough for PyMCStateSpace to assume it's time varying
        self.ssm['design', :, :] = pt.zeros((self.k_endog, self.k_states)) 


def model_builder(ss):

    y = np.zeros((ss.n, ss.k_endog))
    
    with pm.Model() as model:
        x0 = pm.Normal('x0', shape=(ss.k_states,))
        P0_diag = pm.Exponential('P0_diag', 1, shape=(ss.k_states,))
        P0 = pm.Deterministic('P0', pt.diag(P0_diag))
[...]

PR #653, opened 2026-02-19
Fix ordering of time-varying statespace matices for UnivariateFilter's kalman step

sebcroft:
This aims to address #612 

Following the error in #612, the design matrix Z should have shape (2,3) not (3,3). Intial thought is the latter shape corresponds to the transition matrix T.

In BaseFilter, the kalman_step function calls the method unpack_args to reorganise the inputs into a standardized order. This is needed because when the user specifies time varying matrices, the kalman_step arguments (y, a, P, c, d, T, Z, R, H, Q) can get 'mixed-up' depending on which variables are assigned to the sequences and non_sequences in pytensor.scan.

UnivariateFilter redefines the kalman_step function and does not currently call unpack_args. Therefore, when the selection matrix R is the only time varying matrix (as in #612) it gets sent to the sequences/start (with y) and shifts the remaining args up by 1. As a result, T is incorrectly used for the Z input, which aligns with the shape error. 

When all matrices are time-invariant unpack_args is not required because the correct ordering is preserved -- this is why UnivariateFilter works for the time-invariant matrices as mentioned in #612.

To address this I have added the method unpack_args to the kalman_step in UnivariateFilter.

**Some extra stuff:**

The above changes should hopefully be enough, but to keep things consistent with the BaseFilter you could possibly deal with the nan values using handle_missing_values instead i.e.:

```
def kalman_step(self, *args):
    
    y, a, P, c, d, T, Z, R, H, Q = self.unpack_args(args)
    
    nan_mask = pt.isnan(y)
    
    y_masked, Z_masked, H_masked, all_nan_flag = self.handle_missing_values(y, Z, H)
    
    result = pytensor.scan(
        self._univariate_inner_filter_step,
        sequences=[y_masked, Z_masked, d, pt.diag(H_masked), nan_mask],
        outputs_info=[a, P, None, None, None],
        name="univariate_inner_scan",
        return_updates=False,
    )
```
I'm happy to add this if you think it's best!

### DIFF
--- a/pymc_extras/statespace/filters/kalman_filter.py
+++ b/pymc_extras/statespace/filters/kalman_filter.py
@@ -778,7 +778,10 @@ class UnivariateFilter(BaseFilter):
 
         return a_filtered, P_filtered, pt.atleast_1d(y_hat), pt.atleast_2d(F), ll_inner
 
-    def kalman_step(self, y, a, P, c, d, T, Z, R, H, Q):
+    def kalman_step(self, *args):
+
+        y, a, P, c, d, T, Z, R, H, Q = self.unpack_args(args)
+
         nan_mask = pt.isnan(y)
 
         W = pt.set_subtensor(pt.eye(y.shape[0])[nan_mask, nan_mask], 0.0)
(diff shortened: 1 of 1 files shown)


jessegrabowski:
Great catch -- thanks for this!

I think it makes sense to also add the missing value handling if you don't object. 

Also make sure you install and run pre-commit in your local dev environment. It's just `pip install pre-commit; pre-commit install; pre-commit run --all` (after you do install it will automatically run each time you git commit)

jessegrabowski (approved):

jessegrabowski:
Thanks @sebcroft ! Sorry it took a minute to get in

PR #654, opened 2026-02-20
Named StateSpaceModel v2 – rebase and updates after #607

opherdonchin:
## Summary

This PR reintroduces the Named `StateSpaceModel` changes after rebasing onto `main` following the merge of #607.

The previous version was submitted as #611. This update incorporates the changes introduced in #607 and resolves the resulting conflicts while preserving the intended named state space behavior.

No new conceptual changes are introduced beyond what was discussed in #611.

## Changes

- Rebased named state space branch onto current `main`
- Resolved conflicts introduced by #607
- Ensured compatibility with updated state space internals
- Cleaned working tree (no unrelated changes included)

## Tests

Environment:
- Linux (Fedora) and Windows 11
- Python 3.14.2 (Linus) and 3.12 (Windows)
- Commit: b173e6e

Results:

- `tests/statespace/core`
  - 182 passed
  - 2 skipped (JAX tests require `nutpie`)
  - 0 failed

- `tests/statespace/filters`
  - 59 passed
  - 1 failed (`test_kalman_filter_jax[cholesky]`)

The single failing test occurs in the JAX Cholesky filter path and appears unrelated to the named model changes. All standard (non-JAX) filter paths and core state space tests pass.

## Notes

- This PR supersedes #611.
- The branch has been rebased onto current `main`.
- No out-of-scope fixes are included.


pymc_extras/statespace/core/statespace.py
```diff
@@ -685,18 +699,20 @@ def make_and_register_data(
         """
         if name not in self.data_names:
             raise ValueError(
-                f"{name} is not a model parameter. All placeholder variables should correspond to model "
-                f"parameters."
+                f"{name} is not a model data-variable. All placeholder variables should correspond to model "
```
jessegrabowski:
```suggestion
                f"{name} is not a model data variable. All placeholder variables should correspond to model "
```

pymc_extras/statespace/core/statespace.py
```diff
@@ -685,18 +699,20 @@ def make_and_register_data(
...
+                f"data-variables."
```
jessegrabowski:
```suggestion
                f"data variables."
```

tests/statespace/test_namespace.py
```diff
@@ -0,0 +1,13 @@
+import pymc as pm
+
+from pymc_extras.statespace.core.statespace import PyMCStateSpace
+
+
+def test_two_statespace_models_can_coexist_with_names(monkeypatch):
```
jessegrabowski: put this test in the test_model file, it doesn't need a new file.

jessegrabowski (changes requested):
Nice, this is looking really clean. Just a few requests.

Also be sure to install pre-commit (`pip install pre-commit; pre-commit install; pre-commit run --all`) so your PR gets formatted and linted correctly.

pymc_extras/statespace/core/statespace.py
```diff
@@ -305,6 +311,12 @@ def __init__(
             console = Console()
             console.print(self.requirement_table)
 
+    def graph_name(self, base: str) -> str:
```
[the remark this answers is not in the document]
opherdonchin: I went for prefixed_name

pymc_extras/statespace/core/statespace.py
```diff
@@ -649,14 +661,16 @@ def make_and_register_variable(
                 f"parameters."
             )
 
-        if name in self._tensor_variable_info:
+        gname = self.graph_name(name)
```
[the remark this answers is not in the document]
opherdonchin: Under the new scheme (base_name for the unprefixed name and name for the name in the graph) it makes most sense to just call this variable name.

pymc_extras/statespace/core/statespace.py
opherdonchin: Corrected.

pymc_extras/statespace/core/statespace.py
```diff
@@ -1852,6 +1874,7 @@ def sample_filter_outputs(
                 n_obs=self.ssm.k_endog,
                 obs_coords=obs_coords,
                 register_data=True,
+                data_name=self.graph_name("data"),
```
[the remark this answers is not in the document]
opherdonchin: Done

jessegrabowski:
Really like where this landed, it's looking very clean.

Could you double-check that we don't need to make any changes to Component or StructuralTimeSeries? Those are in the models/structural module, and are a sort-of parallel StatespaceModel that needs to be cleaned up some day.

jessegrabowski:
We should be able to use `graph_replace` to swap out all the placeholder variables with new placeholders that have the prefixed name. So the components themselves don't need to know about this naming convention, the `__init__` method in `StructuralStateSpace` can just handle it.

You can check `PyMCStateSpace._insert_random_variables` for a pattern on how to collect the dummies and do a replace, but it should look something like `replacements = {var: var.type(name=self.prefix_name(var.name)) for var in explicit_graph_inputs(statespace_matrices)}`

AlexAndorra:
Any help from me needed here @opherdonchin @jessegrabowski ?

opherdonchin:
@AlexAndorra: thanks for the nudge! 

I think what I've done makes sense and follows @jessegrabowski's suggestion. The branch is now rebased onto current main (v0.10.0) with a clean 3-commit history.

What changed in the new commit:

`StructuralTimeSeries.__init__` now uses `pytensor.graph_replace` to create fresh, prefixed placeholder variables in all SSM matrices when a model name is provided. A single `_prefix_placeholder_variables()` method collects all placeholders from `tensor_variable_info` and `tensor_data_info`, builds new variables via `old_var.type(name=self.prefixed_name(old_var.name))`, applies `graph_replace` to all 9 SSM matrices, and rebuilds the symbolic info objects with aligned names. Components remain unaware of the naming convention.

I added appropriate tests in `TestGraphReplacePlaceholderNamespacing`.

Let me know if this is a good fix and if anything else is needed before this is ready to merge?

Thanks!
Opher

pymc_extras/statespace/models/structural/core.py
```diff
@@ -234,6 +240,78 @@ def strip(names):
...
+        new_ssm = ssm.copy()
+        for mat_name, new_mat in zip(LONG_MATRIX_NAMES, replaced_matrices):
+            setattr(new_ssm, mat_name, new_mat)
+
+        new_variable_info = SymbolicVariableInfo(symbolic_variables=tuple(new_variables))
+        new_data_info = SymbolicDataInfo(symbolic_data=tuple(new_data_entries))
+
+        return new_ssm, new_variable_info, new_data_info
+
+    def _validate_symbolic_info(self):
```
jessegrabowski: This function is never called outside of tests. If you don't want a check in `__init__` we should make this a test utility.

tests/statespace/models/structural/test_core.py
```diff
@@ -195,3 +231,139 @@ def test_sequence_type_component_arguments(arg_type):
 
     assert ss_mod.k_endog == len(state_names)
     assert sorted(ss_mod.observed_states) == sorted(list(state_names))
+
+
+class TestGraphReplacePlaceholderNamespacing:
+    """Tests for the graph_replace-based placeholder namespacing in StructuralTimeSeries."""
```
jessegrabowski:
```suggestion

```
The docstring adds nothing the name doesn't already tell me

jessegrabowski: All of the tests need an AI cleanup pass. Tell it to make the tests maximally self-documenting with minimal additional comments

tests/statespace/core/test_statespace.py
```diff
@@ -440,6 +440,16 @@ def test_base_class_raises():
         )
 
 
+def test_two_statespace_models_can_coexist_with_names(monkeypatch):
```
jessegrabowski: I'd rather see a real end-to-end test here (including sampling, but using the mock_sample pymc.testing fixture)

pymc_extras/statespace/models/structural/core.py
```diff
@@ -234,6 +240,78 @@ def strip(names):
         self._shock_names = strip(self._shock_info.names)
         self._param_names = strip(self._param_info.names)
 
+    def _prefix_placeholder_variables(self, ssm, tensor_variable_info, tensor_data_info):
+        """Replace placeholder variables in SSM matrices with prefixed-name copies.
+
+        Creates new placeholder variables whose names are prefixed with the model name,
+        applies graph_replace to swap them into all SSM matrices, and rebuilds the
+        symbolic info objects to reference the new placeholders.
```
jessegrabowski:
```suggestion

```

pymc_extras/statespace/models/structural/core.py
```diff
@@ -234,6 +240,78 @@ def strip(names):
...
+            replacements[old_var] = new_var
+            new_data_entries.append(SymbolicData(name=new_var.name, symbolic_data=new_var))
+
+        if not replacements:
+            return ssm, tensor_variable_info, tensor_data_info
+
+        matrices = [getattr(ssm, name) for name in LONG_MATRIX_NAMES]
+        replaced_matrices = graph_replace(matrices, replace=replacements, strict=False)
+
+        new_ssm = ssm.copy()
```
jessegrabowski: is this copy necessary?

jessegrabowski (changes requested):
some small quibbles, but this looks pretty much good to go!

opherdonchin:
Hi [Jesse](https://github.com/jessegrabowski), thanks for the review! I think I've addressed all your suggestions:

- Trimmed the `_prefix_placeholder_variables` docstring to a one-liner
- Removed `ssm.copy()` (unnecessary, as confirmed)
- Moved the validation logic out of production code — `_validate_symbolic_info` is gone from [`structural/core.py`](pymc_extras/statespace/models/structural/core.py) and replaced with a `@staticmethod _variable_names_match` helper inside the test class
- Removed the redundant class docstring from `TestGraphReplacePlaceholderNamespacing`
- Did a cleanup pass on all tests in that class — stripped comments, made tests self-documenting by name
- Replaced the shallow `monkeypatch` test in [`test_statespace.py`](tests/statespace/core/test_statespace.py) with a real end-to-end test using `mock_pymc_sample` that builds two named models in the same `pm.Model` and samples from both

In the process of writing the end-to-end test I also found and fixed a bug: `"obs"` was hardcoded in `build_statespace_graph` in [`statespace.py`](pymc_extras/statespace/core/statespace.py), which would cause `ValueError: Variable name obs already exists` when two named models shared a `pm.Model`. Changed it to `self.prefixed_name("obs")`.

Let me know if anything needs further adjustment!

jessegrabowski:
looks like you need to run pre-commit. 

The P0 special handling is a code smell. The name of variable is just syntactic sugar in pytensor. When we do graph_replace it's done by object, so it doesn't matter what the name is. I *think* it should be safe to just rename P0 as well. Could you try and see? Give it 5 minutes of effort, if it doesn't work it's fine we can leave it for later.

Can you also fire up a notebook and make sure that all post estimation methods (sample_conditional_posterior/prior, sample_unconditional_posterior/prior, impulse_response_function, forecast) work fine with multiple models in one context? You don't have to write tests for all of them, just check once and verify.

opherdonchin:
Thanks again for the detailed review, @jessegrabowski. I wish I had this kind of comfort with the codebase. 

Here's what I've addressed since the last round:

- **Pre-commit**: ran `ruff-format`; one long line in test_statespace.py was wrapped.
- **P0 special-case handling removed**: dropped all the `if not sv.name.endswith("_P0")` / `if sv.name != "P0"` guards from test_core.py. As you suspected, `graph_replace` works by object identity, so the fact that `__setitem__` overwrites the tensor's `.name` to `"initial_state_cov"` doesn't affect correctness. The `_variable_names_match` helper is now a cleaner prefix-only check.
- **Verification notebook**: I ran through all six post-estimation methods (`sample_conditional_posterior/prior`, `sample_unconditional_posterior/prior`, `impulse_response_function`, `forecast`) with two named `LevelTrend` models sharing a single `pm.Model`. Everything comes back clean — no NaNs, no name collisions. I'm attaching the notebook here rather than committing it to the repo.

[verify_named_models.ipynb](https://github.com/user-attachments/files/26519844/verify_named_models.ipynb)

Let me know if there's anything else worth doing before this is merge-ready!
Opher

jessegrabowski:
Some tests are failing because the test model was being passed a name (like "data"). This was a harmless no-op before, but now results in the variable names changing and the test failing. Solution is to just not name these models.

Other failures outside statespace are not your problem.

opherdonchin:
## Fixes in this update (commits `533bd76`, `db0527c`)

### 1. Dead code in `StructuralTimeSeries.__init__`

`self._name = name or "StructuralTimeSeries"` was set but never read anywhere. The real attribute `self.name` is set by `PyMCStateSpace.__init__` via `super().__init__(name=name)`, and `prefixed_name()` uses `self.name` throughout. The docstring was also incorrectly claiming a default of `name` to `"StructuralTimeSeries"`. Removed the dead line and corrected the docstring.

### 2. Stale docstring in `Component.build()`

The `name` parameter description read *"Name of the exogenous data being modeled. Default is 'data'"* — leftover legacy text for `data_name`, not the model prefix. Updated with a correct description. I also explicitly mentioned `model.prefixed_name(p)` as the way to get the expected prior name for each parameter — e.g. after `ssm = (trend + seasonal).build(name="m1")`, you'd write `pm.Normal(ssm.prefixed_name("initial_trend"), ...)` rather than hard-coding `"m1_initial_trend"`.

### 3. `getattr` → `name in pymc_model` in `_insert_random_variables` / `_insert_data_variables`

The old `getattr(pymc_model, name, None) is not None` would return truthy for any model attribute or property that happened to share a name with a parameter (e.g. `logp`, `compile_fn`). Changed to `name in pymc_model`, which uses PyMC's `__contains__` and checks only `named_vars`.

### 4. Question: `name=` support in SARIMAX / VARMAX / ETS / DFM

Issue #602 explicitly called for adding `name=` to the `__init__` of `BayesianSARIMAX`, `VARMAX`, `ETS`, and `DFM`. None of these currently accept or forward `name=` to `super().__init__()`, so users can't use multiple instances of those models in a single PyMC model. Should this be addressed here, or tracked as a follow-up issue?

jessegrabowski:
Thanks for the changes. I'm not an MCP server, please speak to me like a human.

To answer your question, you can do either. It's in scope for the PR, but I also understand if you want to wrap it up and open an issue for later.

opherdonchin:
Sorry. 

It's not about the AI assist so much as this being my first PR and I'm really not sure about how things work. 

If it's ok to  put this out there, that would be great. I have colleagues who want to be able to use it from the official version and they don't need the other structures. 

I'll open an issue and get the other structures working, though. It shouldn't take much time.

Opher

jessegrabowski:
AI's great and this PR is great too. Sorry I'm just on edge because I've been dealing with a slop avalanche elsewhere. 

We can merge after tests pass (except the ones that arent your issue)

opherdonchin:
I appreciate it. Thanks! 

This was really fun and educaitonal to work on. If you want to point me at something else, feel free. I'll start on it after I get the other models done.

Opher

opherdonchin:
Sorry. Previous push didn't actually fix the bugs. Try now.

jessegrabowski:
> Sorry. Previous push didn't actually fix the bugs. Try now.

Did you push? I'll trigger the CI as soon as you do

opherdonchin:
Yes. I pushed.

jessegrabowski:
Are you sure? There's no new changes since your "try now" comment

opherdonchin:
I'm pretty sure I'm sure. When I said "try now" that was right after I pushed what I think is the last change. I just check the code in the commit that is in the pull request and the name parmeters that were causing the bugs have been removed, the dead-end self._name is gone from StructuralTimeSeries and the misleading doc strings have been updated.

We could run the tests again, but I'm at least 64% sure.

Opher

jessegrabowski:
Could you update your local main and rebase onto main? I just merged a PR that clears the test failures that we were seeing here (but that were unrelated to your changes).

I can help with it if you're not up for a git adventure.

Dekermanjian:
Hi @opherdonchin,

I have pulled the PR and cleaned up the tests so that they pass. However, there is some additional work to be done here specifically with doing the same prefixing with the coords and figuring out what this means when sampling from the posterior. Can you help me understand what applications you are hoping to use this feature for?

opherdonchin:
@jessegrabowski 

I'm sorry. I got caught up in a different project for a bit. I'll do that now.

@Dekermanjian 

Good point about the coords, but I think in terms of sampling there is only one posterior with multiple models in it. I'm not sure how that would be affected. 

The use case we had was as follows. Multiple subjects each did an experiment. The behavior of each subject in the experiment is modeled with a state space model. However, we want to have hierarchical priors on the parameters of the state space model from multiple reasons: regularization, estimation of the variance across individuals, and between group comparisons. We did not see a convenient way to do this with a single state space model although it may in fact be possible and this approach seemed simple and intuitive.

It seems like there may be other use cases, for instance if you want to combine data from two very different types of time dependent processes.

Opher

Dekermanjian:
> The use case we had was as follows. Multiple subjects each did an experiment. The behavior of each subject in the experiment is modeled with a state space model. However, we want to have hierarchical priors on the parameters of the state space model from multiple reasons: regularization, estimation of the variance across individuals, and between group comparisons. We did not see a convenient way to do this with a single state space model although it may in fact be possible and this approach seemed simple and intuitive.

Hey @opherdonchin, have you seen this example on [multivariate SSM](https://github.com/AlexAndorra/pydata-berlin-statespace-models/blob/main/notebooks/multivariate-casestudy.ipynb) that allows you to set hierarchical priors on your observed states? It assumes your time series across subjects are equal, if they aren't it will interpolate so beware.

Dekermanjian:
@opherdonchin I opened a PR on your branch `named-statespace-v2` that fix the failing tests. If you still wish to work on this you will need to implement the prefixed names on the coords.

PR #663, opened 2026-03-06
MarginalRV: fix inner graph mutation when deriving support_point

ricardoV94:
Also adds a test for sampling compatibility with numpyro (although we don't seem to have it installed in the CI). This would actually depend on https://github.com/pymc-devs/pytensor/pull/1943


jessegrabowski (approved):

ricardoV94:
Maybe we should test explicitly that calling model.initial_point(), doesn't mutate the marginal RV inner graph?

jessegrabowski:
Sure, that seems reasonable. Is it the source of the pm.draw bug I was having?

ricardoV94:
Yes

PR #668, opened 2026-04-06
Compatibility with PyMC V6

jessegrabowski:
Placeholder issue. First task is to pull in https://github.com/pymc-devs/pymc-extras/pull/642


ricardoV94:
PyMC V6 is again compatible with PyTensor V3 so time to try again

ricardoV94:
need to install netcdf manually

jessegrabowski:
test_censored_xdist is warning DeprecationWarning: Calling an XRV without an explicit rng is deprecated. Use xt.random.rng or xt.random.shared_rng

It's coming upstream from pymc, we need to update over there. I don't think we fail on warning in PyMC so we missed it.

ricardoV94:
Yeah we did, we fail on some places not all.

ricardoV94:
Censored fixed by https://github.com/pymc-devs/pymc/pull/8255

ricardoV94:
@jessegrabowski is this ready, need a rebase only?

jessegrabowski:
Need to pull in the fix from https://github.com/pymc-devs/pymc-extras/pull/674, which is now closed so idk what's going on @fonnesbeck

fonnesbeck:
It was closed because Osvaldo fixed the underlying issue in ArviZ, so the problem should not exist in ArviZ 1.0 anymore.

aloctavodia:
The fix is available by installing arviz-stats from main. We are planning to do a release this Thursday or Friday.

ricardoV94:
We're installing preliz and pymc from main, in follow up we should go back to official releases (once those are out).

jessegrabowski:
> We're installing preliz and pymc from main, in follow up we should go back to official releases (once those are out).
I thought we specifically said in discord that we wanted to test against main

ricardoV94:
>  I thought we specifically said in discord that we wanted to test against main

I asked what we were doing, not what we wanted to do...

jessegrabowski:
But we don't want to? I thought it was a good idea.

ricardoV94:
> But we don't want to? I thought it was a good idea.

Once you release you want it to be compatible with the last released version of PyMC not the unreachable code that's in main... so I would say no?

PR #670, opened 2026-04-10
Fix failing tests

jessegrabowski:
Uncovered in https://github.com/pymc-devs/pymc-extras/pull/654 

There are some upstream changes in pymc/pytensor/preliz (unrelated to v3/v6) that are causing failing tests here. Nothing exciting.

### DIFF
--- a/pymc_extras/inference/smc/sampling.py
+++ b/pymc_extras/inference/smc/sampling.py
@@ -215,7 +215,10 @@ def arviz_from_particles(model, particles):
         strace = NDArray(name=model.name)
         strace.setup(n_particles, 0)
     for particle_index in range(0, n_particles):
-        strace.record(point={k: np.asarray(by_varname[k][0][particle_index]) for k in varnames})
+        strace.record(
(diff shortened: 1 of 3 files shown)


zaxtax:
Should we update the version bounds for preliz? Otherwise LGTM!

jessegrabowski:
Why is the preliz version implicated? Let's a a round of version pin updates next, a lot of stuff looks very out of date.

zaxtax:
Oh, because presumably this fix wasn't needed on older versions of preliz. So unless we are sure we still support all versions, why not tighten the bound?

jessegrabowski:
The core issue is a warning because preliz is using depreciated `.broadcastable` attribute from pytensor tensors. Correct fix is to filter this warning in pytest config and do a preliz patch. No need to version pin.

jessegrabowski:
I opened a PR in preliz to fix the warning: https://github.com/arviz-devs/preliz/pull/710

zaxtax (approved):
LGTM!

PR #674, opened 2026-04-19
Fix for divide by zero in psislw call from Pathfinder

fonnesbeck:
This is intended to fix the divide by zero issue by wrapping `logiw` in `xr.DataArray` with `dims="__sample__"` before calling `psislw`. This is  the documented-recommended input form per arviz's own `psislw` docstring. 

Regression test added.

### DIFF
--- a/pymc_extras/inference/pathfinder/importance_sampling.py
+++ b/pymc_extras/inference/pathfinder/importance_sampling.py
@@ -6,6 +6,7 @@ from typing import Literal
 
 import arviz as az
 import numpy as np
+import xarray as xr
 
 from numpy.typing import NDArray
 from scipy.special import logsumexp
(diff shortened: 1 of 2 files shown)


jessegrabowski:
We need this to target the v0.11 branch on my fork of pymc-extras [here](https://github.com/pymc-devs/pymc-extras/pull/668) . @ricardoV94 refused to help me figure out how to set up things like on pymc/pytensor to auto-target it, and I am playing helpless in order to complain for comedic effect.

This won't work on main because we're not cut over to arviz 1.0 yet, so CI will fail.

aloctavodia:
`az.psislw` doesn't exist anymore on its own. Currently, that function should be used as an accessor (as already done in #668 ) or from the `array_stats` interface. I added a guard here https://github.com/arviz-devs/arviz-stats/pull/353/changes/bee38918d5a091ea012cd523ae3ff45d5a981a8d. The test should pass then, as the _gpd_fit function should return np.nan, np.nan instead of getting into the divide by zero part.

PR #676, opened 2026-04-23
Add ConvergentFilter

### LINKED ISSUE #394: Use Kalman Filter convergence to better handle long time series
For system with many states, kalman filtering is slow because we have to compute the so-called Kalman gain matrix, $$K = P_\text{prior}Z^T F^{-1}$$, with $$F = ZP_{\text{prior}}Z^T + H$$ From there, we can compute the posterior covariance matrix, $$P_\text{posterior} = P_\text{prior} - KFK^T $$ These matrices have the following shapes and meanings:

| name | description                                         |                shape               |
| ------|---------------------------------------|---------------------------|
| Z       | Map hidden states to observed states | $n_\text{obs} \times n_\text{hidden}$|
| P       | Hidden state covariance matrix | $n_\text{hidden} \times n_\text{hidden}$|
| H      | Measurement error covariance | $n_\text{obs} \times n_\text{obs}$|
| F       | Estimated residual covariance | $n_\text{obs} \times n_\text{obs}$ |
| K       | Kalman gain (optimal update to hidden states) | $n_\text{hidden} \times n_\text{obs}$|

Typically the number of observed states will be quite small relative to the number of hidden states, so the inversion of $F$ isn't so bad. But when the number of hidden states is large, computing $$P_\text{posterior}$$ requires at least 3 matrix multiplications. Actually it's more, because we use the "Joseph Form", $$P_\text{posterior} = (I - KZ) ) P_\text{prior} (I - KZ)^T + KHK^T$$, which guarantees the result is always positive semi-definite (because it avoids subtraction between two PSD matrices), but instead costs 9 matrix multiplications.

Then, to compute next-step forecasts, we have to do several more multiplications:

$$ P_\text{prior}^+ = T P_\text{posterior} T^T + R Q R^T $$

Where $$T$$ is the same size as $$P$$ (it's the system dynamics), while $$R$$ and $$Q$$ are relatively smaller (they have to do with how shocks enter the system).

My point is this is all quite expensive to compute. Interestingly though, $$P_\text{posterior}$$ does not depend on the data at all, and if there are no time-dependent matrices, the covariance matrix will converge to a fixed point given by:

$$P = TPT^T + RQR^T - TPZ^T(ZPZ^T +H)^{-1}ZPT^T$$

This is an algebraic riccati equation, and can be solved in pytensor as `pt.linalg.solve_discrete_are(A=T.T, B=Z.T, Q=R @ Q @ R.T, R=H)`. Once we have this, we actually don't need to compute $P_\text{prior}$ or $P_\text{posterior}$, ever again, we can just use the steady-state $P$. Actually, once $P$ is fixed, so is $F^{-1}$, so the filter updates become *extremely* inexpensive.

How to use this is not 100% clear to me. I had previously made a `SteadyStateFilter` class that computed and used the steady-state $P$ from the first iteration. This means we don't do "online learning" for the first several steps. That *seems* OK to me, since we don't believe all that initial noise anyway. But on the other hand I've never seen this approach suggested in a textbook, so it makes me a bit suspicious that it's the right thing to do. I'm not against offering this option, but one negative is that JAX doesn't offer `solve_discrete_are`, and right now statespace is basically 100% dependent on JAX for sampling.

The "safer" option would be to use an `ifelse` in the `update` step to check for convergence. At every iteration of `kalman_step`, we can compute $||P_\text{prior}^- - P_\text{prior}^+||$ and stop computing updates once its below some tolerance. Here's a graph of the supremium norm for a structural model with level, trend, cycle, and 12 seasonal lags. $P_0$ was initialized to `np.eye(15) * 1e8`. The two plots are the same, but the right plot begins at t=20:

Untitled

Here's a table of tolerance levels and convergence iterations:

Tolerance | Convergence Iteration
----------|------------------------|
1 | 25 |
1e-1| 49 |
1e-2| 108|
1e-3 | 216|
1e-4 | 337|
1e-5 |457|
1e-6 |583|
1e-7 |703|
1e-8 |827|

We could leave convergence tolerance as a free parameter for the user to play with. But we can see that if we pick `1e-2` for instance, anything after 100 time steps is basically free. This would be quite attractive for estimating large, expensive systems or extremely long time series.

jessegrabowski:
Closes #394
Related to #332 

For certain classes of models, the latent state covariance matrix estimated by the Kalman Filter converges to a fixed point, after which is is no longer necessary to compute via expensive linear algebra. This PR adds a new filter, `ConvergentFIlter`, that exploits this property. To take advantage, a model must:

- Have no missing values, and
- No time varying matrices

The `ConvergentFilter` uses a `while` scan, so this cannot be used with jax. All timings are done with numba.

These break the convergence assumptions. But nothing else is required. If these criteria are met, we get big speedups. Here is a table of models, where `n` is the number of timesteps, `m` is the number of hidden states, `p` is the number of observed states, and `r` is the number of shocks. The biggest speedups come from long time series, where there is more iterations in the compute-saving post-convergence scan. Here are some timings for the compiled `logp_dlogp` function for models taken from the example notebooks:

| Model                              | Shape (m, p, r) | n    | StandardFilter | ConvergentFilter | Speedup |
|------------------------------------|-----------------|------|----------------|------------------|---------|
| VARMAX(2,0) 5-obs                  | (10, 5, 5)      | 200  | 3.4            | 1.3              | 2.5x    |
| VARMAX(2,0) 5-obs                  | (10, 5, 5)      | 1000 | 17.2           | 5.1              | 3.4x    |
| VARMAX(2,0) 5-obs                  | (10, 5, 5)      | 5000 | 96.0           | 26.6             | 3.6x    |
| SARIMAX(1,0,1)                     | (2, 1, 1)       | 200  | 1.1            | 0.6              | 1.9x    |
| SARIMAX(1,0,1)                     | (2, 1, 1)       | 1000 | 5.4            | 2.1              | 2.6x    |
| SARIMAX(1,0,1)                     | (2, 1, 1)       | 5000 | 27.9           | 10.1             | 2.8x    |
| SARIMAX(2,1,2)(2,0,2,12)           | (28, 1, 1)      | 200  | 5.7            | 3.1              | 1.8x    |
| SARIMAX(2,1,2)(2,0,2,12)           | (28, 1, 1)      | 1000 | 30.7           | 14.8             | 2.1x    |
| SARIMAX(2,1,2)(2,0,2,12)           | (28, 1, 1)      | 5000 | 153.7          | 63.9             | 2.4x    |
| ETS(A,Ad,A,12)                     | (15, 1, 1)      | 200  | 3.4            | 1.8              | 1.9x    |
| ETS(A,Ad,A,12)                     | (15, 1, 1)      | 1000 | 17.6           | 6.4              | 2.8x    |
| ETS(A,Ad,A,12)                     | (15, 1, 1)      | 5000 | 90.7           | 33.6             | 2.7x    |
| DFM 1-factor 2-lag 4-obs           | (10, 4, 5)      | 200  | 2.9            | 1.5              | 1.9x    |
| DFM 1-factor 2-lag 4-obs           | (10, 4, 5)      | 1000 | 15.6           | 4.9              | 3.2x    |
| DFM 1-factor 2-lag 4-obs           | (10, 4, 5)      | 5000 | 78.3           | 24.8             | 3.2x    |
| Structural LvlTrend+FreqSeas (airpass) | (14, 1, 14) | 200  | 2.5            | 1.5              | 1.7x    |
| Structural LvlTrend+FreqSeas (airpass) | (14, 1, 14) | 1000 | 12.8           | 4.6              | 2.8x    |
| Structural LvlTrend+FreqSeas (airpass) | (14, 1, 14) | 5000 | 66.5           | 23.3             | 2.9x    |

Special thanks to @JeanVanDyk, who derived components of the closed forms used in the handoff step between the two scans. I tried using the full analytic forms he worked out end-to-end, but it was slower than autodiff, I believe because of the number of matmuls involved. It might be faster on GPU, so I want to revist that again at some point.

### DIFF
--- a/pymc_extras/statespace/filters/kalman_filter.py
+++ b/pymc_extras/statespace/filters/kalman_filter.py
@@ -845,7 +845,9 @@ class UnivariateFilter(BaseFilter):
 class ConvergentFilter(StandardFilter):
     """Kalman filter that exploits Riccati convergence to a fixed-gain steady state.
 
-    For a stationary observable system with no missing data, the predicted-covariance Riccati recursion
+    For a time-invariant system whose Riccati recursion reaches a steady state -- detectability and
+    stabilizability are sufficient, which is weaker than stationarity and admits unit-root models such
+    as the local level -- the predicted-covariance recursion
     ``P_{t+1|t} = T (P_{t|t-1} - K_t F_t K_t^T) T^T + R Q R^T`` is data-independent and converges to the
     discrete algebraic Riccati equation (DARE) fixed point ``P_star``. Once converged, the Kalman gain
     ``K_t = P_{t|t-1} Z^T F_t^{-1}`` and innovation covariance ``F_t = Z P_{t|t-1} Z^T + H`` are constant
@@ -878,7 +880,11 @@ class ConvergentFilter(StandardFilter):
     Constraints (validated at graph build time):
 
     - All model matrices (``c, d, T, Z, R, H, Q``) must be time-invariant. Time-varying inputs raise
-      :class:`ValueError`; Riccati convergence is not meaningful for a non-stationary system.
+      :class:`ValueError`: a constant steady-state gain only exists when the system matrices are
+      constant. This is a statement about time-invariance, not stationarity -- a time-invariant model
+      whose Riccati recursion never settles within the series simply never triggers the ``until``
+      clause, so ``k = n``, the post-convergence tail is empty, and the result reduces to
+      :class:`StandardFilter`.
     - ``data`` must contain no missing values. A masked observation changes the effective ``Z_t`` (and
       so ``K_t`` and ``F_t``), breaking the cached-``K_star`` assumption. The statespace core replaces
       ``NaN`` with ``missing_fill_value`` before the filter runs, so both ``NaN`` and that sentinel are
--- a/pymc_extras/statespace/filters/kalman_filter.py
+++ b/pymc_extras/statespace/filters/kalman_filter.py
@@ -870,7 +870,7 @@ class ConvergentFilter(StandardFilter):
 
     2. **Speed.** In the post-convergence tail, per-step Cholesky and the full Joseph-form backward are
        unnecessary because ``K``, ``F``, and ``P`` are constant. The custom pullback hoists
-       ``F^{-1}, H^{-1}, A = T (I - K_star Z), I - K_star Z`` out of the scan body and the per-step
+       ``F^{-1}, A = T (I - K_star Z), I - K_star Z`` out of the scan body and the per-step
        backward reduces to matvec and outer products. On typical state-space cells
        (m=10..50, p=3..10, n=2000..10000) this gives a ~1.8--2.9x speedup on ``logp + grad``
        versus :class:`StandardFilter` autodiff, on top of a ~7--12x forward speedup.
@@ -884,6 +884,9 @@ class ConvergentFilter(StandardFilter):
       ``NaN`` with ``missing_fill_value`` before the filter runs, so both ``NaN`` and that sentinel are
       rejected. For :class:`SharedVariable` or :class:`TensorConstant` data the check runs at build
       time; fully symbolic data gets an :class:`~pytensor.raise_op.Assert` that fires at runtime.
+
+    Measurement-error-free models (singular or zero ``H``) are fully supported, gradients included;
+    the tail backward avoids forming ``H^{-1}``.
 
     Only ``d loglike_obs`` is supported as an upstream gradient; other output adjoints are treated as
     disconnected. For gradients of other filter outputs, use :class:`StandardFilter`.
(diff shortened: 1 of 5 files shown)


pymc_extras/statespace/filters/kalman_filter.py
```diff
@@ -829,3 +834,527 @@ def kalman_step(self, *args):
...
+        m_shape = (n_fx, self.n_states, self.n_states)
+        P_filt_fx = pt.broadcast_to(P_filt_star, m_shape)
+        P_hat_fx = pt.broadcast_to(P_star, m_shape)
+        F_fx = pt.broadcast_to(F_star, (n_fx, self.n_endog, self.n_endog))
+
+        def cat(a, b):
+            return pt.concatenate([a, b], axis=0)
+
+        return (
+            cat(a_filt_tr[:k], a_filt_fx),
```
ricardoV94, line 1020: If this is just initial_state concat buffer you can get from scan(...).owner.inputs[0] like a king

ricardoV94 (approved):

PR #533, opened 2025-07-02
Make basic INLA interface and simple marginalisation routine

Michal-Novomestsky:
Addresses #532 and #344.

Relies on https://github.com/pymc-devs/pytensor/pull/1582.

PR #677, opened 2026-04-30
Move Q into Graph

Michal-Novomestsky:
Part of debugging for https://github.com/pymc-devs/pymc-extras/pull/533


Michal-Novomestsky:
Rebased onto #533

PR #678, opened 2026-05-06
Representation simplification

jessegrabowski:
I originally copied the `Representation` class from statsmodels to get started as quickly as possible. Over time I've realized I made some mistakes. This PR goes back to fix them.

- The design assumed all matrices were time varying, and had a silent batch dimension. This is silly, pytensor has `vectorize_graph`, we don't need to early add and slice away a secret batch dim
- Likewise, that secret batch dim led to add matrices being built via set_subtensor, which results in some gnarly Ops.
- Also, statespace matrices were only allowed to be up to 3d. Any higher and it would raise. Certain matrices (x0, P0) were forbidden from having batch dims. All of this got in the way of e.g. #450 
- Logic for deciding what was time varying and what was static at the Kalman Filter was too complex, too magical, and again ruled out batch dims.

This PR:

- Removes all of the time-varying and dimension checks. Batch away.
- Requires models to explicitly register a matrix as time-varying using the `self.ssm.declare_time_varying` method.
- The list of time-varying matrices is set explicitly and passed around to where it is needed.  
- Assigning matrices to the `ssm` should be more ergonomic, `self.ssm['transition'] = T` instead of `self.ssm['transition', : ,:] = T`

### DIFF
--- a/pymc_extras/statespace/core/compile.py
+++ b/pymc_extras/statespace/core/compile.py
@@ -5,7 +5,7 @@ import pytensor.tensor as pt
 
 from pymc_extras.statespace.core import PyMCStateSpace
 from pymc_extras.statespace.filters.distributions import LinearGaussianStateSpace
-from pymc_extras.statespace.utils.constants import SHORT_NAME_TO_LONG
+from pymc_extras.statespace.utils.constants import LONG_NAME_TO_SHORT
 
 
 def compile_statespace(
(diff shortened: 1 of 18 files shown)


ricardoV94 (approved):
But generality???

Not joking was the final graph with a dummy time dimension easy to rewrite?

jessegrabowski:
> Not joking was the final graph with a dummy time dimension easy to rewrite?

Harder, because I had stupid Subtensor Ops wrapping everything. I was allocating `transition = zeros((1, *shape))` then `ssm['transition', :, :] = x` was doing `transition[0, :, :].set(x)`. So that had to always either be rewritten away or I had to hope whatever rewrite I was doing could reason through that. Very dumb, very pointless.

ricardoV94:
Isn't the end goal that all these matrices can be time-varying one day?

jessegrabowski:
Yes and they are, this PR doesn't change that. You just have to explicitly tell it. Before:

```
        S_I = pt.zeros((self.n_timesteps, self.k_states))
        forcing = self.make_and_register_variable("forcing", shape=(self.n_timesteps,))
        S_I = S_I[:, 1].set(self.dt*forcing)
        self.ssm["state_intercept"] = S_I
```

Now:

```
        forcing = self.make_and_register_variable("forcing", shape=(self.n_timesteps,))
        self.ssm["state_intercept"] = self.dt*forcing[:, None]
        self.ssm.declare_time_varying('state_intercept')
```

ricardoV94:
so you made the non-time varying simpler and the time varying more verbose. Seems like a good choice UX-wise

jessegrabowski:
It also let me drop a bunch of logic in the scan to try to guess which ones are time varying. Now batch dims are just batch dims unless the matrix is explicitly marked as time-varying, then the far-left dim is expected to be time and it gets added in as a sequence

Dekermanjian (approved):
This looks great! I like that we are explicitly defining our time varying matrices now.

PR #679, opened 2026-05-06
Remove redundant pt.specify_shape calls

jessegrabowski:
Once upon a time, pytensor's static shape inference was a lot worse, especially in the linalg Ops. That's all been cleaned up, so most of the `specify_shape` Ops in statespace are no longer needed. They're harmless, but still cruft.


ricardoV94 (approved):
nice

PR #683, opened 2026-05-13
Bump version for release

jessegrabowski:
- Remove git install for pymc/pytensor
- fix bug in statespace fixture scopes
- remove python 3.11 from our supported tags in pyproject.toml
- bump ruff pyupdate version to 3.12

### DIFF
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -14,9 +14,9 @@ dependencies:
   - pydantic>=2.0.0
   - h5netcdf
   - pytensor>=3.0,<4.0
+  - pymc>=6.0,<7.0
   - pip
   - pip:
       - jax
       - blackjax
-      - pymc>=6.0,<7.0
       - preliz>=0.25
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -4,7 +4,7 @@ channels:
 - nodefaults
 dependencies:
   - scikit-learn
-  - better-optimize>=0.3.2
+  - better-optimize>=0.4.1
   - dask<2025.1.1
   - xhistogram
   - statsmodels
@@ -13,9 +13,10 @@ dependencies:
   - pytest-cov
   - pydantic>=2.0.0
   - h5netcdf
+  - pytensor>=3.0,<4.0
   - pip
   - pip:
       - jax
       - blackjax
-      - pymc @ git+https://github.com/pymc-devs/pymc.git
-      - preliz @ git+https://github.com/arviz-devs/preliz.git
+      - pymc>=6.0,<7.0
+      - preliz>=0.25
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -13,7 +13,6 @@ dependencies:
   - pytest-cov
   - pydantic>=2.0.0
   - h5netcdf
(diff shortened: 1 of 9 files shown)


conda-envs/environment-test.yml
```diff
@@ -13,9 +13,10 @@ dependencies:
...
   - pydantic>=2.0.0
   - h5netcdf
+  - pytensor>=3.0,<4.0
   - pip
   - pip:
       - jax
       - blackjax
-      - pymc @ git+https://github.com/pymc-devs/pymc.git
-      - preliz @ git+https://github.com/arviz-devs/preliz.git
+      - pymc>=6.0,<7.0
```
ricardoV94: shouldn't it not be pip anymore? and preliz, is it only in pip?

pyproject.toml
```diff
@@ -34,9 +34,9 @@ keywords = [
 license = {file = "LICENSE"}
 dynamic = ["version"]  # specify the version in the __init__.py file
 dependencies = [
-  "pymc@git+https://github.com/pymc-devs/pymc.git",
+  "pymc>=6.0,<7.0",
   "scikit-learn",
-  "better-optimize>=0.3.1",
+  "better-optimize>=0.4.1",
   "pydantic>=2.0.0",
   "preliz>=0.20.0",
```
ricardoV94: doesn't this need more? also pin<1?

conda-envs/environment-test.yml
jessegrabowski: i'm waiting for conda-forge to pick up the new versions. did it this way so i could check the CI, i will update before merging this

pyproject.toml
jessegrabowski: Yeah the dependencies are pretty thin. I will review it.

conda-envs/environment-test.yml
ricardoV94: got it

pyproject.toml
ricardoV94: I meant preliz needs a higher lower pin no?

jessegrabowski: I went though everything, basically only arviz was missing as a strict dependency.

conda-envs/environment-test.yml
```diff
@@ -13,9 +13,10 @@ dependencies:
   - pytest-cov
   - pydantic>=2.0.0
   - h5netcdf
+  - pytensor>=3.0,<4.0
```
ricardoV94: PyMC always pins a specific range of pytensor so this is not really needed. Not saying it matters

pyproject.toml
```diff
@@ -34,11 +34,12 @@ keywords = [
 license = {file = "LICENSE"}
 dynamic = ["version"]  # specify the version in the __init__.py file
 dependencies = [
-  "pymc@git+https://github.com/pymc-devs/pymc.git",
+  "pymc>=6.0,<7.0",
+  "arviz>=1.1",
   "scikit-learn",
```
ricardoV94: Unrelated: scikit-lean is pretty heavy dep no? where do we use it?

ricardoV94 (approved):

jessegrabowski:
I had to pin `micromamba-version: "2.6.0-0"` to get the windows ci runner to work. Worth keeping in mind in case we hit this in other packages.

PR #685, opened 2026-05-20
batched SSM compatibility

Dekermanjian:
The following is a WIP implementation of allowing the SSM module to work with multiple time-series via batched inputs. 

This first draft allows batched sampling and supports the following internals:

* `sample_unconditional_prior`
* `sample_conditional_prior`
* `sample_unconditional_posterior`
* `sample_conditional_posterior`

Additional testing is still required and redundancies in code need to be cleaned-up.

Update 1:
I updated the way batching is defined from implicitly defined to explicitly defined by a `batch_size` parameter and I fixed an error that I introduced in the data registry control flow. All `StateSpace` tests now are passing locally. This just means that adding the batching implementation did not break any existing logic. 

@jessegrabowski, Here are the next steps before this will be ready:

- [ ] Quick review just so that I know directionally if this is how the vision of batching should be
- [x] Continue to add additional `StateSpace` methods support until all methods are supported
- [ ] Write tests for batching that runs through the full SSM workflow


pymc_extras/statespace/core/statespace.py
```diff
@@ -278,6 +280,7 @@ def __init__(
         self.k_posdef = k_posdef
         self.measurement_error = measurement_error
         self.mode = mode
+        self.batch_size = batch_size
```
jessegrabowski: This is an abstraction leakage. The model constructor should be agnostic to the data, but this requires that I choose the batch dims up front.

pymc_extras/statespace/core/statespace.py
```diff
@@ -1011,6 +1022,115 @@ def _register_kalman_filter_outputs_with_pymc_model(outputs: tuple[pt.TensorVari
...
+                Z,
+                R,
+                H,
+                Q,
+                missing_fill_value,
+                cov_jitter,
+                time_varying_names,
+            )
+
+        input_signature = ",".join(
```
jessegrabowski: this should be extracted out to a make_signature function that can be reused across call sites

pymc_extras/statespace/core/statespace.py
```diff
@@ -1092,32 +1212,48 @@ def build_statespace_graph(
...
         self._save_exogenous_data_info()
         self._insert_data_variables()
 
         obs_coords = pm_mod.coords.get(OBS_STATE_DIM, None)
         self._fit_data = data
 
+        data_dims = None
+
+        if self.batch_size:
+            data_dims = (BATCH_DIM, TIME_DIM, OBS_STATE_DIM)
```
jessegrabowski, line 1283: What if there are more than one batch dim?

pymc_extras/statespace/core/statespace.py
```diff
@@ -1092,32 +1212,48 @@ def build_statespace_graph(
...
         self._insert_data_shape_into_n_timesteps(data)
 
-        filter_outputs = self.kalman_filter.build_graph(
-            pt.as_tensor_variable(data),
-            *self.unpack_statespace(),
-            missing_fill_value=missing_fill_value,
-            cov_jitter=cov_jitter,
-            time_varying_names=self.ssm.time_varying_names,
-        )
+        if data_dims and BATCH_DIM in data_dims:
```
jessegrabowski, line 1297: we should be able to always call vectorized filter. if there are no batch dims, it should just pass through

pymc_extras/statespace/core/statespace.py
```diff
@@ -1307,34 +1451,62 @@ def _kalman_filter_outputs_from_dummy_graph(
...
-            c,
-            d,
-            T,
-            Z,
-            R,
-            H,
-            Q,
-            time_varying_names=self.ssm.time_varying_names,
-        )
+        if data_dims and BATCH_DIM in data_dims:
```
jessegrabowski, line 1512: Again you shouldn't need this check, just always call vectorize

pymc_extras/statespace/core/statespace.py
```diff
@@ -1011,6 +1022,115 @@ def _register_kalman_filter_outputs_with_pymc_model(outputs: tuple[pt.TensorVari
...
+        )
+
+        signature = f"{input_signature}->{output_signature}"
+
+        return pt.vectorize(
+            vectorize_smoother,
+            signature=signature,
+        )
+
     def build_statespace_graph(
```
jessegrabowski, line 1187: This is where the entry point for all the batch information should be. The type of data needs to change to allow list[pd.DataFrame] and also xr.Dataarray. Ideally we can discover the batch dims by inspecting the pymc model

pymc_extras/statespace/core/statespace.py
```diff
@@ -2384,12 +2665,24 @@ def _build_forecast_model(
 
             mu_frozen, cov_frozen = graph_replace([mu, cov], replace=sub_dict, strict=True)
 
-            x0 = pm.Deterministic(
-                "x0_slice", mu_frozen[t0_idx], dims=mu_dims[1:] if mu_dims is not None else None
-            )
-            P0 = pm.Deterministic(
-                "P0_slice", cov_frozen[t0_idx], dims=cov_dims[1:] if cov_dims is not None else None
-            )
+            if self.batch_size:
```
jessegrabowski: Here you can construct a tuple of `slice(None, None)` based on the number of batch dims, then only have one slice case. Something like `s = [slice(None, None)] * (n_batch - 2) + [t0_idx, slice(None, None)]`

pymc_extras/statespace/core/statespace.py
```diff
@@ -2753,31 +3053,59 @@ def impulse_response_function(
                     Q = pt.linalg.cholesky(Q) / pt.diag(Q)
 
             if shock_trajectory is None:
-                shock_trajectory = pt.zeros((n_steps, self.k_posdef))
+                if self.batch_size:
+                    shock_trajectory = pt.zeros((self.batch_size, n_steps, self.k_posdef))
```
jessegrabowski: We should land on `batch_sizes : tuple[int]` to hold all the batch dim sizes, with `()` if there isn't a batch. Then `pt.zeros((*self.batch_sizes, n_steps, self.k_posdef))` is always valid, no need for cases

pymc_extras/statespace/core/statespace.py
```diff
@@ -2753,31 +3053,59 @@ def impulse_response_function(
...
             else:
                 shock_trajectory = pt.as_tensor_variable(shock_trajectory)
 
             time_varying_T = "transition" in self.ssm.time_varying_names
 
+            if self.batch_size:
+                shock_trajectory = shock_trajectory.swapaxes(0, 1)
+
+            def bmv(A, x):
+                return pt.matmul(A, x[..., None])[..., 0]
```
jessegrabowski: I thought we automatically do something like this already

pymc_extras/statespace/core/statespace.py
```diff
@@ -2793,7 +3121,13 @@ def irf_step(*args):
                 return_updates=False,
             )
 
-            pm.Deterministic("irf", irf, dims=[TIME_DIM, ALL_STATE_DIM])
+            if self.batch_size:
+                irf = irf.swapaxes(0, 1)
```
jessegrabowski, line 3170: This si the right pattern but you need to reason about the case where there are multiple batches

jessegrabowski (changes requested):
This is a really great start, i'll fire up the notebook and play it. Here are some thoughts off the cuff.

@ricardov94 your voice would be very welcome here :)

pymc_extras/statespace/filters/distributions.py
```diff
@@ -157,6 +157,9 @@ def rv_op(
         append_x0=True,
         method="svd",
     ):
+        def bmv(A, x):
```
jessegrabowski: There are a couple helpers in the PR that get repeated, make sure to factor stuff out to minimize code duplication before the final version

PR #686, opened 2026-05-26
Fix test failures

jessegrabowski:
- Filter assumption warning pending bugfix upstream (https://github.com/pymc-devs/pytensor/pull/2173)
- Tighten SARIMA priors to help get stationary draws

### DIFF
--- a/tests/statespace/models/test_SARIMAX.py
+++ b/tests/statespace/models/test_SARIMAX.py
@@ -6,11 +6,10 @@ import pymc as pm
 import pytensor
 import pytensor.tensor as pt
 import pytest
+import statsmodels.api as sm
 
 from numpy.testing import assert_allclose, assert_array_less
 from pymc.testing import mock_sample_setup_and_teardown
-
-import statsmodels.api as sm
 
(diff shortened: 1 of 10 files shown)


jessegrabowski:
I also turned off pre-commit.ci for this repo and added a pre-commit action, following pytensor's template. I got frustrated because pre-commit failed on remote and i couldn't see any logs. Also I'd rather our tests not run if pre-commit is failing.

tests/statespace/models/test_SARIMAX.py
Dekermanjian, line 465: Why were these tests failing? It looks like we are asserting shapes, containment of data variables, and checking that the data values match. How did the AR priors make these fail?

Dekermanjian (approved):
This looks good. I agree that the CI should exit early if pre-commit fails. I am just wondering how the AR parameters were making the tests fail.

tests/statespace/models/test_SARIMAX.py
jessegrabowski: The priors were non-stationary, but they set stationary_initialization = True. So the result of solve_discrete_lyapunov was NaN -> test failure

PR #688, opened 2026-05-29
Allow closed form marginalization and implement `conditional` model transform

ricardoV94:
This has been a long time in the making, it supersedes pre-LLM #441

### Summary

This PR reworks the internals of marginalization, moving the logic onto a pymc-logprob-like rewrite machinery, and adds a new `conditional()` API.

### What changes

- **Marginalization as rewrites.** The core logic is reimplemented as logprob rewrites (`rewrites.py`) — `MarginalSubgraph`/`DeferredMarginalSubgraph` Ops plus dispatchable node rewriters — instead of living in dedicated functions/methods. Each supported case becomes a self-contained rule.
  - This gives room to do cleanup rewrites to facilitate pattern matching, and it's extensible from the outside.

- **New `conditional()` API.** `conditional(model, rvs_to_recover)` returns a model where marginalized variables come back as free RVs carrying their conditional posterior distributions, ready for `sample_posterior_predictive` or `compile_logp`. Recovering all variables uses the chain-rule factorization `p(idx|y)·p(sub_idx|idx,y)`, so forward-sampling gives exact joint posterior draws.
  - The conditional itself is now a dispatch on the `MarginalRV` Op, so each marginalization case provides its own recovery rule. Previously this lived hardcoded inside `recover_marginals` and was finite-discrete-only and not extensible. `recover_marginals` is now built on top of `conditional()`

- **Reorganized distributions.** Per-distribution logic moves into a `distributions/` package: finite-discrete cases (Bernoulli, Categorical, DiscreteUniform, DiscreteMarkovChain) in `enumerable.py`, with a new `normal.py`.

- **Normal-Normal marginalization (POC).** A proof-of-concept closed-form conjugate path, mostly to show the rewrite framework can support continuous/analytic cases beyond finite-discrete enumeration. Not a polished feature yet — e.g. affine `mu = a*x + b` is not handled (currently xfailed).

- **Cleanup.** The deprecated, unused `MarginalModel` class is removed, and tests are reorganized/expanded (`test_model.py`, new `test_rewrites.py`).

```python
import numpy as np
import scipy.stats
import pymc as pm

from pymc_extras import marginalize, conditional

# A Normal-Normal model: latent x with a Normal prior, Normal likelihood on y
sigma_prior, offset, sigma_lik, y_obs = 3.0, 1.5, 4.0, 10.0
with pm.Model() as m:
    mu = pm.Normal("mu", 0, 10)
    x = pm.Normal("x", mu=mu, sigma=sigma_prior)          # latent to marginalize
    y = pm.Normal("y", mu=x + offset, sigma=sigma_lik, observed=y_obs)

# 1. Marginalize x out (closed-form conjugate), then sample the remaining params
marginal_m = marginalize(m, "x")
idata = pm.sample(model=marginal_m)

# 2. Recover x's conditional posterior as a free RV and forward-sample it
# recover_marginals(idata, model=marginal_m) does the same as this step,
cond_m = conditional(marginal_m)
pp = pm.sample_posterior_predictive(idata, model=cond_m, sample_vars=["x"])
print("posterior mean of x:", float(pp.posterior_predictive["x"].mean()))
# posterior mean of x: 7.44

# Evaluate the conditional logp directly (it's the conjugate Normal posterior)
logp_fn = cond_m.compile_logp(vars=[cond_m["x"]])

mu_val, x_test = 1.0, 3.0
prec_p, prec_l = 1 / sigma_prior**2, 1 / sigma_lik**2
post_prec = prec_p + prec_l
post_mu = (mu_val * prec_p + (y_obs - offset) * prec_l) / post_prec
expected = scipy.stats.norm.logpdf(x_test, post_mu, np.sqrt(1 / post_prec))

print("logp(x=3 | mu=1):", logp_fn({"mu": mu_val, "x": x_test}), " expected:", expected)
# logp(x=3 | mu=1): -1.8369420567636037  expected: -1.8369419927807948
```

PyMC examples update PR: https://github.com/pymc-devs/pymc-examples/pull/885


zaxtax:
I don't understand the difference between what conditional and recover_marginals does? It seems like conditional is just recover_marginals with a better forward sampling algorithm for joint sampling the variables.

This is also assuming that both functions let us recover the conditional logps of the marginalised variables

ricardoV94:
> I don't understand the difference between what conditional and recover_marginals does? It seems like conditional is just recover_marginals with a better forward sampling algorithm for joint sampling the variables.
> 
> This is also assuming that both functions let us recover the conditional logps of the marginalised variables

conditional gives you back the conditional of the marginal model... which you can use to take draws or eval probabilities. recover_marginals is some syntactic sugar that calls `sample_posterior_predictive(conditional(marginal_model))`. Basically I elevated the logic of getting the conditional into a public model transformation.

Why? Because the logp case no longer makes sense to "automatically evaluate" for continuous variables. So the we give users a model from which they can extract/define the logp function and do whatever they want with it.

ricardoV94:
pymc-examples notebook update: https://github.com/pymc-devs/pymc-examples/pull/885

jessegrabowski (approved):
we talked about this together on a call, it's complex but im happy to merge it and see what happens.

ricardoV94:
I AM STILL GONNA CLEAN UP SOME OF THE SLOP but I like the core structure already

zaxtax:
I like the general approach. Will review tomorrow!

pymc_extras/model/marginal/distributions/core.py
```diff
@@ -0,0 +1,81 @@
...
+    fgraph.replace_all(tuple(zip(inner_inputs, inputs)), import_missing=True)
+    fgraph.replace_all(tuple(dummy_to_outer_replacements), import_missing=True)
+
+    [rv_support_point] = fgraph.outputs
+    return rv_support_point
+
+
+@singledispatch
+def _marginalized_conditional(op, node):
+    """Build the conditional distribution for a marginalized variable.
```
zaxtax: I think we need more detail here about what exactly is `marginalized_conditional`

pymc_extras/model/marginal/rewrites.py
```diff
@@ -0,0 +1,333 @@
...
+    updates are discovered here. The OpFromGraph constructor handles cloning.
+
+    Returns (inputs, outputs) where outputs = [marginalized_rv, *deps, *rng_updates].
+    """
+    subgraph_outputs, boundary = node.op.split_node_inputs(node)
+
+    n_rvs = 1 + node.op.n_dependent_rvs
+    rng_updates = {}
+    for rv in subgraph_outputs[:n_rvs]:
+        # TODO: Why not pass all rvs at once?
```
zaxtax: I have the same question as the TODO

pymc_extras/model/marginal/rewrites.py
```diff
@@ -0,0 +1,333 @@
...
+        for k, v in collect_default_updates([rv], inputs=boundary, must_be_shared=False).items():
+            rng_updates.setdefault(k, v)
+
+    outputs = subgraph_outputs + list(rng_updates.values())
+    return boundary, outputs
+
+
+@node_rewriter(tracks=[MarginalRV])
+def local_unmarginalize(fgraph, node):
```
zaxtax, line 155: what's the difference between local_unmarginalize and unmarginalize

pymc_extras/model/marginal/model.py
```diff
@@ -0,0 +1,706 @@
...
+    # Add the conditional as a new free RV
+    sample_graph.name = var_name
+    value = sample_graph.type()
+    value.name = var_name
+    fg.add_input(value)
+    conditional_free_rv = model_free_rv(sample_graph, value, None, *op.marginalized_dims)
+    fg.add_output(conditional_free_rv, reason="conditionalize")
+
+
+def recover(
```
zaxtax: Why does recover need to change its name. Does the function really do something radically different?

zaxtax (commented):
This looks really solid. This machinery though needs better documentation. Is there a plan to add a guide for how to add a distribution to the marginalize and conditional machinery?

pymc_extras/model/marginal/rewrites.py
ricardoV94: local_unrmarginalize is a pytensor graph rewrite, unmarginalize is a user function that takes model -> returns model, they work at different layers of abstraction. This one isn't user facing

pymc_extras/model/marginal/model.py
ricardoV94: Because the old one was too long? The import context `pme.marginal.recover` gives enough context imo? It's a stylistic preference only

pymc_extras/model/marginal/distributions/core.py
ricardoV94: Agree, and added. It's a bit too AI, but the main point is there

pymc_extras/model/marginal/rewrites.py
ricardoV94: great, it wasn't needed

pymc_extras/model/marginal/distributions/core.py
```diff
@@ -0,0 +1,95 @@
...
+    fgraph.replace_all(tuple(zip(inner_inputs, inputs)), import_missing=True)
+    fgraph.replace_all(tuple(dummy_to_outer_replacements), import_missing=True)
+
+    [rv_support_point] = fgraph.outputs
+    return rv_support_point
+
+
+@singledispatch
+def marginalized_conditional(op, node):
+    """Build the conditional distribution of a marginalized variable given its dependents.
```
ricardoV94, line 59: Docstring here @zaxtax

ricardoV94:
> This looks really solid. This machinery though needs better documentation. Is there a plan to add a guide for how to add a distribution to the marginalize and conditional machinery?

Let's iterate a bit more to see where we land? Kwarg based options already put a bit of a wrinkle. I would spend some cycles pushing functionality first

ricardoV94:
Dev docs preview at: https://pymcio--688.org.readthedocs.build/projects/extras/en/688/developer/extending_marginalization.html

zaxtax (approved):
Looks great!

PR #638, opened 2026-02-04
Add Generalized Pareto (GPD) and Extended GPD distributions

maresb:
> [!NOTE]
> **AI Assistance Disclosure:** This PR was developed with significant assistance from an AI coding assistant (Claude) via Cursor. The code has not yet been fully reviewed by the author.

## Summary

This PR adds two new probability distributions to PyMC-Extras:

1. **Generalized Pareto Distribution (GPD)** - A fundamental distribution in extreme value theory for modeling exceedances over a threshold
2. **Extended Generalized Pareto Distribution (ExtGPD)** - A transformation of GPD that provides smoother behavior in the lower tail, useful for modeling entire distributions without explicit threshold selection

### Features

- Full implementation with `logp`, `logcdf`, `icdf`, and `support_point` methods
- Random sampling via scipy's `genpareto`
- Comprehensive test coverage using PyMC's testing utilities
- The `icdf` method enables efficient inverse CDF sampling when used with `pm.Truncated`, avoiding rejection sampling failures

### Mathematical Background

The ExtGPD CDF is defined as:
$$G(x \mid \mu, \sigma, \xi, \kappa) = \left[H\left(\frac{x - \mu}{\sigma}\right)\right]^\kappa$$

where $H$ is the standard GPD CDF. The parameter $\kappa > 0$ controls lower tail behavior, and when $\kappa = 1$, ExtGPD reduces to standard GPD.

## Test plan

- [x] `pytest tests/distributions/test_continuous.py::TestGenParetoClass` - GPD tests
- [x] `pytest tests/distributions/test_continuous.py::TestExtGenParetoClass` - ExtGPD tests
- [x] `pytest tests/distributions/test_continuous.py::TestGenPareto` - GPD random sampling
- [x] `pytest tests/distributions/test_continuous.py::TestExtGenPareto` - ExtGPD random sampling

PR #689, opened 2026-05-31
Add GenPareto and ExtGenPareto (Generalized Pareto) distributions

maresb:
## Summary

Adds two distributions to `pymc_extras.distributions`:

- `GenPareto`: Generalized Pareto distribution, parametrized like SciPy's `genpareto` (`c = ξ`).
- `ExtGenPareto`: Naveau et al. Type-1 extended GPD, `F = H**κ`, where `H` is the GPD CDF.

Both include PyTensor kernels for `logp`, `logcdf`, `logccdf`, `icdf`, inverse-CDF RNGs, support points, and default transforms for latent variables.

## Implementation

The implementation is split by layer:

- `_pytensor_genpareto.py` / `_pytensor_extgenpareto.py`: PyTensor-only math, shaped to match `pytensor-distributions`.
- `_pymc_genpareto.py` / `_pymc_extgenpareto.py`: PyMC wrappers, `SymbolicRandomVariable`s, parameter checks, support points, and transforms.
- `continuous.py`: re-exports the public classes.

The wrappers validate `σ > 0` and `κ > 0` with `check_parameters`.

## Transform

The default transform is `y = logit(F(x))`: `F` maps the support onto `(0, 1)` and `logit` onto the real line, so the unconstrained variable is standard Logistic for every `ξ`. In testing with random `ξ` it produced far fewer divergences than interval-style or lower-bound transforms.

It is bijective wherever the quantile is representable. For very small `κ`, ExtGPD numerically collapses toward `μ`; in that case `backward` floors the readout just inside support, while the transformed density remains Logistic.

## Tests

Tests mirror the split:

- `test_generalized_pareto.py`: PyMC integration, dispatch, parameter checks, RNG shape/size, support points, and transform behavior.
- `test_pytensor_genpareto.py` / `test_pytensor_extgenpareto.py`: formula checks against SciPy/high-precision references, tail stability, boundary behavior, gradients, and functional `ppf` / `isf` / `rvs`.

Supersedes #638 and follows up earlier GPD work in #294.

### DIFF
--- a/pymc_extras/distributions/continuous.py
+++ b/pymc_extras/distributions/continuous.py
@@ -24,7 +24,9 @@ import pytensor.tensor as pt
 
 from pymc import ChiSquared, CustomDist
 from pymc.distributions import transforms
-from pymc.distributions.dist_math import check_parameters
+from pymc.distributions.dist_math import (
+    check_parameters,
+)
 from pymc.distributions.distribution import Continuous
 from pymc.distributions.shape_utils import rv_size_is_none
 from pymc.logprob.utils import CheckParameterValue
@@ -32,6 +34,9 @@ from pymc.pytensorf import floatX
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.variable import TensorVariable
 from scipy import stats
+
+from pymc_extras.distributions.pymc_extgenpareto import ExtGenPareto  # noqa: F401
+from pymc_extras.distributions.pymc_genpareto import GenPareto  # noqa: F401
 
 
 class GenExtremeRV(RandomVariable):
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -11,6 +11,7 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+
 import numpy as np
 import pymc as pm
 
--- a/docs/api/distributions.rst
+++ b/docs/api/distributions.rst
@@ -15,6 +15,8 @@ like regular PyMC distributions and can be used directly inside a model.
    GeneralizedPoisson
    BetaNegativeBinomial
    GenExtreme
+   GenPareto
+   ExtGenPareto
    R2D2M2CP
    Skellam
    histogram_approximation
--- a/pymc_extras/distributions/__init__.py
+++ b/pymc_extras/distributions/__init__.py
@@ -17,7 +17,13 @@
 Experimental probability distributions for stochastic nodes in PyMC.
 """
 
-from pymc_extras.distributions.continuous import Chi, GenExtreme, Maxwell
+from pymc_extras.distributions.continuous import (
+    Chi,
+    ExtGenPareto,
+    GenExtreme,
+    GenPareto,
+    Maxwell,
+)
 from pymc_extras.distributions.discrete import (
     BetaNegativeBinomial,
     GeneralizedPoisson,
@@ -33,7 +39,9 @@ __all__ = [
     "BetaNegativeBinomial",
     "Chi",
     "DiscreteMarkovChain",
+    "ExtGenPareto",
     "GenExtreme",
+    "GenPareto",
     "GeneralizedPoisson",
     "JointCategorical",
     "Maxwell",
--- a/pymc_extras/distributions/pymc_extgenpareto.py
+++ b/pymc_extras/distributions/pymc_extgenpareto.py
@@ -1,0 +1,345 @@
+#   Copyright 2022 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import pytensor.tensor as pt
+
+from pymc.distributions.dist_math import (
+    check_icdf_parameters,
+    check_icdf_value,
+    check_parameters,
+)
+from pymc.distributions.distribution import Continuous, SymbolicRandomVariable
+from pymc.distributions.shape_utils import implicit_size_from_params, rv_size_is_none
+from pymc.distributions.transforms import _default_transform
+from pymc.pytensorf import floatX, normalize_rng_param
+from pytensor.tensor.random.utils import normalize_size_param
+
+from pymc_extras.distributions import pytensor_extgenpareto as extgenpareto
+from pymc_extras.distributions import pytensor_genpareto as genpareto
+from pymc_extras.distributions.pymc_genpareto import _AbstractGPDLogitCDFTransform
+
+
+class ExtGenParetoRV(SymbolicRandomVariable):
+    name = "extgenpareto"
+    extended_signature = "[rng],[size],(),(),(),()->[rng],()"
+    _print_name = ("ExtGenPareto", "\\operatorname{ExtGenPareto}")
+
+    @classmethod
+    def rv_op(cls, mu, sigma, xi, kappa, *, size=None, rng=None):
+        mu = pt.as_tensor(mu)
+        sigma = pt.as_tensor(sigma)
+        xi = pt.as_tensor(xi)
+        kappa = pt.as_tensor(kappa)
+        rng = normalize_rng_param(rng)
+        size = normalize_size_param(size)
+        if rv_size_is_none(size):
+            size = implicit_size_from_params(mu, sigma, xi, kappa, ndims_params=cls.ndims_params)
+        next_rng, u = rng.uniform(size=size)
+        # Carrier draw u = F; excess = -log(1 - u ** (1/kappa)), via log1mexp so
+        # small-kappa draws do not collapse to the lower endpoint (1 - u**.. -> 1).
+        excess = extgenpareto._ext_gpd_excess_from_log_prob(pt.log(u), kappa)
+        draws = genpareto._gpd_quantile_from_excess(excess, mu, sigma, xi)
+        return cls(inputs=[rng, size, mu, sigma, xi, kappa], outputs=[next_rng, draws])(
+            rng, size, mu, sigma, xi, kappa
+        )
+
+
+class ExtGenPareto(Continuous):
+    r"""Extended Generalized Pareto distribution.
+
+    The extended GPD (Type 1) of Naveau et al. (2016) [1]_ is equivalent to
+    the EGP3 model of Papastathopoulos and Tawn (2013) [2]_. It transforms
+    the GPD CDF :math:`G` through the carrier :math:`v \mapsto v^\kappa`:
+
+    .. math::
+
+       F(x \mid \mu, \sigma, \xi, \kappa) =
+           \left[G\!\left(x \mid \mu, \sigma, \xi\right)\right]^{\kappa}.
+
+    The GPD parameters :math:`\mu, \sigma, \xi` describe the tail (large :math:`x`),
+    as in the GPD. :math:`\kappa > 0` is an additional shape parameter for the body
+    (small :math:`x`); :math:`\kappa = 1` recovers the GPD. The extra parameter
+    gives more flexibility to model the body while leaving the GPD tail unchanged.
+    Typical applications include threshold exceedances [2]_ and modeling rainfall
+    intensities [1]_.
+
+    .. plot::
+        :context: close-figs
+        :include-source: False
+
+        import matplotlib.pyplot as plt
+        import numpy as np
+        import scipy.stats as st
+        import arviz as az
+        plt.style.use('arviz-darkgrid')
+        x = np.linspace(0.02, 4, 200)  # above 0: for kappa < 1 the density diverges at x = 0
+        xi = 0.2
+        for kappa in [0.8, 1.0, 1.2, 2.0, 3.0]:
+            G = st.genpareto.cdf(x, c=xi)
+            g = st.genpareto.pdf(x, c=xi)
+            scaled_pdf = np.power(G, kappa - 1) * g  # ExtGPD density divided by kappa.
+            plt.plot(x, scaled_pdf, label=rf'$\kappa$ = {kappa}')
+        plt.xlabel('x', fontsize=12)
+        plt.ylabel('tail-scaled density', fontsize=12)
+        plt.ylim(0, 1.5)
+        plt.legend(loc='upper right', title=rf'$\mu = 0,\ \sigma = 1,\ \xi = {xi}$')
+        plt.show()
+
+    ========  =========================================================================
+    Support   :math:`\begin{cases} x \geq \mu, & \xi \geq 0 \\ \mu \leq x < \mu - \sigma/\xi, & \xi < 0 \end{cases}`
+    Mean      :math:`\begin{cases} \mu + \dfrac{\sigma}{\xi}\left[\kappa B(\kappa, 1 - \xi) - 1\right], & \xi < 1,\ \xi \neq 0 \\ \mu + \sigma\left(\psi(\kappa + 1) + \gamma\right), & \xi = 0 \\ \infty, & \xi \geq 1 \end{cases}`
+    Variance  :math:`\begin{cases} \left(\dfrac{\sigma}{\xi}\right)^2 \left[\kappa B(\kappa, 1 - 2\xi) - \kappa^2 B(\kappa, 1 - \xi)^2\right], & \xi < 1/2,\ \xi \neq 0 \\ \sigma^2\left(\pi^2/6 - \psi'(\kappa + 1)\right), & \xi = 0 \\ \infty, & \xi \geq 1/2 \end{cases}`
+    Median    :math:`\begin{cases} \mu + \dfrac{\sigma}{\xi}\left[\left(1 - 2^{-1/\kappa}\right)^{-\xi} - 1\right], & \xi \neq 0 \\ \mu - \sigma\ln\!\left(1 - 2^{-1/\kappa}\right), & \xi = 0 \end{cases}`
+    Mode      :math:`\begin{cases} \mu + \dfrac{\sigma}{\xi}\left(T^{-\xi} - 1\right), & \kappa > 1,\ \xi \geq -1,\ \xi \neq 0 \\ \mu + \sigma\ln\kappa, & \kappa > 1,\ \xi = 0 \\ \mu, & \kappa \leq 1,\ \xi \geq -1 \\ \mu - \sigma/\xi, & \kappa \geq 1,\ \xi < -1 \end{cases}` where :math:`T = \dfrac{1 + \xi}{\kappa + \xi}`
+    ========  =========================================================================
+
+    Here :math:`B(a, b) = \Gamma(a)\Gamma(b)/\Gamma(a + b)` is the Beta function,
+    :math:`\psi` the digamma, :math:`\psi'` the trigamma, and :math:`\gamma` the
+    Euler--Mascheroni constant.
+
+    Parameters
+    ----------
+    mu : tensor_like of float
+        Location parameter (the threshold).
+    sigma : tensor_like of float
+        Scale parameter (sigma > 0).
+    xi : tensor_like of float
+        Tail shape parameter (large :math:`x`), as in :class:`GenPareto`.
+    kappa : tensor_like of float
+        Body shape parameter (small :math:`x`, ``kappa > 0``). ``kappa = 1``
+        reduces the distribution to :class:`GenPareto`.
+
+    References
+    ----------
+    .. [1] Naveau, P., Huser, R., Ribereau, P., & Hannart, A. (2016). Modeling
+        jointly low, moderate, and heavy rainfall intensities without a threshold
+        selection. Water Resources Research, 52(4), 2753-2769.
+        https://doi.org/10.1002/2015WR018552
+    .. [2] Papastathopoulos, I., & Tawn, J. A. (2013). Extended generalised
+        Pareto models for tail estimation. Journal of Statistical Planning and
+        Inference, 143(1), 131-143.
+        https://arxiv.org/abs/1111.6899
+
+    Notes
+    -----
+    All of the Notes on :class:`GenPareto` apply here too. The :math:`r`-th moment
+    is finite iff :math:`\xi < 1/r`, as for the GPD.
+
+    For positive :math:`\xi`, tail-only data can make :math:`\sigma` and
+    :math:`\kappa` hard to identify separately. In the small-:math:`\sigma`,
+    large-:math:`\kappa` regime, observations above the threshold are informed
+    mainly by a combined tail-survival scale rather than by the two parameters
+    individually, making the fit sensitive to very broad :math:`\sigma` priors.
+
+    .. plot::
+        :context: close-figs
+        :include-source: False
+
+        import matplotlib.pyplot as plt
+        import numpy as np
+        import arviz as az
+        plt.style.use("arviz-darkgrid")
+
+        def gpd_log_survival(x, sigma, xi):
+            return -(1.0 / xi) * np.log1p(xi * x / sigma)
+
+        def gpd_log_pdf(x, sigma, xi):
+            return -np.log(sigma) - (1.0 / xi + 1.0) * np.log1p(xi * x / sigma)
+
+        def extgpd_log_pdf(x, sigma, xi, kappa):
+            log_s = gpd_log_survival(x, sigma, xi)
+            log_g = np.log1p(-np.exp(log_s))
+            return np.log(kappa) + (kappa - 1.0) * log_g + gpd_log_pdf(x, sigma, xi)
+
+        def boundary_log_pdf(x, xi, lam):
+            alpha = 1.0 / xi
+            return (
+                np.log(lam)
+                - (alpha + 1.0) * np.log(xi)
+                - (alpha + 1.0) * np.log(x)
+                - lam * (xi * x) ** (-alpha)
+            )
+
+        x = np.linspace(1e-3, 8.0, 1200)
+        xi = 1.5
+        lam = 2.0
+        alpha = 1.0 / xi
+        sigmas = [0.7, 0.2, 0.05, 0.01]
+
+        plt.plot(x, np.exp(boundary_log_pdf(x, xi, lam)), "k--", lw=2.5, label="limit")
+        colors = plt.cm.plasma(np.linspace(0.08, 0.82, len(sigmas)))
+        for sigma, color in zip(sigmas, colors):
+            kappa = lam / sigma**alpha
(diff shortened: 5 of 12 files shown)


pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1000 @@ def dist(cls, a, **kwargs):
...
+
+def _uniform_draw(size, rng):
+    """``(next_rng, draw)`` from a uniform RV (the pytensor random idiom)."""
+    return uniform(size=size, rng=rng, return_next_rng=True)
+
+
+class GenParetoRV(SymbolicRandomVariable):
+    name = "genpareto"
+    extended_signature = "[rng],[size],(),(),()->[rng],()"
+    _print_name = ("GenPareto", "\\operatorname{GenPareto}")
```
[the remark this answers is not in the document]
maresb: The signature is correct as written. `[rng],[size],(),(),()` lists three scalar `()` inputs -- `mu`, `sigma`, `xi` -- matching `rv_op`'s three parameters; the leading `[rng]` / `[size]` are the rng/size tokens, not scalar params. Verified shape inference: a batched `mu=[0, 10, 100]` draw returns shape `(3,)`.

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1000 @@ def dist(cls, a, **kwargs):
...
+        draws = _gpd_quantile_from_excess(-pt.log(v), mu, sigma, xi)
+        return cls(inputs=[rng, size, mu, sigma, xi], outputs=[next_rng, draws])(
+            rng, size, mu, sigma, xi
+        )
+
+
+class ExtGenParetoRV(SymbolicRandomVariable):
+    name = "extgenpareto"
+    extended_signature = "[rng],[size],(),(),(),()->[rng],()"
+    _print_name = ("ExtGenPareto", "\\operatorname{ExtGenPareto}")
```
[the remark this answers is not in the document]
maresb: Correct here too: `[rng],[size],(),(),(),()` lists four scalar `()` inputs -- `mu`, `sigma`, `xi`, `kappa` -- matching `rv_op`'s four parameters. Verified with batched vector `mu`/`kappa` draws (shape `(3,)`).

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1000 @@ def dist(cls, a, **kwargs):
...
+    """``log1p(u) / u`` continued to its removable ``u = 0`` limit (= 1), C1-smooth.
+
+    ``safe_u`` keeps the unused exact branch from evaluating ``0/0`` so the
+    gradient graph stays NaN-free.
+    """
+    cutoff = _series_cutoff(u.dtype)
+    use_series = pt.lt(pt.abs(u), cutoff)
+    series = 1.0 - u / 2.0 + u**2 / 3.0 - u**3 / 4.0 + u**4 / 5.0
+    safe_u = pt.switch(use_series, np.asarray(1.0, dtype=u.dtype), u)
+    return pt.switch(use_series, series, pt.log1p(safe_u) / safe_u)
```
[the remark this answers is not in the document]
maresb: Verified this does not upcast under current PyTensor: float32 in -> float32 out from `_log1p_div` (and through the full logp and RNG draws). Under the default `numpy+floatX` cast policy the Python-float literals are weakly typed, so a float32 combined with a Python float stays float32; the only bare constant is the leading `1.0`, and `1.0 - (float32) -> float32`. No float64 leaks into the graph, so no change is needed (the dtype-aware `_series_cutoff` already handles the float32-vs-float64 difference that matters here).

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1000 @@ def dist(cls, a, **kwargs):
...
+
+
+def _expm1_div(u: TensorVariable) -> TensorVariable:
+    """``expm1(u) / u`` continued to its removable ``u = 0`` limit (= 1), C1-smooth."""
+    cutoff = _series_cutoff(u.dtype)
+    use_series = pt.lt(pt.abs(u), cutoff)
+    series = 1.0 + u / 2.0 + u**2 / 6.0 + u**3 / 24.0 + u**4 / 120.0
+    safe_u = pt.switch(use_series, np.asarray(1.0, dtype=u.dtype), u)
+    return pt.switch(use_series, series, pt.expm1(safe_u) / safe_u)
```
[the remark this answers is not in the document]
maresb: Same as `_log1p_div` -- verified float32 in -> float32 out (helper, logp, and draws); the weakly-typed literals do not promote under the default cast policy.

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1000 @@ def dist(cls, a, **kwargs):
...
+    The repair is restricted to the exact ``{0} x {+-inf}`` cases so that a
+    genuine ``nan`` in ``a`` or ``b`` (e.g. ``xi = nan`` from bad input) still
+    propagates instead of being silently turned into the ``xi = 0`` branch.
+    """
+    prod = a * b
+    zero_times_inf = pt.or_(
+        pt.and_(pt.eq(a, 0), pt.isinf(b)),
+        pt.and_(pt.eq(b, 0), pt.isinf(a)),
+    )
+    return pt.switch(zero_times_inf, 0.0, prod)
```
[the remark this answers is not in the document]
maresb: Verified `_safe_mul(float32, float32) -> float32`: `pt.switch(cond, 0.0, prod)` keeps `prod`'s float32 dtype because the Python-float `0.0` is weakly typed. Confirmed no promotion through the GPD logp/draw graphs end-to-end.

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1000 @@ def dist(cls, a, **kwargs):
...
+        #   m -> t + log(kappa) - log(_log1p_div(-exp(-t))),
+        # built without forming exp(-y) before a log, finite until t overflows. The
+        # two branches agree to machine precision at the switch, and both run on
+        # clamped inputs so the discarded one (and its gradient) stays finite.
+        t = pt.softplus(value)
+        log_F = -pt.softplus(-pt.minimum(value, 700.0))
+        m_bulk = _ext_gpd_excess_from_log_prob(log_F, kappa)
+        s = pt.exp(-pt.maximum(t, 700.0))
+        m_tail = t + pt.log(kappa) - pt.log(_log1p_div(-s))
+        return pt.switch(value < 700.0, m_bulk, m_tail)
```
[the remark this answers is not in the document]
maresb: Good catch on the spirit, though the real issue is not dtype promotion (the literal stays float32 under weak promotion) -- it is that `700.0` is a *float64-specific* threshold. Under float32, `exp(-y)` underflows around y ~ 88, so y in ~[88, 700) was routed through the bulk branch (where `log F_ext` had already underflowed) and the excess blew up to +inf well inside the reachable range. Fixed in 5b2285f by deriving the crossover from `finfo(value.dtype)` -- ~700 for float64 (bit-identical) and ~80 for float32 -- with a regression test for the float32 upper tail.

tests/distributions/test_continuous.py
```diff
@@ -187,3 +210,1059 @@ def test_logcdf(self):
...
+        # integrates to 1 over the unconstrained line.
+        from scipy.integrate import trapezoid
+
+        with pm.Model() as model:
+            builder("x", **dist_kwargs)
+        y = model.value_vars[0]
+        logp = pytensor.function([y], model.logp(sum=True))
+        ys = np.linspace(-30, 30, 30001)
+        density = np.exp(np.array([float(logp(yi)) for yi in ys]))
+        np.testing.assert_allclose(trapezoid(density, ys), 1.0, atol=1e-3)
```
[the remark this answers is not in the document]
maresb: Addressed in 6d97d94. The integrand is the smooth, exactly-Logistic transformed density, so the quadrature error is dominated by the +/-30 tail truncation (~2e-13), not the grid spacing -- 2001 points already integrate to ~1e-13, far inside the 1e-3 tolerance. Reduced 30001 -> 2001 (5x fewer evals); compile time dominates either way, so this is simpler than vectorizing.

tests/distributions/test_continuous.py
```diff
@@ -187,3 +210,1059 @@ def test_logcdf(self):
...
+        with pm.Model() as model:
+            GenPareto("x", mu=mu, sigma=sigma, xi=xi)
+            idata = pm.sample(
+                200,
+                tune=300,
+                chains=2,
+                progressbar=False,
+                random_seed=1,
+                compute_convergence_checks=False,
+            )
```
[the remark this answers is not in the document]
maresb: Addressed in 5b2285f: cut draws/tune 200/300 -> 100/200 and forced `cores=1` (the run uses a fixed seed, so it stays deterministic). The clean PIT geometry still keeps all samples in support with zero divergences across the three cases.

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,809 @@ def dist(cls, a, **kwargs):
...
+
+def _log1p_div(u: TensorVariable) -> TensorVariable:
+    """``log1p(u) / u`` continued to its removable ``u = 0`` limit (= 1), C1-smooth.
+
+    ``safe_u`` keeps the unused exact branch from evaluating ``0/0`` so the
+    gradient graph stays NaN-free.
+    """
+    cutoff = _series_cutoff(u.dtype)
+    use_series = pt.lt(pt.abs(u), cutoff)
+    series = 1.0 - u / 2.0 + u**2 / 3.0 - u**3 / 4.0 + u**4 / 5.0
```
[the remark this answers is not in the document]
maresb: Re-raised in the latest review round at line 409 and addressed there: verified float32 is preserved end-to-end (helper, logp, draws), no upcast under the default cast policy.

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,809 @@ def dist(cls, a, **kwargs):
...
+    safe_u = pt.switch(use_series, np.asarray(1.0, dtype=u.dtype), u)
+    return pt.switch(use_series, series, pt.log1p(safe_u) / safe_u)
+
+
+def _expm1_div(u: TensorVariable) -> TensorVariable:
+    """``expm1(u) / u`` continued to its removable ``u = 0`` limit (= 1), C1-smooth."""
+    cutoff = _series_cutoff(u.dtype)
+    use_series = pt.lt(pt.abs(u), cutoff)
+    series = 1.0 + u / 2.0 + u**2 / 6.0 + u**3 / 24.0 + u**4 / 120.0
```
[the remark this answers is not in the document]
maresb: Re-raised at line 418 in the latest review -- verified float32 is preserved; see the reply there.

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,809 @@ def dist(cls, a, **kwargs):
...
+    The repair is restricted to the exact ``{0} x {+-inf}`` cases so that a
+    genuine ``nan`` in ``a`` or ``b`` (e.g. ``xi = nan`` from bad input) still
+    propagates instead of being silently turned into the ``xi = 0`` branch.
+    """
+    prod = a * b
+    zero_times_inf = pt.or_(
+        pt.and_(pt.eq(a, 0), pt.isinf(b)),
+        pt.and_(pt.eq(b, 0), pt.isinf(a)),
+    )
+    return pt.switch(zero_times_inf, 0.0, prod)
```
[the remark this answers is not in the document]
maresb: Re-raised at line 446 in the latest review -- verified `_safe_mul` keeps float32; see the reply there.

tests/distributions/test_continuous.py
```diff
@@ -187,3 +207,757 @@ def test_logcdf(self):
...
+    def test_transformed_density_integrates_to_one(self, dist_kwargs, builder):
+        # The transform's log-Jacobian must be correct: exp(transformed logp)
+        # integrates to 1 over the unconstrained line.
+        from scipy.integrate import trapezoid
+
+        with pm.Model() as model:
+            builder("x", **dist_kwargs)
+        y = model.value_vars[0]
+        logp = pytensor.function([y], model.logp(sum=True))
+        ys = np.linspace(-30, 30, 30001)
```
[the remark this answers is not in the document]
maresb: Re-raised at line 891 in the latest review and addressed in 6d97d94 (reduced the integration grid 30001 -> 2001, error is +/-30-truncation-dominated at ~1e-13); see the reply there.

maresb:
@ricardoV94, this is a monster, but it's been battle-tested. I've also tried to ensure that all your feedback from previous PRs has been taken into account.

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1033 @@ def dist(cls, a, **kwargs):
             class_name="Maxwell",
             **kwargs,
         )
+
+
+# ===========================================================================
```
ricardoV94: Please no dangling comments in global space.  Put them inside class/functions where they make sense.

Assuming they are even needed. They don't seem to be

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1033 @@ def dist(cls, a, **kwargs):
...
+# All three reduce to the exponential law as xi -> 0 (m -> z).
+
+
+def _safe_mul(a, b):
+    """``a * b``, repairing ONLY the indeterminate ``0 * inf`` to ``0``.
+
+    Only ``xi * z`` needs this: at ``xi = 0`` with an infinite observation the
+    product is mathematically ``0`` (the exponential GPD carries no shape term),
+    but ``0.0 * inf`` is ``nan`` under IEEE, and a ``nan`` in the unused branch of
+    a ``switch`` can still leak (the backend may lower it to ``cond*a + ...``).
```
ricardoV94: What backend does that?

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1033 @@ def dist(cls, a, **kwargs):
...
+def _in_gpd_support(z, t):
+    """Boolean mask of the GPD support: z >= 0 and (for xi < 0) z <= -1/xi.
+
+    ``t = xi * z``; the upper edge is ``s = 1 + t > 0`` (the same ``t`` the density
+    is built from, so the mask and the value never disagree at the wall).
+    """
+    return pt.and_(z >= 0, 1 + t > 0)
+
+
+def _propagate_nonfinite_shape(result, xi, kappa=None):
```
ricardoV94: is this really needed? We don't worry about say inf sigma in normal likelihood and things like that

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1033 @@ def dist(cls, a, **kwargs):
...
+
+
+# Extended Generalized Pareto core. Naveau et al. (2016) extended GPD with
+# carrier G(v) = v ** kappa: the CDF is F = H ** kappa with H the GPD CDF, so
+# kappa > 0 reshapes the lower tail while xi keeps controlling the upper tail.
+# kappa = 1 recovers the plain GPD.
+
+
+def ext_gen_pareto_logp(value, mu, sigma, xi, kappa):
+    """Pure-PyTensor extended-GPD log-density; out-of-support values map to ``-inf``."""
```
ricardoV94: pure PyTensor is such a useless statement. Really can we trim 90% of these comments and docstrings. Just visual noise

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1033 @@ def dist(cls, a, **kwargs):
...
+
+    The lower bound depends on ``xmax``, so the prior support encodes the hard
+    "data fall inside the support" requirement -- the standard way to handle a
+    distribution whose support edge is an unknown the data constrain (cf.
+    ``theta >= max(data)`` for ``Uniform(0, theta)``). This addresses *sampling
+    stability*; it is distinct from the deeper float64 precision limit right at the
+    wall noted above (reached only when the wall is pinned within ``~1e-13`` of the
+    data).
+
+    The threshold ``mu`` is held fixed -- the standard peaks-over-threshold setup,
```
ricardoV94: Can we make this paragraph sound less like llm speech? At least remove the load bearing. I believe there are exactly 0 occurrences of this term in the first 2 decades of PyMC history :)

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1033 @@ def dist(cls, a, **kwargs):
...
+        median = _gpd_quantile_from_excess(excess, mu, sigma, xi)
+        gpd_median = _gpd_quantile_from_excess(np.log(2.0), mu, sigma, xi)
+        point = pt.switch(pt.le(median, mu), gpd_median, median)
+        if not rv_size_is_none(size):
+            point = pt.full(size, point)
+        return point
+
+
+class _GPDProbabilityIntegralTransform(Transform):
+    """Default transform for the GPD family: ``y = logit(F(x))``.
```
ricardoV94: interesting

pymc_extras/distributions/continuous.py
```diff
@@ -350,3 +358,1033 @@ def dist(cls, a, **kwargs):
...
+        # logp_x would let that NaN cancel against the framework's + logp(backward) and
+        # resurface as a *finite* Logistic, silently accepting an invalid parameter as a
+        # valid latent. Route it through a switch whose other branch is a bare constant
+        # (no logp_x to cancel), so the transformed logp stays NaN -- matching pm.Gamma
+        # / pm.StudentT, whose transformed latent logp is NaN (caught loudly at
+        # initialization) for a non-finite shape rather than a spurious finite density.
+        return pt.switch(pt.isnan(logp_x), np.nan, jac)
+
+
+class _GenParetoPIT(_GPDProbabilityIntegralTransform):
```
ricardoV94: What are these classes about/for?

tests/distributions/test_continuous.py
```diff
@@ -11,35 +11,62 @@
...
+    gen_pareto_logp,
+)
 
 pytestmark = pytest.mark.filterwarnings(
     "ignore:Numba will use object mode to run Generalized Extreme Value:UserWarning"
 )
 
+# The Generalized Pareto family legitimately evaluates to +/-inf at the
+# (measure-zero) support boundary and at probabilities 0 / 1; NumPy flags those
+# as divide-by-zero / invalid / overflow during ``.eval()``. They are the correct
```
ricardoV94: we suppress these warnings are you sure we are the ones firing them, not the test? if so suppress in the tests with np.errstate context

tests/distributions/test_continuous.py
```diff
@@ -187,3 +214,1072 @@ def test_logcdf(self):
...
+            )
+        # The clean PIT geometry should give zero divergences for all three at once.
+        assert int(idata.sample_stats.diverging.values.sum()) == 0
+        for i, (xi, mu, sigma) in enumerate(geometries):
+            xs = idata.posterior[f"x{i}"].values
+            assert np.all(xs >= mu - 1e-9), f"x{i} (xi={xi}) below mu"
+            if xi < 0:
+                assert np.all(xs <= mu - sigma / xi + 1e-9), f"x{i} (xi={xi}) past wall"
+
+    def test_observed_is_unaffected(self):
```
ricardoV94: This test is stupid

ricardoV94 (commented):
@maresb content looks great, delivery leaves a lot to be desired.

Can we:
1) Move functionality to new file to keep it organized (can still allow importing from distributions/continuous)
2) Aggressivepy pair down comment and docstring verbosity. When everything is emphasized ad nauseum nothing is. If a human or llm can infer what the comments are saying from the code we don't need it. Keep what's actually tricky/hard to recover.
3. Measure the total CI time incurred by the new tests? It's hard to gauge locally because of caching/different backends/GA being single core by default. Gotta assess if reasonable.
4. Some guards seem overly defensive and beyond what we do elsewhere or have felt the need for. Specially the fear of masking infinite/nan values. May be fine or may ne too much. Need your own opinion here.
5. Whenever I bash on style I'm bashing on the tool not you. All grievances are my own problem.

ricardoV94:
> PyMC's own conventions (e.g. a non-finite shape gives NaN like pm.Gamma, a non-finite scale gives -inf like pm.Normal).

There aren't so strict conventions. Parameter bounds are usually wrapped in a check_parameter_values that can be ignored (`Model(check_bounds=False)`) or converted to a ninf switch when compile_logp is called.

maresb:
Thanks a lot @ricardoV94 for the speedy and savage review. I really appreciate the time you put in to read over this, and of course I don't take the criticism personally since I didn't really "write" this stuff myself. And probably I should have put in more effort before asking for review, but I had to stop somewhere. (I did put in a lot of time doing fairly rigorous correctness testing though; it's probably not perfect but I think it's close.)

This is probably overengineered, but I was going for correctness first. I actually feel like you're much better qualified to give an opinion on the various nan/inf checks than I am. I'm curious if the checks will get in the way of rewrites.

Regarding the tests, I did some extensive profiling, and I don't think there are likely many obvious wins beyond removing certain tests altogether. I'll reproduce the profiling for you, but from memory, no individual test went beyond a few seconds, and the time to run `test_continuous` raised from ~25 sec. to ~50 sec. I'll split out both the functions and the tests into separate files to make this easier to see.

Thanks so much for your time and patience, I really appreciate it!!!

maresb:
While some of the complicated numerics were motivated purely by aggressive parameter scans, a very large proportion were actually necessary to preserve stability that was leading to real divergences during sampling.

maresb:
**CI test-cost breakdown**

The GPD additions are **91 tests**, all in the `tests/distributions` shard (which runs 240 passed / 8 skipped overall). The same 91 run under both backends:

- ubuntu / cvm: shard 167s; 31 of the shard's 50 slowest tests are ours (~40s)
- ubuntu / numba: shard 331s; the same 31 are ~65s

They split by layer:

- **PyMC integration** (`test_generalized_pareto.py`, 52 tests): these build full `pm.Model` logp / RV / transform graphs, so they are the compile-heavy ones and account for most of the slow end.
- **PyTensor reference** (`test_pytensor_genpareto.py` / `test_pytensor_extgenpareto.py`, 39 tests): lighter; they cover the PyTensor-only math layer and are written to migrate to `pytensor-distributions` along with that code.

By complete (`--durations=0`) timing the split is roughly 65% PyMC / 35% PyTensor. So the heavier integration tests stay here, while the ~40% by count that cover the pure-math layer leave when it upstreams.

pymc_extras/distributions/continuous.py
maresb: Done, removed. The dangling top-of-file comment is gone, along with the rest of the architecture narration.

maresb:
Thanks Ricardo, this was the right read ("content's there, delivery needs work").

I addressed the main delivery issues:

- Split the implementation out of `continuous.py`: the PyTensor kernels now live in `_pytensor_genpareto.py` / `_pytensor_extgenpareto.py`, and the PyMC wrappers live in `_pymc_genpareto.py` / `_pymc_extgenpareto.py`. `continuous.py` only re-exports the public classes.
- Removed the architecture-narration comments and most of the defensive implementation commentary. The remaining comments are mostly around numerical branches and transform behavior.
- Removed the speculative/non-useful guards (`_safe_mul`, non-finite-shape propagation, and later dead transform special cases) after checking they were not needed.
- Split and pruned the tests so the wrapper tests cover PyMC integration and the PyTensor files own the reference/math checks.

One tradeoff to note: the public distribution docstrings are still substantial. I kept the modeling guidance because these distributions are easy to misuse in PyMC, but the implementation comments are much leaner now. The remaining unusual implementation choice is the PIT transform; I called out the caveat in the PR body and kept direct regression coverage for it.

I have replied in each thread below and rewritten the PR description. Re-requesting your review when you have a chance.

pymc_extras/distributions/continuous.py
maresb: Good question, and the honest answer is none on the reachable path, so I removed the guard. The `0*inf` it defended is already handled by the `z == inf -> -inf` branch downstream; with the plain expression in its place, the boundary and `z -> inf` results (value and gradient) are unchanged.

pymc_extras/distributions/continuous.py
maresb: Agreed, it is not needed. Removed. The wrappers state the actual contract instead (`sigma > 0` and `kappa > 0` via `check_parameters`), and otherwise behave like the other likelihoods you mention.

pymc_extras/distributions/continuous.py
maresb: Agreed. Dropped the "pure PyTensor" meta-comments and trimmed the implementation comments throughout. I did keep the two public class docstrings substantial on purpose (the standard math/plot/params/refs plus a modeling-recommendations block for the `xi < 0` support trap, which is easy to misuse), but the in-code commentary is much leaner now.

pymc_extras/distributions/continuous.py
maresb: Done. Removed "load-bearing" and reworded the paragraph to drop the LLM-ish phrasing.

pymc_extras/distributions/continuous.py
maresb: Glad it caught your eye. The logit-CDF (PIT) default transform is based on measured sampling behavior, not just preference: in runs with random `xi` it produced far fewer divergences than an interval-style or lower-bound transform. The transformed variable has Logistic density by construction (`F` maps the support onto (0, 1), `logit` onto the real line). The one caveat, now in the PR description, is that for tiny `kappa` the ExtGPD collapses toward `mu` in finite precision, where `backward` floors the readout into the support.

pymc_extras/distributions/continuous.py
maresb: They are the default transforms for `GenPareto` / `ExtGenPareto`: `y = logit(F(x))`, mapping the constrained support to the unconstrained real line so the variables can be used as latents. Each subclass now has a one-line docstring saying exactly that.

tests/distributions/test_continuous.py
maresb: Good catch. I checked, and nothing under test actually raises those `FloatingPointError` warnings, so the filter was dead. Removed it.

tests/distributions/test_continuous.py
maresb: Agreed. Removed.

pymc_extras/distributions/_pymc_extgenpareto.py
```diff
@@ -0,0 +1,346 @@
...
+    def rv_op(cls, mu, sigma, xi, kappa, *, size=None, rng=None):
+        mu = pt.as_tensor(mu)
+        sigma = pt.as_tensor(sigma)
+        xi = pt.as_tensor(xi)
+        kappa = pt.as_tensor(kappa)
+        rng = normalize_rng_param(rng)
+        size = normalize_size_param(size)
+        if rv_size_is_none(size):
+            size = implicit_size_from_params(mu, sigma, xi, kappa, ndims_params=cls.ndims_params)
+        next_rng, u = uniform(size=size, rng=rng, return_next_rng=True)
```
ricardoV94:
```suggestion
        next_rng, u = rng.uniform(size=size)
```

ricardoV94 (approved):
@maresb this is immense work and looks great.

Can I just ask you remove the leading underscore in the file names, we never used that AFAICT.

ricardoV94:
(needs a rebase also)

maresb:
Squashed, updating to main now...

maresb:
Addressing feedback...

ricardoV94 (approved):

ricardoV94:
squash and merge or you want to change the git history? we shouldn't have a commit "remove underscore in file names" that were introduced in the same PR

maresb:
Thanks @ricardoV94, done! Will merge once green if you don't get it first.

PR #691, opened 2026-06-10
Deprecate save_kalman_filter_outputs_in_idata and always register data

jessegrabowski:
Fetch filter and smoother outputs after sampling with `sample_filter_outputs(idata)` instead of the `save_kalman_filter_outputs_in_idata` build flag, now deprecated. Data is always registered as a `pm.Data`, so a model built without coords picks up `time` and `observed_state` coords. Subclasses get `_register_additional_statespace_variables` as a place to register their own nodes rather than overriding `build_statespace_graph`.

Idempotency is out of scope here — a second build still raises on the duplicate `data` name. Making it work needs the fit-state off the model first, which is separate work.

### DIFF
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -941,7 +941,6 @@ class PyMCStateSpace:
     def build_statespace_graph(
         self,
         data: np.ndarray | pd.DataFrame | pt.TensorVariable,
-        register_data: bool = True,
         missing_fill_value: float | None = None,
         cov_jitter: float | None = None,
         mvn_method: Literal["cholesky", "eigh", "svd"] = "svd",
@@ -958,10 +957,6 @@ class PyMCStateSpace:
         data : Union[np.ndarray, pd.DataFrame, pt.TensorVariable]
             The observed data used to fit the state space model. It can be a NumPy array, a Pandas DataFrame,
             or a Pytensor tensor variable.
-
-        register_data : bool, optional, default=True
-            If True, the observed data will be registered with PyMC as a pm.Data variable. In addition,
-            a "time" dim will be created an added to the model's coords.
 
         mode : Optional[str], optional, default=None
             The Pytensor mode used for the computation graph construction. If None, the default mode will be used.
@@ -1043,7 +1038,6 @@ class PyMCStateSpace:
             data,
             n_obs=self.ssm.k_endog,
             obs_coords=obs_coords,
-            register_data=register_data,
             missing_fill_value=self.missing_fill_value,
         )
 
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -6,7 +6,6 @@ from typing import Any, Literal
 
 import numpy as np
 import pandas as pd
-import pymc as pm
 import pytensor
 import pytensor.tensor as pt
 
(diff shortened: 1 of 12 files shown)


jessegrabowski:
@ricardoV94 if you have a sec could you take a look at this? I kind of hate my hack fix because I think it points to a deeper structural issue. The thing i'm trying to solve is the following. I have a notebook where I build up a statespace + pymc model incrementally, as in:

```
import pymc as pm
import pymc_extras.statespace.structural as st

ss_mod = st.LevelTrend(name="trend", order=2, innovations_order=1).build(verbose=False)

with pm.Model(coords=ss_mod.coords) as pymc_mod:
    pm.Normal("initial_trend", dims=["state_trend"])
    pm.Deterministic("P0", pt.eye(2, dtype=floatX) * 1e6, dims=["state", "state_aux"])
    pm.Exponential("sigma_trend", 1, dims=["shock_trend"])
```

Then in a separate cell, I attach the data (after doing prior predictive or whatever)

```py
import pandas as pd

n=40
start="2020-01-01"
data = pd.DataFrame(
        np.arange(n, dtype=floatX)[:, None] + 1.0,
        columns=["data"],
        index=pd.date_range(start, periods=n, freq="D"),
    )

with pm_mod:
    ss_mod.build_statespace_graph(data)
```

Currently, if I call this cell again, it will error because the data is already registered in the pymc model. The bigger problem is that I think the PyMCStatespace class is just too big, and it's falling into the same trap that the ADVI module in PyMC fell into: state everywhere, storing way too much information in the class. Your commentary would be welcome.

pymc_extras/statespace/core/statespace.py
```diff
@@ -1092,6 +1091,13 @@ def build_statespace_graph(
             self.mode = mode
 
         pm_mod = modelcontext(None)
+        owner = getattr(pm_mod, "_statespace_model", None)
```
ricardoV94: don't like this, what if I clone models? Don't put stuff in pymc models please :)

tests/statespace/core/test_statespace.py
```diff
@@ -538,7 +537,157 @@ def test_build_statespace_graph_raises_if_data_has_missing_fill():
...
+    ss_mod, build_priors = _make_idempotent_test_model()
+    data = _idempotent_test_data()
+
+    with pm.Model(coords=ss_mod.coords) as pymc_mod:
+        build_priors()
+
+        ss_mod.build_statespace_graph(data, save_kalman_filter_outputs_in_idata=save_kf)
+        logp_first = pymc_mod.compile_logp()(pymc_mod.initial_point())
+
+        ss_mod.build_statespace_graph(data, save_kalman_filter_outputs_in_idata=save_kf)
```
ricardoV94: I would expect this would build two distinict sts graphs?

tests/statespace/core/test_statespace.py
```diff
@@ -538,7 +537,157 @@ def test_build_statespace_graph_raises_if_data_has_missing_fill():
...
+def test_build_statespace_graph_twice_with_different_data():
+    # Rebuilding with data of a different length must refresh the "time" coordinate rather than complaining
+    # that the new data does not match the index registered by the first build.
+    ss_mod, build_priors = _make_idempotent_test_model()
+
+    with pm.Model(coords=ss_mod.coords) as pymc_mod:
+        build_priors()
+
+        ss_mod.build_statespace_graph(_idempotent_test_data(n=40))
+        ss_mod.build_statespace_graph(_idempotent_test_data(n=63, start="2021-06-01"))
```
ricardoV94: Don't like it, if you want it should then be called [rebuild|update]_statespace_graph or something.

ricardoV94:
refactor commits in separate PR?

pymc_extras/statespace/core/statespace.py
jessegrabowski: this is old, i went for the bigger pr to avoid this.

ricardoV94: I checked commit :(

jessegrabowski:
@ricardoV94 i re-scoped this PR, it didn't do what I wanted it to do anyway. I split out the reorganization into #727 and i'll keep working towards breaking up the god StatespaceModel class, which will give me the idempotency I want. For this PR I'm now just ripping out a couple useless settings.

ricardoV94 (approved):

PR #695, opened 2026-06-19
Bump pre-commit and re-run notebooks

jessegrabowski:
Updates our wildly out-of-date ruff and runs the new settings.

This re-linted the notebooks, which I then realized were also wildly out of date. So I updated and re-ran them all. 

I didn't run the INLA one or the GP approximation one because the GP one seems out of date, and the INLA one seems like it never ran.


jessegrabowski:
Closing — this was accidentally opened from a branch on the upstream repo (pymc-devs) rather than my fork. Reopened from jessegrabowski:pre-commit-bump at https://github.com/pymc-devs/pymc-extras/pull/696.

PR #700, opened 2026-06-26
Automatically minibatch a model

ricardoV94:


pymc_extras/model/transforms/minibatch.py
```diff
@@ -0,0 +1,361 @@
...
+
+        with mb_model:
+            approx = pm.fit()
+
+    """
+    if not isinstance(batch_size, int):
+        raise TypeError("batch_size must be an integer")
+
+    if observed is None:
+        observed = [rv for rv in model.observed_RVs if rv.ndim > 0]
```
ricardoV94, line 149: grab the ones with batch ndim not just rv.ndim > 0

PR #704, opened 2026-07-06
Support more affine graphs in normal-normal marginal

ricardoV94:
- **Avoid duplicating shared RVs when unmarginalizing**
- **Control for broadcasting in normal-normal marginalization**
- **Support affine dependent means in normal-normal marginalization**

There's a more general approach for chains of Normal-Normal (with one to many, many to one and so on, and several links deep) that I might pursue in later that would subsume these "hand-crafted" restricted cases. But right now this is mostly demonstration purpose and ironing out rough edges through experimentation

### DIFF
--- a/pymc_extras/model/marginal/distributions/discrete_markov_chain.py
+++ b/pymc_extras/model/marginal/distributions/discrete_markov_chain.py
@@ -26,7 +26,7 @@ from pymc_extras.model.marginal.distributions.enumerable import (
 from pymc_extras.model.marginal.rewrites import (
     MarginalSubgraph,
     extract_marginal_subgraph,
-    marginal_rewrites_db,
+    marginal_ir_rewrites_db,
 )
 
 
@@ -411,6 +411,6 @@ def discrete_markov_chain_marginal(fgraph, node):
     return build_enumerable_marginal_rv(node, inputs, outputs, MarginalDiscreteMarkovChainRV)
 
 
-marginal_rewrites_db.register(
+marginal_ir_rewrites_db.register(
     "discrete_markov_chain_marginal", discrete_markov_chain_marginal, "basic"
 )
--- a/pymc_extras/model/marginal/distributions/normal.py
+++ b/pymc_extras/model/marginal/distributions/normal.py
@@ -16,7 +16,7 @@ from pymc_extras.model.marginal.distributions.core import (
 from pymc_extras.model.marginal.rewrites import (
     MarginalSubgraph,
     extract_marginal_subgraph,
-    marginal_rewrites_db,
+    marginal_ir_rewrites_db,
 )
 
 
@@ -130,4 +130,4 @@ def normal_normal_marginal_rewrite(fgraph, node):
     return new_outputs[: len(node.outputs)]
 
 
-marginal_rewrites_db.register("normal_normal_marginal", normal_normal_marginal_rewrite, "basic")
+marginal_ir_rewrites_db.register("normal_normal_marginal", normal_normal_marginal_rewrite, "basic")
--- a/pymc_extras/model/marginal/distributions/normal.py
+++ b/pymc_extras/model/marginal/distributions/normal.py
@@ -6,6 +6,7 @@ from pytensor.graph import node_rewriter
 from pytensor.graph.replace import graph_replace
 from pytensor.graph.traversal import ancestors
 from pytensor.tensor import broadcast_to, constant, sqrt
+from pytensor.tensor.rewriting.basic import broadcasted_by
 
 from pymc_extras.model.marginal.distributions.core import (
     MarginalRV,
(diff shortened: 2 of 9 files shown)


pymc_extras/model/marginal/distributions/normal.py
```diff
@@ -48,9 +91,43 @@ def normal_normal_marginal_rv_logp(op: NormalNormalMarginalRV, values, *inputs,
...
+    perm = (*batch_axes, *shared_axes)
+    dep_shape = get_symbolic_rv_shapes([dependent_rv])[0]
+
+    def to_event(t):
+        return flatten(broadcast_to(t, dep_shape).transpose(perm), ndim=len(batch_axes) + 1)
+
+    mean = to_event(new_mu)
+    u = to_event(slope * sigma_m)
+    cov = u[..., :, None] * u[..., None, :] + to_event(sigma_d**2)[..., :, None] * eye(u.shape[-1])
+    return logp(MvNormal.dist(mu=mean, cov=cov), to_event(value))
```
ricardoV94: This makes a diagonal + rank1 update MvNormal logp that we should try to optimize natively in pytensor. We may need MvNormal logp to stop eagerly producing optimized code for cholesky factors, as that form is not trivially optimizable like some with navie solve and slogdet would be. Since this is niche functionality I'm leaving the naive form, and I'll open a PyTensor (and potentially PyMC) issue to follow up. Similarly we may be able to get rid of the PrecisionMvNormal we have now in PyMC.

CC @jessegrabowski @bwengals if it tickles your fingers

jessegrabowski: it does. I just need some help thinking about rank as a matrix property and how to implement parameterized assumptons. Or maybe we don't need to and I'm being a dolt?

ricardoV94: This one is rank1 update so it's simpler and easier to see statically (a constant matrix or one bulit from an outer mul)

jessegrabowski: sure and maybe that's a fine place to start, but i'd like to have eyes on the full problem.

bwengals: Maybe a pytensor function for rank one updates, `A2 = pt.rankoneupdate(A, a, u, v)`?  Then it's not a matrix property but pytensor still sees it.  Def agree that MvNormal logp shouldn't handle eagerly producing optimized code for cholesky factors, that closes a lot of doors.

ricardoV94: You don't need an op, A2 = A + pt.assume(u @ v.T, rank=1), in case you don't trust PyTensor to be able to see this (or you are passing a root matrix that you know is rank1)

pymc_extras/model/marginal/distributions/normal.py
```diff
@@ -61,13 +103,18 @@ def normal_normal_conditional(op, inputs, dep_rvs):
...
     precision_d = 1 / sigma_d**2
-    posterior_precision = precision_m + precision_d
+    posterior_precision = precision_m + slope**2 * precision_d
     posterior_sigma = sqrt(1 / posterior_precision)
-    posterior_mu = (mu_m * precision_m + (dep_rv - offset) * precision_d) / posterior_precision
+    posterior_mu = (
+        mu_m * precision_m + slope * (dep_rv - offset) * precision_d
+    ) / posterior_precision
 
     return Normal.dist(mu=posterior_mu, sigma=posterior_sigma)
```
zaxtax, line 119: Are we sure we want these parameters inside the normal? I can imagine a situation where someone wrote things this way to avoid Neal's funnel and it would be irritating if the rewrite undid that work.

pymc_extras/model/marginal/distributions/discrete_markov_chain.py
```diff
@@ -336,6 +336,6 @@ def discrete_markov_chain_marginal(fgraph, node):
     return build_enumerable_marginal_rv(node, inputs, outputs, MarginalDiscreteMarkovChainRV)
 
 
-marginal_rewrites_db.register(
+marginal_ir_rewrites_db.register(
```
zaxtax, line 414: Can this rename be split out into it's own commit? Not a deal-breaker but made it slightly harder to see what's going on

pymc_extras/model/marginal/distributions/normal.py
ricardoV94: I don't think it's worth the effort of figuring out what parametrization the original graph had or the new one. Usually you don't need to mcmc sample with the conditional anyway, just do forward/sample_posterior_predictive which don't care about centered / non-centered.

It's also not always better to do non-centered. And in this case would require us to get the size from the arguments. So I'd leave as revisit when we find a need?

pymc_extras/model/marginal/distributions/discrete_markov_chain.py
ricardoV94: done

pymc_extras/model/marginal/distributions/normal.py
zaxtax: If this only triggers when the latter normal is observed, feel free to disregard my comment. But if it's any pair of Normals, I think this is something to be concerned about

ricardoV94: This is for the conditional, observed or not. You usually don't to mcmc on the conditional. 

In any case right now we don't support broadcasting so there's no hierarchical funnel, it's a 1-to-1 normal relationship

zaxtax (approved):

PR #709, opened 2026-07-10
Add serialization to Scaled distribution

anevolbap:
## Add serialization to `Scaled`

`Scaled(dist, factor)` had `create_variable` and a `dims` property but no `to_dict`/`from_dict`, and was not registered with the deserializer. So it could not round-trip like `Prior` and `Censored`. This adds that.

### What changed
- `Scaled.to_dict()` returns `{"class": "Scaled", "data": {"dist": dist.to_dict(), "factor": <encoded>}}`.
- `Scaled.from_dict()` reconstructs a `Scaled`, deserializing the inner `dist` and decoding the factor.
- Registered `_is_scaled_type` / `Scaled.from_dict` with `register_deserialization`, so `deserialize(...)` rebuilds a `Scaled`.

I used the same `{"class", "data"}` wrapper as `Censored` on purpose. The `Prior` deserializer matches any dict that has a `"dist"` key, so a flat `{"dist": ..., "factor": ...}` would be picked up as a `Prior`. The wrapper avoids that collision.

### Factor encoding
`factor` is `XTensorLike`. The contract is the smallest that covers real use:
- scalar: stored as-is (the common case, e.g. `factor=50_000`)
- numpy array: stored as a list, decoded back with `np.array`
- `xarray.DataArray`: stored as `{"class": "DataArray", ...}`, decoded by the existing DataArray deserializer
- pytensor variable: evaluated first, then handled as above

Anything else is passed through untouched, so a non-serializable factor fails at the serialization layer, not silently.

### Tests
Added to `TestScaled` in `tests/test_prior.py`: `to_dict` shape, `deserialize` of a hand-written dict, and round-trips for scalar, numpy-array, and DataArray factors.

### Motivation
pymc-marketing's `BassModel` save/load needs `Scaled` priors to serialize. Today pymc-marketing carries a shim (`_serialize_scaled`/`_deserialize_scaled` in `special_priors.py`) that can be removed once this lands. Raised in pymc-labs/pymc-marketing#2724.

### DIFF
--- a/pymc_extras/prior.py
+++ b/pymc_extras/prior.py
@@ -1733,8 +1733,8 @@ class Scaled:
     def to_dict(self) -> dict[str, Any]:
         """Convert the scaled distribution to a dictionary.
 
-        The ``factor`` is encoded with :func:`_serialize_value`: a scalar is
-        stored as-is, a numpy array as a list, and an ``xarray.DataArray`` as a
+        The ``factor`` is encoded the same way as ``Prior`` parameters: a
+        scalar is stored as-is, a numpy array as a list, and an ``xarray.DataArray`` as a
         ``{"class": "DataArray", ...}`` mapping. A pytensor variable is
         evaluated first. Non-tensor factors (arbitrary Python objects) are left
         untouched, so a factor that is not JSON-serializable will fail at the
--- a/pymc_extras/prior.py
+++ b/pymc_extras/prior.py
@@ -101,6 +101,7 @@ from pydantic.dataclasses import dataclass
 from pymc.distributions.shape_utils import Dims
 from pytensor.graph import Variable
 from pytensor.tensor import TensorVariable
+from pytensor.xtensor.type import XTensorType
 from xarray import DataArray, Dataset
 
 from pymc_extras.deserialize import deserialize, register_deserialization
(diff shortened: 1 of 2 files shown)


pymc_extras/prior.py
```diff
@@ -1722,6 +1722,68 @@ def create_variable(self, name: str, xdist: bool = False) -> TensorVariable | XT
...
+            The dictionary format of the scaled distribution.
+        """
+
+        def handle_factor(value):
+            if isinstance(value, Variable):
+                if isinstance(value.type, pt.TensorType):
+                    value = value.eval()
+
+                # Avoid XTensor import warnings, remove this when the warnings are gone
+                elif value.type.__class__.__name__.startswith("XTensor"):
```
ricardoV94: they are gone

anevolbap:
pymc-extras `main` currently fails to import against pymc `main` because `marginalize.py` imports `extract_dims`, which pymc removed. That is already covered by #708 (pymc 6.1 / pytensor 3.1 compatibility). This PR only touches `prior.py` and has no fgraph dependency, so it should be unaffected and go green once #708 merges.

pymc_extras/prior.py
```diff
@@ -1722,6 +1722,68 @@ def create_variable(self, name: str, xdist: bool = False) -> TensorVariable | XT
 
         return det_class(name, var * self.factor, dims=self.dims)
 
+    def to_dict(self) -> dict[str, Any]:
```
ricardoV94, line 1733: did you say this is very similar to Censored? In that case can we share a helper?

ricardoV94 (approved):
Looks great

pymc_extras/prior.py
anevolbap: Done, now `isinstance(value.type, XTensorType)`.

pymc_extras/prior.py
anevolbap: Shared it as `_serialize_value`, now used by both Prior and Scaled. I left Censored out for now since it serializes its bounds a bit differently, but happy to unify that too. I can also extract the decode side so Prior/Scaled/Censored share both directions, in a separate PR if you prefer.

ricardoV94 (approved):
Thanks

ricardoV94:
@anevolbap did you mark this as draft? If so anything else you wanted to do before merge?

ricardoV94:
Ah this needs a rebase from main

anevolbap:
Rebased onto main. The sdist failure was the `extract_dims` import, fixed by #708, which the branch was missing.

anevolbap:
CI is green now, the sdist failure was the missing #708 fix. Anything left before this can go in?

ricardoV94:
Thanks @anevolbap

PR #713, opened 2026-07-18
Reworking Minibatching for ADVI

zaxtax:
This is a dedicated PR for how I think fitting on data too large to give to a model directly. Namely, I don't think models should have to know about how a data is batched. I think we should have an external iterator that makes these decisions for us. This would let us move a lot of complexity out of the library. It also enables more complicated setups. For example where we are streaming in multiple datasets.

Additionally, I added a nicety in the way of `observeds` for fitting. The idea with it is that many Minibatch examples end up needing a placeholder `pm.Data` and it would simplify the model if we could define a joint probability distribution and observed certain variables after the fact.

One area, I am less confident in my design is how I scale the likelihood. Right now, I assume I can call `len` on the iterator and use that to guide scaling, and otherwise hope there is a field I can read in each batch. Feedback and input here would be welcome.

Also right now this PR is more designed to build than be a clean minimal commit. Once we agree on what this PR should do, happy to rebase and clean it up.


pymc_extras/inference/advi/objective.py
```diff
@@ -43,7 +52,16 @@ def get_logp_logq(model: Model, guide: AutoGuideModel, path_derivative_gradient:
         if rv not in model.observed_RVs
     }
 
-    logp = graph_replace(model.logp(), inputs_to_guide_rvs)
+    if logp_scalings:
+        scaled = set(logp_scalings)
+        rest = [var for var in (*model.basic_RVs, *model.potentials) if var not in scaled]
+        model_logp = model.logp(vars=rest)
```
ricardoV94: model.logp(sum=False) -> gives you the logp per variable then you can sum / scale however you want each component

pymc_extras/inference/advi/fit.py
```diff
@@ -0,0 +1,97 @@
+from __future__ import annotations
```
jessegrabowski, line 1: unused

pymc_extras/inference/advi/optimizers.py
```diff
@@ -0,0 +1,137 @@
+"""Optax-like gradient transformations operating on dicts of numpy parameters.
+
+The API follows optax: an optimizer is a ``GradientTransformation`` with
+``init(params) -> state`` and ``update(grads, state, params) -> (updates, state)``,
+where updates are increments added to the parameters with :func:`apply_updates`.
+Actual optax optimizers are plug-compatible.
+"""
+
+from __future__ import annotations
```
jessegrabowski: I would prefer no module level docstrings

pymc_extras/inference/advi/optimizers.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def apply_updates(
+    params: dict[str, np.ndarray], updates: dict[str, np.ndarray]
+) -> dict[str, np.ndarray]:
+    """Add the updates to the parameters."""
+    return {name: np.asarray(param + updates[name]) for name, param in params.items()}
+
+
+def chain(*transforms: GradientTransformation) -> GradientTransformation:
```
jessegrabowski, line 59: Not a blocker but this is duplicative of what we're doing in pytensor_ml e.g. [here](https://github.com/pymc-devs/pytensor_ml/blob/80645590b40ec48d128f0bd08312b57aa6314881/pytensor_ml/optim/base.py#L175), would be nice if we can refactor to use that instead in a follow-up

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,705 @@
...
+        (e.g. "numba", "jax", "c"). Mutually exclusive with ``compile_kwargs["mode"]``.
+    compile_kwargs : dict, optional
+        Additional kwargs passed to pytensor compilation.
+    random_seed : optional
+        Seed for the default guide's initialization. Seeds for the training and
+        posterior draws are passed to :meth:`fit` and :meth:`sample_posterior`.
+
+    Example
+    -------
+    >>> with pm.Model() as model:
```
jessegrabowski, line 167: nit: prefer .. code-block::python sphinx directive to the old-style python interpreter examples. I guess nobody actually types anymore, but it always drives me crazy when I can't easily copy/paste the example because it's got the interpreter >>> and ... mixed in (also I never even knew that's why it's there, I had never fired up a plain interpreter until very recently)

pymc_extras/inference/advi/optimizers.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def update(updates, state, params=None):
+        global_norm = np.sqrt(sum(np.sum(np.square(g)) for g in updates.values()))
+        scale = np.minimum(1.0, max_norm / (global_norm + 1e-12))
+        return {name: g * scale for name, g in updates.items()}, state
+
+    return GradientTransformation(init, update)
+
+
+def scale_by_adam(b1: float = 0.9, b2: float = 0.999, eps: float = 1e-8) -> GradientTransformation:
```
jessegrabowski, line 97: Looking more closely I think we should just make the cut over to pytensor_ml. The API is basically the same but the updates come out as actual pytensor updates, so you don't have this blended numpy/pytensor setup. It would also open up scan-based training loop for e.g. jax backend.

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,705 @@
...
+        An optax-like optimizer (actual optax optimizers are compatible). By default
+        a clipped-Adam update on ``learning_rate`` is compiled *into* the step
+        function, so no parameters or gradients round-trip through Python per step
+        (fast path); passing an explicit optimizer switches to a Python-side update
+        loop.
+    learning_rate : float or callable, optional
+        Learning rate, or a schedule mapping the step number to one, for the default
+        compiled optimizer. Defaults to a :func:`linear_onecycle_schedule` peaking at
+        0.008 over the ``n`` steps of each :meth:`fit` call. Ignored when
+        ``optimizer`` is given (set the learning rate on the optimizer itself).
```
jessegrabowski: nutpie has this issue too, where learning rate and optimizer arguments have this weird coupling. i think it's the worst part of the optax-like setup. No suggestion for fix, just complaining.

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,705 @@
...
+    clip_norm : float, optional
```
jessegrabowski: I think this goes too far as a separate argument. A user should just build the chain she wants if she wants clipping.

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,705 @@
...
+        self.convergence_window = convergence_window
+        self.relative_tolerance = relative_tolerance
+        self.model = model
+        self.compile_kwargs = resolve_backend_compile_kwargs(backend, compile_kwargs)
+        self.random_seed = random_seed
+        self.state: SVIState | None = None
+
+        self._guide: AutoGuideModel | None = guide if isinstance(guide, AutoGuideModel) else None
+        self._param_names: list[str] | None = None
+        self._fit_model: Model | None = None
```
jessegrabowski, line 197: I'm worried that attaching the models to the training is repeating some of the mistakes of the existing pymc ADVI. Is there any way to make this all more modular and lean on `pm.modelcontext`?

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,705 @@
...
+            return
+        self._sampling_fn = compile_sampling_fn(
+            model=model,
+            guide=self._guide,
+            draws=draws,
+            **self.compile_kwargs,
+        )
+        self._sampling_draws = draws
+
+    def _should_stop(self, step: int, loss_history: list[float]) -> bool:
```
jessegrabowski: How does lightning handle early stopping? It's a part of the trainer class? It seems like it should be a property of the optimizer maybe, but I'm not sure. Just concerned by how much stuff the Trainer class is doing.

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,705 @@
...
+                        if isinstance(rv.owner.op, MinibatchRandomVariable):
+                            # The model already rescales this likelihood: a total_size
+                            # passed to the observed RV wraps it in a
+                            # MinibatchRandomVariable whose logp carries the
+                            # N / batch_size factor, so adding ours would scale twice
+                            continue
+                        if var in ancestors([model.rvs_to_values[rv]]):
+                            self._logp_scalings[rv.name] = scale
+                continue
+            if name not in first_batch:
```
jessegrabowski, line 340: I don't fully understand the first batch thing

pymc_extras/inference/advi/training.py
```diff
@@ -1,38 +1,705 @@
...
+                )
+        except KeyboardInterrupt:
+            pass
+
+        state = finalize(state)
+        self.state = state
+
+        return state
+
+    def sample_posterior(
```
jessegrabowski, line 560: Same question/concern about the scale of the Trainer above. We might want a pmx.sample_posterior instead, to match the usual pymc api

tests/inference/advi/test_fit.py
```diff
@@ -0,0 +1,229 @@
...
+    draws_a = fit_advi(random_seed=42, **kwargs)["posterior"].dataset["theta"].values
+    draws_b = fit_advi(random_seed=42, **kwargs)["posterior"].dataset["theta"].values
+    draws_c = fit_advi(random_seed=13, **kwargs)["posterior"].dataset["theta"].values
+
+    np.testing.assert_array_equal(draws_a, draws_b)
+    assert not np.array_equal(draws_a, draws_c)
+
+
+@pytest.mark.filterwarnings("ignore:The RandomType SharedVariables")
+def test_fit_advi_random_seed_jax(conjugate_model):
```
jessegrabowski: This seems like a test of pytensor, not your ADVI

tests/inference/advi/test_fit.py
```diff
@@ -0,0 +1,229 @@
...
+        random_seed=1,
+        convergence_window=50,
+        relative_tolerance=10.0,
+    )
+
+    # With a huge tolerance, training stops at the first convergence check
+    assert idata["fit"].dataset.sizes["step"] == 100
+
+
+def test_guide_initialized_at_initial_point():
```
jessegrabowski: is this a test of pymc?

jessegrabowski (changes requested):
Requesting changes but I think it's more of a request discussion/feedback, only a few concrete asks.

pymc_extras/inference/advi/training.py
zaxtax: Lightning using callbacks for early stopping. I have mixed feelings about this

pymc_extras/inference/advi/training.py
zaxtax: Basically it's there because we need a shape for observed variables since those are going to be wrapped in a `pm.Data` object. Since this is going to be tied to batch size, it's hard to do in advance of looking at the data. Any alternatives come to mind?

pymc_extras/inference/advi/optimizers.py
zaxtax: I'm totally fine with a follow-up PR doing that kind of cutover

pymc_extras/inference/advi/optimizers.py
zaxtax: I'm totally fine with a follow-up PR doing that kind of cutover

pymc_extras/inference/advi/training.py
zaxtax: The problem is that we modify the model during fitting and that needs to happen somewhere. If there is an error or shape issue during fitting it could leave the original model is a hard to restore state. I think it's better to make a copy that's effectively discarded later.

pymc_extras/inference/advi/training.py
zaxtax: Dropping early stopping from PR

pymc_extras/inference/advi/training.py
zaxtax: Let's move those to the optimziers

jessegrabowski (approved):
looks good to me now, we can iron it out through usage. I will open a follow-up PR to explore integration with pytensor_ml

pymc_extras/inference/advi/objective.py
```diff
@@ -43,7 +54,16 @@ def get_logp_logq(model: Model, guide: AutoGuideModel, path_derivative_gradient:
         if rv not in model.observed_RVs
     }
 
-    logp = graph_replace(model.logp(), inputs_to_guide_rvs)
+    if logp_scalings:
+        logps = model.logp(sum=False)
+        all_vars = model.free_RVs + model.observed_RVs + model.potentials
```
ricardoV94, line 58: you should pass all_vars to model.logp(vars=), to be sure you are getting the same order

pymc_extras/inference/advi/objective.py
```diff
@@ -43,7 +54,16 @@ def get_logp_logq(model: Model, guide: AutoGuideModel, path_derivative_gradient:
...
+        scales = pt.constant([logp_scalings.get(var, 1.0) for var in all_vars])
+        summed_logps = pt.stack([pt.sum(logp) for logp in logps])
+        model_logp = pt.dot(scales, summed_logps)
```
ricardoV94: Why not multiply each summed term by the scale intead of the stack and dot?

`pt.sum([pt.sum(logp) * scale, for logp, scale in zip(logps, scales)])`

pymc_extras/inference/advi/training.py
ricardoV94: how are you modifying the model, you should not modify a user model, do .copy() at the start at least

zaxtax:
I don't modify the model, I copy it!

ricardoV94:
This is not independent of #635 though is it?

Also make explicit that (auto)batch axis is assumed to be on the left somewhere in the user facing docstrings.

zaxtax:
> This is not independent of #635 though is it?
> 
> Also make explicit that (auto)batch axis is assumed to be on the left somewhere in the user facing docstrings.

It's not independent, but it shares commits and doesn't require the other PR to be merged in.

tests/inference/advi/test_fit.py
```diff
@@ -0,0 +1,213 @@
...
+        while True:
+            yield {"y": rng.normal(1.0, 1.0, size=64)}
+
+    with pm.Model() as model:
+        theta = pm.Normal("theta", 0, 10)
+        pm.Normal("y", theta, 1, shape=(64,))
+
+    trainer = Trainer()
+    with model:
+        trainer.fit(1_000, batches(), observeds=["y"], random_seed=1)
```
fonnesbeck, line 240: `observed_vars` or similar would be better than `observeds` if this is not already set in stone. More consistent with passed lists of var names in other pymc functions.

fonnesbeck: Related to this, can we not use `pm.observe`?

zaxtax: In the implementation, that's what we do!

https://github.com/pymc-devs/pymc-extras/pull/713/changes#diff-9baa443098884cbd96fc90b83279a4702a26fac98b255227cf8366590a93917cR335

zaxtax: And nothing is set in stone here 🙂

notebooks/ADVI Minibatch Iterator.ipynb
```diff
@@ -0,0 +1,1716 @@
+{
```
juanitorduz, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    def batch_iterator(rng, X, y, batch_size):
</pre><p>it would be great to have some rule of thumb of what a good batch_size will be in general (roughly, to start with)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-extras/pull/713/discussion/#comment-3783776179'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

zaxtax:
@juanitorduz early stopping will be supported, but decided to make it a follow-up PR. I don't know what's required to make this all run on a GPU via JAX.

> View / edit / reply to [this conversation](https://app.reviewnb.com/pymc-devs/pymc-extras/pull/713/discussion/#comment-5293542414) on ReviewNB
> 
> _juanitorduz commented on 2026-08-14T12:57:20Z_ ----------------------------------------------------------------
> 
>     * Any chance we could (in the future) support early stopping based on certain metrics?
> 
>     * Can we run this on GPU via JAX?

PR #717, opened 2026-07-24
Compatibility with PyTensor v3.2 and PyMC v6.2

jessegrabowski:
Primarily for pymc/pytensor, but I also capped versions on anything with a minimum, and also updated arviz to clear some matplotlib pins.

### DIFF
--- a/tests/inference/pathfinder/test_multipath.py
+++ b/tests/inference/pathfinder/test_multipath.py
@@ -1,4 +1,5 @@
 import contextlib
+import sys
 
 import numpy as np
 import pytest
@@ -55,6 +56,9 @@ def test_parallel_paths_match_serial_per_path():
     """Each chain gets the same per-path approximation under parallel and serial execution (same
     seed, within cross-process BLAS noise) -- guarding against state leaking across the shared
     compiled functions."""
+    if sys.platform == "win32":
+        pytest.skip("non-deterministic on Windows CI workers")
+
     model = make_ard_regression()
     kw = dict(
         method="pathfinder",
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -14,7 +14,7 @@ dependencies:
   - pydantic>=2.0.0,<3.0
   - h5netcdf
   - pymc>=6.2.0,<6.3
-  - pytensor>=3.2.2,<3.3
+  - pytensor>=3.2.3,<3.3
   - arviz>=1.2,<2.0
   - preliz>=0.27,<0.28
   - pip
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -4,17 +4,17 @@ channels:
 - nodefaults
 dependencies:
   - scikit-learn
-  - better-optimize>=0.4.2
+  - better-optimize>=0.4.2,<1.0
   - dask<2025.1.1
   - xhistogram
   - statsmodels
   - numba
   - pytest
   - pytest-cov
-  - pydantic>=2.0.0
+  - pydantic>=2.0.0,<3.0
   - h5netcdf
-  - pymc>=6.1.0,<6.2
-  - pytensor>=3.1.3,<3.2
+  - pymc>=6.2.0,<6.3
+  - pytensor>=3.2.2,<3.3
   - preliz>=0.26,<0.27
   # arviz-plots 1.1.0 imports matplotlib.style.core, removed in matplotlib 3.11
   - matplotlib<3.11
@@ -22,5 +22,3 @@ dependencies:
   - pip:
       - jax
       - blackjax
-      - pytensor>=3.0.4
-      - preliz>=0.25
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -15,9 +15,8 @@ dependencies:
   - h5netcdf
   - pymc>=6.2.0,<6.3
   - pytensor>=3.2.2,<3.3
+  - arviz>=1.2,<2.0
   - preliz>=0.26,<0.27
-  # arviz-plots 1.1.0 imports matplotlib.style.core, removed in matplotlib 3.11
-  - matplotlib<3.11
   - pip
   - pip:
       - jax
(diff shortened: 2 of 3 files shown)


jessegrabowski:
@ricardoV94 I think there's a bug in the fused add-mul kernel that is causing the discrete markov chain logp test to fail. Robot report follows

---------------

# PyTensor 3.2 regression: `sum()` over a broadcast multi-array advanced index returns wrong values

## Summary

On PyTensor **3.2.2** (regressed from **3.1.3**), reducing over an `AdvancedSubtensor`
gather whose index arrays have **different shapes that broadcast together** produces the
wrong result. A new rewrite introduced in 3.2, **`fuse_indexed_into_elemwise`** (registered
under the `numba` tag, in `pytensor/tensor/rewriting/fused_elemwise.py`), fuses
`reduce(elemwise(advanced_subtensor(...)))` into a single `FusedElemwise` kernel, and that
fused kernel mis-broadcasts the index arrays.

The raw gather is correct; only the fused gather-plus-reduction is wrong.

## Environment

- Affected: `pytensor==3.2.2`
- Correct:  `pytensor==3.1.3`
- Backend:  numba (default `mode`/`linker=auto` resolves here). The offending rewrite is
  numba-tagged, so the C backend is unaffected.

## Minimal reproduction

```python
import numpy as np
import pytensor
import pytensor.tensor as pt
from pytensor.compile.mode import get_default_mode

x_np = rng.random((2, 2, 2))                        # axes: (a, b, c)

x = pt.tensor("x", shape=(2, 2, 2))
i_a = pt.arange(2)[:, None]                         # (2, 1)  broadcasts over axis 0
i_b = pt.arange(2)                                  # (2,)    broadcasts over axis 1
i_c = pt.tensor("i_c", shape=(2, 2), dtype="int64") # (2, 2)  explicit

gathered = x[i_a, i_b, i_c]                          # AdvancedSubtensor -> (2, 2)
out = gathered.sum(axis=-1)                          # sum(x[idx]) -> triggers 3.2 fusion

i_c_np = np.array([[0, 1], [1, 0]])
ref = np.array(
    [[x_np[a, b, i_c_np[a, b]] for b in range(2)] for a in range(2)]
).sum(axis=-1)

f_default = pytensor.function([x, i_c], out)
f_nofuse = pytensor.function(
    [x, i_c],
    out,
    mode=get_default_mode().excluding("fuse_indexed_into_elemwise"),
)

print("pytensor", pytensor.__version__)
print("default :", f_default(x_np, i_c_np), np.allclose(f_default(x_np, i_c_np), ref))
print("no-fuse :", f_nofuse(x_np, i_c_np), np.allclose(f_nofuse(x_np, i_c_np), ref))
print("ref     :", ref)
```

## Observed output (pytensor 3.2.2)

```
pytensor 3.2.2
default : [0.9067484  1.33613234] False
no-fuse : [0.65348932 1.51939135] True
ref     : [0.65348932 1.51939135]
```

The default (fused) result is wrong; excluding the fusion rewrite gives the correct answer.
`FAST_COMPILE` (no rewrites) is also correct — confirming the defect is in the rewrite, not
in the gather semantics.

## Confirmation

Compiling the graph with the rewrite excluded fixes it:

```python
mode=get_default_mode().excluding("fuse_indexed_into_elemwise")
# or, at process level:
# PYTENSOR_FLAGS="optimizer_excluding=fuse_indexed_into_elemwise"
```

The dprint of the broken graph shows the fused op:

```
FusedElemwise{Log, reduce[add@(1,)]}   # (when the elemwise is a log; add-reduction here)
 └─ AdvancedSubtensor{idx_list=(0, 1, 2)}
```

## Impact in pymc-extras

`discrete_mc_logp` in `pymc_extras/distributions/timeseries.py` gathers per-step transition
log-probabilities with exactly this pattern and reduces over the last axis:

```python
mc_logprob += log_P[(*batch_index, *time_index, *indices)].sum(axis=-1)
```

where `batch_index` (from `ix_`) is `(n, 1)`, `time_index` is `(t,)`, and the state indices
are `(n, t)`. This is the mixed-shape broadcast case above. It surfaces as a failure in
`tests/model/marginal/test_discrete_markov_chain.py::test_conditional_discrete_markov_chain_higher_order[batched=True]`
(batched, time-varying, `n_lags=2`), which passes on pytensor 3.1.3 and passes on 3.2.2 with
the rewrite excluded.

ricardoV94:
likely it's the big new fusion extension we rolled out. Although there's also some llvm bugs going around in this space, but that's usually segfaults

ricardoV94:
@jessegrabowski should be fixed by https://github.com/pymc-devs/pytensor/pull/2306

jessegrabowski:
@ricardoV94 this is ready

tests/inference/pathfinder/test_multipath.py
```diff
@@ -68,13 +75,23 @@ def test_parallel_paths_match_serial_per_path():
...
+        concurrent = pmx.fit(cores=4, **kw).posterior
+
+    assert sequential.sizes["chain"] == concurrent.sizes["chain"] == kw["num_paths"]
+    seq_means = _per_path_means(sequential)
+    con_means = _per_path_means(concurrent)
+
+    # On failure, attribute the mismatch: for each concurrent path, which sequential chain is it
+    # nearest to? nearest == its own index means a perturbation of its own result (upstream
+    # per-process JIT nondeterminism); nearest != its own index means a result attributed to the
+    # wrong chain -- a leak in the parent's concurrent reassembly (a bug in this repo).
```
ricardoV94: red flag, what's going on?

jessegrabowski: Something on windows that I can't test on my local. There seems to be non-deterministic behavior on the workers that was causing the windows CI to fail e.g. [here](https://github.com/pymc-devs/pymc-extras/actions/runs/30312754265/job/90131809744). This appears to be new.

ricardoV94: skip, open issue and remove made up bot stuff?

jessegrabowski: sure

tests/inference/pathfinder/test_multipath.py
```diff
@@ -51,6 +51,7 @@ def _per_path_means(posterior):
     return np.asarray(rows)
 
 
+@pytest.mark.skip(reason="non-deterministic on Windows CI workers")
```
ricardoV94: do we run this only on windows, we could skip only in it

jessegrabowski: your wish is my command

ricardoV94 (approved):

PR #722, opened 2026-07-31
Streaming Pathfinder: minibatch L-BFGS with same-batch curvature pairs

YichengYang-Ethan:
Adds `fit_streaming_pathfinder`: Pathfinder initialization for datasets that never fit in memory, driven through the merged DataLoader contract. The loader must expose `total_size` (the dataset row count N); the driver refuses one that cannot, since the importance weights need an exact full-data pass and there is no honest way to guess N.

Two design decisions carry the PR.

**Curvature pairs come from a single minibatch.** Both gradients of each `(s, y)` pair are evaluated on the same batch, so the batch noise cancels in the difference (Schraudolph, Yu & Günter 2007); the batch advances only after the pair is formed. Cross-batch pairing violated the curvature condition on 0.3-30% of steps, rising with batch size; same-batch produced zero violations on every cell run.

**Polyak-Ruppert tail averaging replaces ELBO iterate selection.** Every step is a quasi-Newton step plus line search on one minibatch, so each accepted iterate sits near that batch's MAP — a location error, not a covariance one, which no rule that picks an iterate can fix (an oracle over all 200 iterates only reached pareto_k 2.4). Averaging the last 75% of positions does: against an exact full-data Laplace reference (Bayesian logistic regression, batch 2048, `num_iters=200`), pareto_k at k=8, N=1e5 runs 0.26-0.58 over 12 seeds with the worst posterior-mean coordinate 0.05-0.30 reference-sd out. On real data the picture is friendlier because the objective is quadratic: 1M rows of Binance tick data against a closed-form posterior streamed through the same loader gives pareto_k 0.45 and 0.03 reference-sd; 30.7M rows give 0.30 and 0.11.

**Where it degrades, stated plainly.** A pre-registered held-out battery (three families never used during development, shipped defaults, three seeds each against full-data Laplace references) puts sharper edges on the range: a Poisson GLM (4 params) is clean at pareto_k 0.08-0.33; Student-t regression with estimated nu (5 params) is marginal at 0.58-0.84 with mean error within 0.26 ref-sd; a hierarchical random-intercept model (23 params) fails outright at 1.9-2.4 with the worst coordinate 4.7-7.2 ref-sd out. Partial pooling's coupled scales defeat a single tail-averaged Gaussian well below the dimension wall, and the docstring now says so. Accuracy falls with N — pareto_k 0.48-1.04 and error up to 1.58 ref-sd over 6 seeds at N=4e5, above 0.7 on three of six — and much faster with dimension: at k=100 it is 2.7-4.3 and 52-59 ref-sd, and nothing raises, because the line search's gradient-finiteness guard turns what used to be a crash into a quiet failure. I do not quote a growth rate; three cells do not pin one. These draws are a proposal, not a posterior; read `pareto_k` on every fit. One more caveat: if the loader transforms its batches, `full_pass` must apply the same transform, or the weights go wrong with no diagnostic.

**The loss series has a documented convention.** `losses[i-1]` is the objective at the position step `i` reached, evaluated on the batch installed after that step — not the value the step's Armijo test accepted, which is measured on the very batch the position was chosen to minimize and reads optimistically low. Measured against the exact full-data objective over 6 seeds x 1200 steps, the recorded series is not detectably biased (mean error +10.2, se 7.2) while the accepted-value alternative is biased at -367.7 (se 9.7). `test_recorded_loss_is_measured_after_the_batch_advance` pins it. This matters because the convergence callback in #733 is the intended consumer of exactly this series.

Out of scope: the exact-full-data-gradient variant, the positions-only `Trajectory` refactor, initial-step scaling. The tail fraction is what I am least sure of: I hard-coded 0.75 rather than add a knob. Sweeping 0.50/0.60/0.75/0.90, the interior is flat and only 1.00 is clearly wrong; the exact medians moved between reruns, so I kept only the claim both reruns agree on.

AI disclosure, per the PyMC GSoC 2026 guidance: I used Claude extensively on this PR, for implementation, test generation, and for re-deriving the claims above. Every number in this description and in the docstrings was verified by running code, and several earlier versions of them were wrong and were corrected or deleted on that basis. All commits except the two original hand-written ones carry a `Co-Authored-By` trailer.

### DIFF
--- a/pymc_extras/inference/pathfinder/__init__.py
+++ b/pymc_extras/inference/pathfinder/__init__.py
@@ -1,4 +1,10 @@
 from pymc_extras.inference.pathfinder.idata import pathfinder_report
 from pymc_extras.inference.pathfinder.pathfinder import fit_blackjax_pathfinder, fit_pathfinder
+from pymc_extras.inference.pathfinder.streaming_pathfinder import fit_streaming_pathfinder
 
-__all__ = ["fit_blackjax_pathfinder", "fit_pathfinder", "pathfinder_report"]
+__all__ = [
+    "fit_blackjax_pathfinder",
+    "fit_pathfinder",
+    "fit_streaming_pathfinder",
+    "pathfinder_report",
+]
--- a/pymc_extras/inference/pathfinder/stochastic_lbfgs.py
+++ b/pymc_extras/inference/pathfinder/stochastic_lbfgs.py
@@ -13,24 +13,17 @@ search remains valid, and forms each curvature pair from a **single minibatch**
 
     y_k = grad_{B_k}(x_{k+1}) - grad_{B_k}(x_k)
 
-Both gradients are evaluated on the same batch, so the same noise enters both and
-cancels in the difference. Differencing across two batches leaves it, and ``y`` then
-estimates nothing in particular: the secant condition ``y ~= H s`` needs both gradients
-to be of the *same* objective. Schraudolph et al. give that form as their eq. (13), the
-one that lets sampling noise into the update, and report divergence below batch 1000 on
-their quadratic model.
-
-The sign violation this produces, ``s . y < 0``, is not the common case at the step lengths
-this loop runs at: over logistic and Gaussian trajectories at batches 4 to 512, same-batch
-pairing produced no negative ``s . y`` on any cell, and cross-batch produced 0.3% to 12% of
-steps. It becomes the common case as the step shrinks, because the cross-batch noise term
-scales with ``||s||`` while the curvature term scales with ``||s||^2``; at a fixed point the
-cross-batch rate is about one in two once ``||s||`` is below 1e-2. Pairs still failing the
-curvature condition are skipped, not forced into the history.
-
-Each accepted step records the ``(x, g, alpha, s_win, z_win)`` that
-``make_pathfinder_sample_fn`` consumes, with the window ordered oldest-to-newest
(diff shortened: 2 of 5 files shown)


YichengYang-Ethan:
Marking this ready, with a short reading guide.

Two files carry the design; read them in this order.

`stochastic_lbfgs.py` — the optimizer core, independent of PyMC. The one invariant everything hangs on: all gradients within a step are evaluated on one minibatch, and `on_batch_advance()` runs only after the curvature pair is formed, so the batch noise cancels inside `y = g_new - g`. Pairs failing the curvature condition are counted and skipped, never forced into the history; a failed line search holds position and retries on fresh data rather than polluting the L-BFGS memory.

`streaming_pathfinder.py` — the driver. It demands `loader.total_size` up front (the importance weights need an exact full-data pass, so there is no honest way to proceed without N), streams the fit, then takes the Polyak-Ruppert tail average of the iterate positions as the center — each accepted iterate is essentially its own minibatch's MAP, so selection cannot help but averaging can. The final exact-logP pass verifies it visited exactly N rows and refuses a short epoch, naming the missing rows.

Tests to spot-check: `test_no_duplicate_gradient_evaluations` pins the two-evaluations-per-step contract by call count; `test_total_size_demand` pins the loader contract; `test_recorded_loss_is_measured_after_the_batch_advance` pins the loss-series convention that #733 consumes.

The PR description states where it degrades (N=4e5 logistic, k=100) as plainly as where it works; those limits are in the docstring too.

PR #725, opened 2026-08-07
Add pymc_extras.terms: composable subgraphs for PyMC + xarray

williambdean:
## Summary

`pymc_extras.terms` is a **light, extendable protocol** for composing model terms with `+`, `*`, and `-` into predictors that materialize inside a `pm.Model` from an `xarray.Dataset`.

The goal is **not** to implement every way to specify a model. It's a small contract — subclass `ModelTerm`, implement five lifecycle methods, compose with built-ins — that lets you lump terms into expressions and lets the framework handle coords, `pmd.Data` registration, tensor construction, and prediction data updates. If a term is missing, you write it; that's the intended path.

This leans on `pytensor` / `pymc.dims` / `xtensor` rather than reinventing op-level subgraph machinery. The gap it fills is specifically PyMC's: coordinates with values, syncing data from an `xr.Dataset`, and reusable prediction-time `set_data`. Pure pytensor composition already exists; this layer is the glue between expressions and PyMC's model/dims/xarray stack.

## Built-ins

| Piece | Role |
|-------|------|
| `Parameter` | Named free RV (required `name`); dims via `prior.dims` |
| `Intercept` | Same as `Parameter`, default `name="intercept"` |
| `Dot` | `data @ coef`; optional `name` (default `{var_name}_beta`) |
| `Transform` | `func(inner)` (e.g. `ptx.math.exp`) |
| `Sum` / `Product` | Results of `+` / `*` |

Helpers: `collect_coords`, `register_data`, `build_param`, `set_data`, `collect_terms`. These **are** the API — no one-shot `materialize` helper that hides the lifecycle.

## Quick example

```python
import pymc.dims as pmd
import pytensor.xtensor as ptx
from pymc_extras.prior import Prior
from pymc_extras.terms import (
    Dot, 
    Intercept, 
    Parameter, 
    Transform, 
    build_param, 
    collect_coords, 
    register_data,
)

mu = Intercept(name="mu") + Dot(var_name="x", prior=Prior("Normal", dims="feature"))
sigma = Transform(Intercept(name="sigma"), func=ptx.math.exp)

coords = collect_coords(mu, sigma, ds=ds)
with pm.Model(coords=coords) as model:
    register_data(mu, ds=ds)
    register_data(sigma, ds=ds)
    pmd.Normal("y_obs", mu=build_param(mu), sigma=build_param(sigma), observed=ds["y"])
```

Dimensional free parameters and shared-data dual coefficients:

```python
mu = Parameter("baseline", prior=Prior("Normal", dims="product")) + Dot(var_name="x", name="alpha_coef", prior=Prior("Normal", dims="feature"))
# second branch on the same data matrix:
# Dot(var_name="x", name="beta_coef", prior=Prior("Normal", dims="feature"))
```

## Extending

Custom terms subclass `ModelTerm` and implement the five lifecycle methods (`get_coords`, `add_coords`, `register_data`, `create_variable`, `set_data`). They compose with built-ins the same way — no framework changes.

The module docstring **Extending** section includes a full custom gather / index recipe and shows it composed with `Parameter`, `Dot`, and `Transform`. That pattern is intentional: missing leaves stay in *your* package.

## Design notes

- **Light protocol, not a DSL** — compose with `+` / `*` / `-`; no meta-model language, no one-shot builder.
- **`Parameter` ≡ `Intercept`** — one free-param concept; Intercept is naming sugar for GLM familiarity.
- **`Dot.name`** — two coefficient vectors on one `pmd.Data` without clashes.
- **Composition correctness** — `Sum.get_coords` / `Sum.set_data` recurse into `Product` children so `a - b` and `a + b * c` collect coords and update data.
- **Tests** — composition, naming, dimmed `Parameter`/`Intercept`, dual `Dot` names; see `tests/test_terms.py`.

## Non-goals

- Not every way to specify a model — a light, extendable protocol, not a complete modeling DSL or full hierarchical framework.
- Not reinventing PyTensor ops / subgraph IR — lean on pytensor + `pymc.dims` + xtensor; this layer is coords + `xr.Dataset` data binding.
- No one-shot `materialize` helper — the lifecycle helpers are the API.
- No shipped gather/index term (example only in docs).
- No domain model wrappers (GLM / demand / etc. classes).
- No change to likelihood / sampling APIs.

### DIFF
--- a/docs/api/terms.rst
+++ b/docs/api/terms.rst
@@ -1,19 +1,27 @@
 Composable model terms
 =====================
 
-Declarative, composable terms for building ``pymc.dims`` subgraphs from
-xarray datasets. Terms compose with ``+``, ``*``, and ``-`` operators
-and handle coordinates, data registration, tensor construction, and
-prediction data updating.
+``pymc_extras.terms`` provides a small, expressive core of building blocks
+for ``pymc.dims`` subgraphs from xarray data.  The built-in terms
+(``Parameter`` / ``Intercept``, ``Dot``, ``Transform``) cover common
+linear-predictor pieces, but **this module is not a full modeling toolkit** ---
+it does not ship complete hierarchical, domain-specific, or
+observation-model wrappers.  It is designed
+to be **extended**: subclass ``ModelTerm`` for domain-specific terms
+(gather/index, media transforms, group effects); they compose with
+built-ins via ``+``, ``*``, and ``-`` and use the same helpers.  See the
+``pymc_extras.terms`` module docstring **Extending** section for a full
+custom-term recipe.
 
 .. currentmodule:: pymc_extras.terms
 .. autosummary::
    :toctree: ../generated/
 
    ModelTerm
+   Parameter
+   Intercept
    Sum
    Product
-   Intercept
    Dot
    Transform
    build_param
--- a/docs/api/terms.rst
+++ b/docs/api/terms.rst
@@ -1,0 +1,24 @@
+Composable model terms
+=====================
+
+Declarative, composable terms for building ``pymc.dims`` subgraphs from
+xarray datasets. Terms compose with ``+``, ``*``, and ``-`` operators
+and handle coordinates, data registration, tensor construction, and
+prediction data updating.
+
+.. currentmodule:: pymc_extras.terms
+.. autosummary::
+   :toctree: ../generated/
+
+   ModelTerm
+   Sum
+   Product
+   Intercept
+   Dot
+   Transform
+   build_param
+   get_coords
+   register_data
+   set_data
+   collect_terms
+   collect_coords
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -10,5 +10,6 @@ API Reference
    api/inference
    api/distributions
    api/statespace
+   api/terms
    api/prior
    api/utils
--- a/pymc_extras/terms.py
+++ b/pymc_extras/terms.py
@@ -97,6 +97,46 @@ Gotchas
           model = pm.modelcontext(None)
           unique = list(dict.fromkeys(ds[self.data_source].values))
           model.add_coords({self.data_source: unique})
+
+Extending
+---------
+``Parameter`` and ``Intercept`` are the same kind of term.  ``Intercept``
+is a convenience alias with ``name`` defaulting to ``"intercept"``; use
+``Parameter`` when the role is not an intercept (e.g. a cohort scale or
+product effect).  When the prior carries ``dims``, coordinates are
+automatically extracted from the dataset during ``collect_coords``.
+
+Custom terms subclass ``ModelTerm`` and implement the five lifecycle
+methods.  Here is an example of a gather / index term that maps a
+cohort-level parameter to customer-level via an index column.  This is
+**not shipped** — copy it into your project as a starting point:
+
+.. code-block:: python
+
+    @dataclass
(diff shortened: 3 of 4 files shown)


ricardoV94:
There was a pymc building blocks channel in the pymc discord and the labs one a while ago. the idea comes up now and again. Last time it got hanged on some people who wanted fancy hierarchical support and others something else, and no clear leadership emerged.

This proposal should engage with that. 

On my side I'd just say there already machinery for defining composable subgraphs: pytensor. And you can build pymc models from them with the fgraph IR. Seeing something like a dot op makes me suspect of wheel reinvention. I didn't evaluate very carefully though.

williambdean:
Totally, understand. There is the chicken or egg issue with implementation needs pymc needing pytensor. I like the idea of being able to use pytensor as much as possible so that ops are not redefined. My goal is to not reinvent as much as possible. However, I would like to be sure to have things that are actually (re)usuable. Thankfully pymc.dims and xtensor makes these wrappers much easier to work with (thank you, Ricardo 😄). But there is still constraints when using for applications that I am not having met which I am trying to fill a bit more here.

williambdean:
I love the idea of just using pytensor. However, then I have to deal with pymc details like coordinates with values and sync up with actually data in some xr.Dataset.

ricardoV94:
Sure but do check those old channels/threads please.

williambdean:
I recall this. https://github.com/pymc-devs/pymc-extras/pull/398 and breadcrumbs from it.
Anything else you are thinking of?

jessegrabowski:
@lucianopaz was also thinking about it quite a bit

lucianopaz:
@williambdean, there was the private [modular-models](https://github.com/pymc-labs/modular-models) project that was later replaced

williambdean:
> @williambdean, there was the private [modular-models](https://github.com/pymc-labs/modular-models) project that was later replaced

Thanks, I am aware. Any commentary on this light protocol?

williambdean:
To be clear, the goal is not implement every type of way to do a model, but just provide a light protocol that's extendable, but allows for some lumping of model terms with some expression of addition and multiplication.

ricardoV94:
@williambdean I think the reason I'm spamming you is that 4-5 people in the past thought they had a nice composable API to do meta-model programming. And without trying to see if they could cover each other needs they could all claim the same you just need without being right.

> provide a light protocol that's extendable

Because the alternative is we end up with 5 different formats for writting 5 different kinds of models. Which then begs the question which one(s) should be in pymc-extras vs specialized packages.

williambdean:
> Because the alternative is we end up with 5 different formats for writting 5 different kinds of models. Which then begs the question which one(s) should be in pymc-extras vs specialized packages.

No worries. I had thought PyMC Extras was a sandbox area, but I can add to either PyMC Marketing experimental module or in my own package.

jessegrabowski:
Extras should be a sandbox

ricardoV94:
> Extras should be a sandbox

We still need to select, we wouldn't implement 5 apis for VI here would we?

jessegrabowski:
Not long term, but if there was a deep uncertainty about how to proceed on the API, or if each API served a different consumer, I wouldn't object to it. The model builder thing is something that has no right answer, if people want to give it a try and see what happens I don't see any issue, and I don't see any reason to harmonize the efforts a priori, because I don't believe there's a right answer.

ricardoV94:
Then go ahead @williambdean

PR #726, opened 2026-08-08
Add Fisher noncentral hypergeometric distribution

kaylimekay:
This PR adds the [Fisher non-central hypergeometric distribution](https://en.wikipedia.org/wiki/Fisher%27s_noncentral_hypergeometric_distribution). It is modeled after the [corresponding PR](https://github.com/pymc-devs/pymc/pull/4249) in the main pymc repo that adds the usual (central) hypergeometric distribution. It incorporates the [scipy implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.nchypergeom_fisher.html#scipy.stats.nchypergeom_fisher) as the random variable.

This addition was first proposed in a [discussion](https://github.com/pymc-devs/pymc/discussions/7909) on the main pymc repo.

This distribution is particularly useful in hit rate analyses when quantifying the odds that a given signal is able to correctly identify some notion of "winners" in data. The author has applied the code in this PR to problems of this type.

### DIFF
--- a/pymc_extras/distributions/discrete.py
+++ b/pymc_extras/distributions/discrete.py
@@ -18,7 +18,11 @@ import pymc as pm
 from pymc.distributions.dist_math import betaln, check_parameters, factln, logpow
 from pymc.distributions.shape_utils import rv_size_is_none
 from pytensor import tensor as pt
+from pytensor.tensor.random.basic import ScipyRandomVariable
 from pytensor.tensor.random.op import RandomVariable
+from pytensor.utils import lazy_scipy_module
+
+stats = lazy_scipy_module("stats")
 
 
 def log1mexp(x):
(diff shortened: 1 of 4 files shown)


pymc_extras/distributions/discrete.py
```diff
@@ -184,6 +188,198 @@ def logp(value, mu, lam):
...
+            - factln(good - value)
+            - factln(fails_value)
+            - factln(bad - fails_value)
+        )
+
+        lower_bound = n - bad
+        lower = pt.switch(pt.gt(lower_bound, 0), lower_bound, 0)
+        upper = pt.switch(pt.lt(good, n), good, n)
+        res = pt.switch(
+            pt.lt(value, lower),
```
ricardoV94, line 334: why aren't these all in one function?

pymc_extras/distributions/discrete.py
```diff
@@ -184,6 +188,198 @@ def logp(value, mu, lam):
...
+                f"FisherNoncentralHypergeometric.logcdf expects a scalar value but received a {np.ndim(value)}-dimensional object."
+            )
+
+        res = pt.switch(
+            pt.lt(value, 0),
+            -np.inf,
+            pt.switch(
+                pt.lt(value, n),
+                pt.logsumexp(
+                    FisherNoncentralHypergeometric.logp(pt.arange(value + 1), good, bad, n, odds),
```
ricardoV94, line 365: Isn't this potentially huge? Should it use a Scan? Is there nothing better? What does scipy do?

tests/distributions/test_discrete.py
```diff
@@ -22,21 +22,31 @@
...
+from pytensor.utils import lazy_scipy_module
 
 from pymc_extras.distributions import (
     BetaNegativeBinomial,
+    FisherNoncentralHypergeometric,
     GeneralizedPoisson,
     Skellam,
 )
 
+stats = lazy_scipy_module("stats")
```
ricardoV94: tests don't need lazy scipy

tests/distributions/test_discrete.py
```diff
@@ -130,6 +140,76 @@ def test_moment(self, mu, lam, size, expected):
...
+            FisherNoncentralHypergeometric,
+            Nat,
+            {
+                "N": N_domain,
+                "k": Nat,
+                "n": Nat,
+                "odds": Domain([0.01, 0.1, 0.9, 1, 1.5, 2, np.inf]),
+            },
+            lambda value, N, k, n, odds: stats.nchypergeom_fisher.logpmf(value, N, k, n, odds),
+            n_samples=-1,
```
ricardoV94: Shouldn't use -1 in the CI, just for you to test locally

ricardoV94 (commented):
Looks good @kaylimekay 

I left some minor notes.

tests/distributions/test_discrete.py
kaylimekay: Right, apologies, this was an oversight.

pymc_extras/distributions/discrete.py
kaylimekay: `_norm` could be inlined and I am happy to do that if you prefer. `_logweight` needs to be called three separate times but I could make it a local inside `logp`.

pymc_extras/distributions/discrete.py
kaylimekay: This whole implementation roughly follows scipy which just [sums](https://github.com/scipy/scipy/blob/791ccf2df0947f370849fb3fe567420fe06a8390/scipy/stats/_distn_infrastructure.py#L3463-L3466) over the PMF to get the CDF. (Note that `_nchypergeom_gen` only implements the [PMF](https://github.com/scipy/scipy/blob/54ef5423f2e4376230ec3bfda6912a07a50958e3/scipy/stats/_discrete_distns.py#L1905). There is no special implementation of the CDF which is just inherited from `rv_discrete`.)

Scipy's PMF calculation itself is taken from a C++ library that computes [logs](https://github.com/scipy/scipy/blob/54ef5423f2e4376230ec3bfda6912a07a50958e3/scipy/stats/biasedurn/fnchyppr.cpp#L379) of the terms in the numerator subtracting a [scale](https://github.com/scipy/scipy/blob/54ef5423f2e4376230ec3bfda6912a07a50958e3/scipy/stats/biasedurn/fnchyppr.cpp#L169) which is the log of the term corresponding to the mean to keep it numerically well behaved. This is the pattern I followed in this implementation with slight differences (I use the mode/support point.)

This line in particular is summing probabilities so it should not be huge. I am not sure if Scan offers some advantage here, but since pytensor has a builtin logsumexp, which is exactly what is needed here, I just reached for that.

kaylimekay:
Just bumping this. I have addressed all review comments.

pymc_extras/distributions/discrete.py
ricardoV94: Scan offers memory savings, since it won't create the whole parange value + 1, which I assume can be pretty large

pymc_extras/distributions/discrete.py
ricardoV94: yes that would be better

PR #728, opened 2026-08-10
Build post-estimation graphs with the fit's filter settings

jessegrabowski:
`forecast`, the conditional samplers and `sample_filter_outputs` built their Kalman graphs with the library defaults for `cov_jitter` and `missing_fill_value` instead of whatever the fit used, so a model with a custom `missing_fill_value` treated real observations as missing. Both are now model-level settings that every shipped model accepts in its constructor; the `build_statespace_graph` arguments still work and override.

### DIFF
--- a/pymc_extras/statespace/core/statespace.py
+++ b/pymc_extras/statespace/core/statespace.py
@@ -125,7 +125,7 @@ class PyMCStateSpace:
     missing_fill_value: float, optional
         Sentinel used to mask missing observations, so PyMC's imputation machinery is not triggered. Set this only
         if your data legitimately contains the default sentinel. Post-estimation graphs are built with this same
-        value. Default -9999.0.
+        value. Default None, which the filter resolves to -9999.0.
 
     Notes
     -----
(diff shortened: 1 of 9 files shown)


pymc_extras/statespace/core/statespace.py
```diff
@@ -244,6 +254,8 @@ def __init__(
         verbose: bool = True,
         measurement_error: bool = False,
         mode: str | None = None,
+        cov_jitter: float = JITTER_DEFAULT,
+        missing_fill_value: float | None = None,
```
Dekermanjian: Jesse, I am confused here. In the docstring it says the default is -9999, but here the default is set to `None`

Dekermanjian (approved):
Other than my confusion about what the default `missing_fill_value` is, this looks good.

pymc_extras/statespace/core/statespace.py
jessegrabowski: yes you're right. I fixed the typing (just int) and set the default to -9999.0. I think I wrote this when I was having a bit of a love affair with the None type -> if branch assignment. Clearly wrong here.

PR #729, opened 2026-08-10
Clean up dead statespace attributes and fix Representation copy aliasing

jessegrabowski:
`PytensorRepresentation.copy()` was `copy.copy` on a `__slots__` object, so a copy shared the original's mutable `_time_varying_names` set — declaring a matrix time-varying on the copy did it on the original too. Latent today since the one caller never declares after copying, but it's a trap.

Everything else is deletion: four methods with no callers anywhere in the package, the tests, the example notebooks, or gEconpy, plus two `BaseFilter` attributes that were written and never read.


Dekermanjian (approved):
This looks good to me, Jesse!

PR #732, opened 2026-08-11
Deprecate observed=None on Prior.create_likelihood_variable

### LINKED ISSUE #731: Prior.create_likelihood_variable(observed=None) fails on the xdist path
`Prior.create_likelihood_variable` accepts `observed=None` on the regular path but raises on the `xdist` path.

```python
import pymc as pm
import pymc.dims as pmd
from pymc_extras.prior import Prior

with pm.Model(coords={"T": range(5)}):
    mu = pmd.Normal("mu", mu=0, sigma=1, dims=("T",))
    Prior("Poisson", dims=("T",)).create_likelihood_variable(
        "y", mu=mu, observed=None, xdist=True
    )
```

```
AttributeError: 'NoneType' object has no attribute 'ndim'
```

The same call with `xdist=False` builds an unobserved `y`. `Censored.create_likelihood_variable` also builds, because it forwards `observed` to `pmd.Censored` rather than storing it in `parameters`.

Cause: `create_likelihood_variable` puts the value in `parameters["observed"]`, and on the xdist branch `_create_parameter` sends every parameter through `_param_value_with_dims`, which reaches `value.ndim` (`prior.py:508` on `main`) for anything that is not already dims-aware and is not a `list | tuple | Number`. `None` is neither.

The case that hits it is a model that builds its likelihood once and samples the prior predictive before any data exists. Found in pymc-labs/pymc-marketing#2771, which works around it with a branch that sets `mu` on a copy and calls `create_variable(name, xdist=True)` directly, skipping the `mu` guard in `create_likelihood_variable` and reimplementing it.

Reproduced on `main`, on the released `v0.14.0`, and on `ee8cc37` (0.12.2.dev1), which all carry the same helper. Traceback above is from pymc 6.0.1 with `ee8cc37`.

anevolbap:
Closes #731

`Prior.create_likelihood_variable("y", mu=mu, observed=None, xdist=True)` raised `AttributeError: 'NoneType' object has no attribute 'ndim'` out of `_param_value_with_dims`, while `Censored` on both paths and `Prior` with `xdist=False` built an unobserved variable.

With `xdist=True` it now raises `ValueError` pointing at `create_variable`. Without `xdist` it still builds but emits a `FutureWarning`, so that path can start raising after a deprecation period.

One test in `TestXDist` covering `Prior` and `Censored`: both raise with `xdist=True`, both warn and still build without it. `tests/test_prior.py` is 111 passed, 2 skipped.

### DIFF
--- a/tests/test_prior.py
+++ b/tests/test_prior.py
@@ -1410,7 +1410,7 @@ class TestXDist:
         assert equivalent_models(obs_m, expected_obs_m)
 
     def test_xdist_likelihood_requires_observed(self):
-        """observed=None is refused on the xdist path, and left alone off it."""
+        """observed=None raises on the xdist path and warns off it."""
 
         coords = {"city": range(3)}
         normal = Prior("Normal", sigma=1, dims=("city",))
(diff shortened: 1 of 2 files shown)


tests/test_prior.py
```diff
@@ -1412,6 +1412,26 @@ def test_xdist_likelihood(self):
 
         assert equivalent_models(obs_m, expected_obs_m)
 
+    def test_xdist_likelihood_requires_observed(self):
+        """observed=None is refused on the xdist path, and left alone off it."""
+        import pymc.dims as pmd
```
ricardoV94: don't use local imports

ricardoV94 (approved):

PR #735, opened 2026-08-14
Stop sharing mutable Kalman filters between graph builds

jessegrabowski:
One `BaseFilter` and one `KalmanSmoother` were shared between the fit graph and every post-estimation graph, and `build_graph` overwrote their settings on every call — so a forecast built after the fit got to decide what the fit had used. Settings now come from the constructor and `build_graph` only reads them, which also retires the `copy()` calls that were containing the damage.

`build_graph` loses its `time_varying_names`, `cov_jitter`, and `missing_fill_value` arguments as a result; callers get a configured pair from `PyMCStateSpace.make_filters()`.


pymc_extras/statespace/core/forward_sampling.py
```diff
@@ -123,6 +123,7 @@ def _sample_conditional(
             if name == "smoothed":
                 # The simulation smoother draws the whole latent path jointly, so the
                 # states carry their cross-time posterior covariance.
+                lgss_filter, lgss_smoother = ss_mod.make_filters()
```
Dekermanjian: This is just a nitpick, but why not unpack this as `kalman_filter, kalman_smoother` so that it is consistent across the codebase and consistent with the arguments of the `SimulationSmoother` below?

Dekermanjian (approved):
This looks good to me!

pymc_extras/statespace/core/forward_sampling.py
jessegrabowski: Agreed, changed

PR #737, opened 2026-08-14
Build the shared-states design matrix explicitly

jessegrabowski:
With shared states, `Cycle`, `TimeSeasonality` and `FrequencySeasonality` built a single `(1, k_states)` design row and let `set_subtensor` broadcast it across the observed series. They build the full matrix now, so the design can be assigned whole and actually gets shape-checked.


ricardoV94 (approved):
We need a nice dims Scan but after that it could be quite helpful for the STS module

PR #739, opened 2026-08-16
Carry the time-varying flag on the matrix instead of in name lists

jessegrabowski:
Which matrices are time-varying was tracked as a set of names on the representation and then hand-carried to everything downstream — `sequence_names` through four call layers, `time_varying_names` into the filter and smoother constructors, and a module-level list in `_forward_simulate_latent_and_obs` to put the scan's arguments back in order. Any transformation that produced a new variable dropped the fact on the floor, so every consumer had to be told again.

Now `declare_time_varying` tags the tensor with a pytensor assumption and consumers ask the matrix. Slicing, substitution, elementwise math, and component composition preserve it; indexing or reducing the time axis clears it.

User-facing API is unchanged — `self.ssm.declare_time_varying("state_cov")` in `make_symbolic_graph` as before. One behavior change: declaring a matrix still at its core rank now raises instead of silently doing nothing, since there's no leading axis to attach the fact to.

### DIFF
--- a/pymc_extras/statespace/core/representation.py
+++ b/pymc_extras/statespace/core/representation.py
@@ -4,7 +4,7 @@ import numpy as np
 import pytensor
 import pytensor.tensor as pt
 
-from pymc_extras.statespace.core.assumptions import declare_time_varying
+from pymc_extras.statespace.core.assumptions import declare_time_varying, is_time_varying
 
 floatX = pytensor.config.floatX
 KeyLike = tuple[str | int, ...] | str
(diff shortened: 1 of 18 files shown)


jessegrabowski:
@ricardoV94 this PR is a bit experimental, would appreciate a real review on it.

pymc_extras/statespace/core/representation.py
```diff
@@ -260,21 +260,33 @@ def _coerce(
 
         return tensor
 
-    @property
-    def time_varying_names(self) -> frozenset[str]:
-        """Names of matrices the model declared as time-varying.
+    def declare_time_varying(self, *names: str) -> None:
```
Dekermanjian, line 263: we import declare_time_varying and then create a function with the same name. Is that intended?

Dekermanjian (approved):
Jesse, I am not well versed on how the pytensor assumptions work. The overall direction makes sense to me and I didn't see anything that was too concerning.

PR #740, opened 2026-08-17
Return substituted statespace matrices instead of storing them

jessegrabowski:
`subbed_ssm` was a single mutable buffer that every post-estimation method overwrote. Each of them substitutes against a throwaway model of `pm.Flat` dummies, so after one `forecast()` the stored matrices pointed at nodes whose owning model had gone out of scope — `unpack_statespace()` was silently order-dependent, valid after `build`, garbage after `build → forecast`. The substitution helpers now return their result and the slot is gone.

`unpack_statespace()` keeps its role as the escape hatch for deriving further quantities from the fitted system, but substitutes afresh on each call:

```python
with pm.Model() as model:
    ...
    ss_mod.build_statespace_graph(data)
    x0, P0, c, d, T, Z, R, H, Q = ss_mod.unpack_statespace()
    pm.Deterministic("T_eigenvalues", pt.linalg.eigvalsh(T))
```

It now requires a model context, and the matrices it returns are new graph nodes rather than the ones the filter was built from. For models whose time axis comes from exogenous data it works before `build_statespace_graph` too; a model that sizes a matrix directly against `n_timesteps` — seasonality with a duration — takes the length off the registered data, and raises naming that if no statespace observation is present.

### DIFF
--- a/notebooks/DFM_Example_(Coincident_Index).ipynb
+++ b/notebooks/DFM_Example_(Coincident_Index).ipynb
@@ -1112,7 +1112,7 @@
     "\n",
     "    # Build symbolic graph\n",
     "    dfm.build_statespace_graph(data=endog)\n",
-    "    matrices = dfm._insert_random_variables()"
+    "    matrices = dfm.unpack_statespace()"
    ]
   },
   {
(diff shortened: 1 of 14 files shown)


Dekermanjian (approved):
Looks good to me!

PR #741, opened 2026-08-18
Take fit state off the statespace model and make rebuilds work

jessegrabowski:
`PyMCStateSpace` cached four things about the fit on itself — the coords, the parameter dims, the observed data and the exogenous data — and every post-estimation method read them back. That made the object single-use: building the same template into a second model silently repointed all four, so an idata sampled from the first model was left paired with a template that had forgotten it. Nothing raised.

Post-estimation now reconstructs what it needs from the template plus the idata it was already handed, and the four attributes are gone. The two sources turn out to be exact complements: the structural coords reach only the template, since they're attached to no sampled or observed variable, and the time axis reaches only the idata.

With no fit state left to protect, re-running `build_statespace_graph` updates the observed data instead of failing on a duplicate `data` name — the re-executed notebook cell this started from. It still raises if the template's parameters aren't in the model, so editing the specification and re-running the cell can't silently inherit the graph that's already there. It no longer accepts `cov_jitter`, `missing_fill_value` or `mvn_method`. The first two are constructor-only now, because a build-time override was written back onto the template and repointed the settings post-estimation reads.

`forecast` and `impulse_response_function` gained a `group` argument, so pass the whole idata and name the group rather than passing `idata.prior`:

```python
forecast = ss_mod.forecast(idata, periods=10, group="prior")
```

A lone group carries no observed data, so there's nothing to recover the time index from.

### DIFF
--- a/pymc_extras/statespace/core/dummy_graph.py
+++ b/pymc_extras/statespace/core/dummy_graph.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
 import pandas as pd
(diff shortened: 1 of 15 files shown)


pymc_extras/statespace/core/dummy_graph.py
```diff
@@ -42,13 +63,17 @@ def infer_variable_shape(name):
         pm.Flat(
             name,
             shape=infer_variable_shape(name),
-            dims=ss_mod._fit_dims.get(name, None),
+            dims=dims.get(name, None),
         )
 
 
 def kalman_filter_outputs_from_dummy_graph(
```
Dekermanjian, line 70: @jessegrabowski, is the indentation here correct?

pymc_extras/statespace/core/forward_sampling.py
```diff
@@ -417,12 +427,16 @@ def sample_statespace_matrices(
         if unknown_matrix_names:
             raise ValueError(f"{sorted(unknown_matrix_names)} not a valid statespace matrix name!")
 
-    with pm.Model(coords=ss_mod._fit_coords) as forward_model:
-        dummy_graph.build_dummy_graph(ss_mod)
+    fit_coords = coords_from_idata(ss_mod, idata, "observed_data")
+
+    with pm.Model(coords=fit_coords) as forward_model:
+        dummy_graph.build_dummy_graph(
+            ss_mod, coords=fit_coords, dims=dims_from_idata(ss_mod, idata, group)
```
Dekermanjian: nitpick, but why not create a fit_dims object like fit_coords instead of passing a function into build_dummy_graph's arguments.

Dekermanjian (approved):

pymc_extras/statespace/core/dummy_graph.py
jessegrabowski: I double checked, it's right. This is a weird display issue

pymc_extras/statespace/core/forward_sampling.py
jessegrabowski: Yep good catch, I changed it.

pymc_extras/statespace/core/dummy_graph.py
Dekermanjian: Okay, that is what I thought. That is really weird.

PR #742, opened 2026-08-19
Fix AttributeError raised when a placeholder is registered twice

jessegrabowski:
`make_and_register_variable` and `make_and_register_data` are supposed to raise `ValueError` when a name is registered twice, naming the shape already registered under it. They read the shape off the `SymbolicVariable`/`SymbolicData` wrapper instead of the tensor inside it, and those dataclasses have no `type`, so you got `AttributeError: 'SymbolicVariable' object has no attribute 'type'` instead of the message.

`Component` has near-copies of both methods with this right, so the original fix landed on one of the two.


Dekermanjian (approved):
Looks good!

PR #747, opened 2026-08-25
Fix use_data_time_dim in unconditional sampling

jessegrabowski:
`sample_unconditional_prior` and `sample_unconditional_posterior` take a `use_data_time_dim` flag that's meant to reuse the time index the model was fit on. It never worked alongside an explicit `steps`. The branches disagreed over whether that time dimension holds `steps` or `steps + 1` entries, so the value the guard demanded crashed downstream on a coord length mismatch, and the value that would actually have worked got rejected by the guard. `steps` is now derived from the data rather than validated against it.

Second, unrelated fix: the numpy simulation helper in the test utilities hardcoded a length-1 mean vector, so no multivariate model with measurement error could be simulated at all. That's why the multivariate structural paths have so little coverage.

### DIFF
--- a/pymc_extras/statespace/core/forward_sampling.py
+++ b/pymc_extras/statespace/core/forward_sampling.py
@@ -267,16 +267,18 @@ def _sample_unconditional(
     fit_dims = dims_from_idata(ss_mod, idata, group)
     temp_coords = fit_coords.copy()
 
-    if not use_data_time_dim and steps is not None:
-        temp_coords.update({TIME_DIM: np.arange(1 + steps, dtype="int")})
-        steps = len(temp_coords[TIME_DIM]) - 1
-    elif steps is not None:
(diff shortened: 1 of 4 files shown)


Labeeb2339 (commented):
I reviewed the `use_data_time_dim` change and the added coverage. Deriving `steps` as the fit time dimension minus the initial state correctly handles the trajectory-versus-coordinate off-by-one, while preserving a plain integer range when the flag is false. The tests cover posterior and prior sampling, implicit and matching explicit horizons, conflicting horizons, and preservation of the fitted time labels. The multivariate measurement-error utility fix also supplies a mean vector with the correct `k_endog` length, matching the covariance shape. I found no additional correctness issue in this diff.

ricardoV94 (approved):

PR #748, opened 2026-08-25
Fix end and start argument handling in forecast

jessegrabowski:
Two bugs in how `forecast` resolves its horizon. `end` meant different things depending on the model's time index — inclusive on a datetime index via `pd.date_range`, exclusive on an integer one via `pd.RangeIndex` — so the same call returned one fewer period on an integer index. `periods` was already consistent across both, and agrees with the datetime `end` behavior, so the integer branch is what moved.

Separately, `start` was only defaulted inside the branch that logs the "no start date provided" warning. Passing `verbose=False`, which the warning itself recommends, left `start` as `None` and forecasting died further down on a `TypeError` about comparing `None` to an int.

The `end` change alters existing behavior: anyone passing `end` on a range-indexed model now gets one more forecast period than before. Three existing test cases had the old semantics baked into them, asking for `end=21` on integer indices against a ten-step `end` on datetime ones, and those are corrected here.

### DIFF
--- a/pymc_extras/statespace/core/forecast.py
+++ b/pymc_extras/statespace/core/forecast.py
@@ -558,11 +558,12 @@ def forecast(
 
     time_index = ss_mod._get_fit_time_index(idata)
 
-    if start is None and verbose:
-        _log.warning(
-            "No start date provided. Using the last date in the data index. To silence this warning, "
-            "explicitly pass a start date or set verbose = False"
-        )
+    if start is None:
+        if verbose:
+            _log.warning(
+                "No start date provided. Using the last date in the data index. To silence this "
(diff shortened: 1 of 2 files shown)


Labeeb2339 (commented):
I reviewed the integer-index and quiet-mode fixes. The `end + 1` change is consistent with the later `forecast_index[1:]` step, so an inclusive endpoint yields the same number of forecast periods as the datetime branch. The new parameterized tests compare the resulting indexes directly and also verify that `verbose=False` suppresses only the warning while still defaulting `start`. I found no additional correctness issue.

ricardoV94 (approved):

PR #749, opened 2026-08-25
Warn when n_steps conflicts with the shock trajectory length

jessegrabowski:
`impulse_response_function` has a guard that's supposed to warn you when you pass both `n_steps` and a `shock_trajectory` whose lengths disagree, since the trajectory wins and `n_steps` gets dropped. The guard compared the trajectory length against itself, so it never fired and the horizon was silently truncated. `n_steps` now defaults to `None`, so a caller-supplied value can be told apart from the default, and the comparison uses what the caller actually passed.

The function's docstring is rewritten to numpydoc style along the way — human-readable types instead of `Optional[Union[...]]`, defaults in the description rather than on the type line, and literals for parameter cross-references. That's the bulk of the diff and it's cosmetic.

This conflicts with #746, which edits the same docstring block to document its new `shock_order` argument. #746 is the larger change and is already green, so merging it first and rebasing this one is the cheaper order.

### DIFF
--- a/pymc_extras/statespace/core/irf.py
+++ b/pymc_extras/statespace/core/irf.py
@@ -28,6 +28,8 @@ if TYPE_CHECKING:
 
 _log = logging.getLogger("pymc.experimental.statespace")
 
+DEFAULT_IRF_STEPS = 40
+
 
 def _shock_permutation(
     shock_names: list[str], shock_order: list[str | EllipsisType] | None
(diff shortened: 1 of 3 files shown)


Labeeb2339 (commented):
I reviewed the horizon handling and the added VARMAX coverage. The implementation now distinguishes an omitted `n_steps` from an explicit conflicting value, compares against the trajectory's first dimension, and preserves the documented trajectory-wins behavior. The tests cover conflicting, matching, omitted, and default-horizon cases, and the PR's full matrix is green. I found no additional code concern.

ricardoV94 (approved):

Dekermanjian (approved):
Looks good!

PR #753, opened 2026-08-29
Fix three cases where ADVI silently ignored the caller

jessegrabowski:
Three places where ADVI silently did something other than what the caller asked: a dropped optimizer buffer, a stale minibatch, and an ignored `observeds`.


zaxtax (approved):
Great catch!

PR #756, opened 2026-08-30
Return fit and optimizer_result groups from fit_advi

jessegrabowski:
`fit_advi` returned a `fit` group holding nothing but an `elbo` vector. It now reports the guide's mean and covariance there, plus an `optimizer_result` group with the trace and the optimizer's state. That's a step toward the idata holding the full optimizer state so a run can be warm-started or restarted from it, and part of a broader effort to remove state from `Trainer`.

### DIFF
--- a/pymc_extras/inference/advi/autoguide.py
+++ b/pymc_extras/inference/advi/autoguide.py
@@ -92,8 +92,7 @@ class AutoGuideModel:
         """Describe the fitted approximation by its mean and its covariance parameterization.
 
         Every guide reports ``mean_vector``, so the family is identifiable from whichever
-        covariance entries accompany it. The mean-field guide reports a marginal standard
-        deviation, exponentiated out of the unconstrained space it optimizes in.
+        covariance entries accompany it. Each guide family implements this.
 
         Parameters
         ----------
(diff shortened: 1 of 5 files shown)


ricardoV94 (commented):
Does this increase the burden of writing a custom guide?

jessegrabowski:
Slightly. The guide will need to tell the idata how it stores its covariance terms, and how to reconstruct the full covariance matrix.

ricardoV94:
Could we make it optional?

jessegrabowski:
What would be the default? Just the generic state object we have now?

ricardoV94:
> What would be the default? Just the generic state object we have now? 

The best we can do with the current imposed structure? If it's what we had sure?

jessegrabowski:
Oh I think I see what you mean by optional. We can have a good default that makes it the same difficulty to extend as we have now, just stick what we currently return into the datatree instead of into a random object.

zaxtax:
It would be nice if the optimization groups mirrored the sampler groups. So something like fit_stats along with fit.

I'm also less confident that covariance_dims makes sense as an abstraction. Especially as it seems to be a thing only advi serves

Issue #745, opened 2026-08-23
StandardFilter's log-likelihood omits the k_endog factor on log(2*pi)


jessegrabowski:
`StandardFilter` adds a single `log(2*pi)` per timestep instead of one per observed series, so reported log-likelihoods are off by `n * (k_endog - 1) * log(2*pi) / 2` whenever `k_endog > 1`. It's additive so posteriors are unaffected, but the values don't match scipy or statsmodels, and `ConvergentFilter`'s tail step carries the same expression.

```python
import numpy as np
import pytensor
import pytensor.tensor as pt
from scipy.stats import multivariate_normal

from pymc_extras.statespace.filters import StandardFilter

rng = np.random.default_rng(0)
k_states, k_endog, n = 3, 2, 20

T, R, Q = np.eye(k_states) * 0.9, np.eye(k_states), np.eye(k_states) * 0.5
Z, H = rng.normal(size=(k_endog, k_states)), np.eye(k_endog) * 0.3
a0, P0 = np.zeros(k_states), np.eye(k_states)
c, d = np.zeros(k_states), np.zeros(k_endog)
data = rng.normal(size=(n, k_endog))

args = [data, a0, P0, c, d, T, Z, R, H, Q]
out = StandardFilter().build_graph(*[pt.as_tensor(x) for x in args])
y_hat, F, loglike = pytensor.function([], [out[2], out[5], out[-1]])()

# the filter's own one-step forecasts, scored with scipy
expected = sum(multivariate_normal(y_hat[t], F[t]).logpdf(data[t]) for t in range(n))

print(f"reported {loglike.sum():.6f}")                                  # -34.763349
print(f"expected {expected:.6f}")                                       # -53.142119
print(f"off by   {loglike.sum() - expected:.6f}")                       #  18.378771
print(f"const    {n * (k_endog - 1) * np.log(2 * np.pi) / 2:.6f}")      #  18.378771
```

PR #757, opened 2026-08-30
[FIX] Fix multivariate SquareRootFilter log-likelihood. Targets #744

### LINKED ISSUE #744: SquareRootFilter reports the wrong log-likelihood when k_endog > 1
`SquareRootFilter`'s per-step loglike uses the wrong transpose in the quadratic form and multiplies the log-determinant by `k_endog`, so it's wrong for any model with more than one observed series, and since both terms move with the parameters, so is the posterior.

```python
import numpy as np
import pytensor
import pytensor.tensor as pt
from scipy.stats import multivariate_normal

from pymc_extras.statespace.filters import SquareRootFilter

k_states, k_endog, n = 3, 2, 20  # k_endog = 1 agrees; anything larger does not

T, R, Q = np.eye(k_states) * 0.9, np.eye(k_states), np.eye(k_states) * 0.5
Z, H = rng.normal(size=(k_endog, k_states)), np.eye(k_endog) * 0.3
a0, P0 = np.zeros(k_states), np.eye(k_states)
c, d = np.zeros(k_states), np.zeros(k_endog)
data = rng.normal(size=(n, k_endog))

# SquareRootFilter takes the Cholesky factor of P0, not P0 itself
args = [data, a0, np.linalg.cholesky(P0), c, d, T, Z, R, H, Q]
out = SquareRootFilter().build_graph(*[pt.as_tensor(x) for x in args])
y_hat, F, loglike = pytensor.function([], [out[2], out[5], out[-1]])()

# the filter's own one-step forecasts, scored with scipy
expected = sum(multivariate_normal(y_hat[t], F[t]).logpdf(data[t]) for t in range(n))

print(f"reported {loglike.sum():.6f}")  # -41.378252
print(f"expected {expected:.6f}")       # -53.142119
```

4nmus:
Fixes the multivariate log-likelihood calculation in SquareRootFilter.
For k_endog > 1implementation had two issues:
    The Mahalanobis term usedCholesky factor, not  v.T @ F^-1 @ v = ||L^-1 v||².
    The log-determinant term was multiplied by k_endog.
My fix reuses the scaled residual L^-1 v from the filtering step to compute the quadratic term and corrects normalization.

Additionally, added a multivariate regression test that compares SquareRootFilter log-likelihood against computed multivariate Gaussian log-density.

For reference, reproducer from #744:
 Before: -41.378252 (expected: -53.142119)
After: -53.142119

Cases of k_endog = 1behavior remains unchanged.

All tests passed on my end.
Closes #744

### DIFF
--- a/tests/statespace/filters/test_kalman_filter.py
+++ b/tests/statespace/filters/test_kalman_filter.py
@@ -247,24 +247,6 @@ def test_output_with_multiple_observed(filter_name, rng):
         )
 
 
-def test_square_root_filter_multivariate_log_likelihood(rng):
-    p, m, r, n = 2, 3, 3, 20
-    inputs = make_test_inputs(p, m, r, n, rng)
-
-    univariate_outputs = get_filter_function("UnivariateFilter")(*inputs)
-
-    square_root_inputs = list(inputs)
-    square_root_inputs[2] = np.linalg.cholesky(square_root_inputs[2])
-    square_root_outputs = get_filter_function("CholeskyFilter")(*square_root_inputs)
-
-    assert_allclose(
-        square_root_outputs[-1],
-        univariate_outputs[-1],
-        atol=ATOL,
-        rtol=RTOL,
-    )
-
-
 @pytest.mark.parametrize("filter_name", filter_names)
 @pytest.mark.parametrize("p", [1, 5], ids=["univariate (p=1)", "multivariate (p=5)"])
 def test_missing_data(filter_name, p, rng):
@@ -378,13 +360,7 @@ def get_loglike_function(filter_name: str, obs_cov: str) -> Callable:
     [
         ("StandardFilter", "dense"),
         ("UnivariateFilter", "diagonal"),
-        pytest.param(
-            "CholeskyFilter",
-            "dense",
-            marks=pytest.mark.xfail(
-                reason="SquareRootFilter scales log |F| by k_endog as well, see issue #744"
-            ),
-        ),
+        ("CholeskyFilter", "dense"),
     ],
 )
 def test_loglike_matches_statsmodels(filter_name, obs_cov, rng):
--- a/tests/statespace/filters/test_kalman_filter.py
+++ b/tests/statespace/filters/test_kalman_filter.py
@@ -247,34 +247,22 @@ def test_output_with_multiple_observed(filter_name, rng):
         )
 
 
-def test_square_root_filter_multivariate_log_likelihood():
-    rng = np.random.default_rng(0)
-    k_states, k_endog, n = 3, 2, 20
+def test_square_root_filter_multivariate_log_likelihood(rng):
+    p, m, r, n = 2, 3, 3, 20
+    inputs = make_test_inputs(p, m, r, n, rng)
 
-    T = np.eye(k_states, dtype=floatX) * 0.9
-    R = np.eye(k_states, dtype=floatX)
-    Q = np.eye(k_states, dtype=floatX) * 0.5
-    Z = rng.normal(size=(k_endog, k_states)).astype(floatX)
-    H = np.eye(k_endog, dtype=floatX) * 0.3
-    a0 = np.zeros(k_states, dtype=floatX)
-    P0 = np.eye(k_states, dtype=floatX)
-    c = np.zeros(k_states, dtype=floatX)
-    d = np.zeros(k_endog, dtype=floatX)
-    data = rng.normal(size=(n, k_endog)).astype(floatX)
+    univariate_outputs = get_filter_function("UnivariateFilter")(*inputs)
 
-    args = [data, a0, np.linalg.cholesky(P0), c, d, T, Z, R, H, Q]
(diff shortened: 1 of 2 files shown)


tests/statespace/filters/test_kalman_filter.py
```diff
@@ -246,6 +246,36 @@ def test_output_with_multiple_observed(filter_name, rng):
...
+    T = np.eye(k_states, dtype=floatX) * 0.9
+    R = np.eye(k_states, dtype=floatX)
+    Q = np.eye(k_states, dtype=floatX) * 0.5
+    Z = rng.normal(size=(k_endog, k_states)).astype(floatX)
+    H = np.eye(k_endog, dtype=floatX) * 0.3
+    a0 = np.zeros(k_states, dtype=floatX)
+    P0 = np.eye(k_states, dtype=floatX)
+    c = np.zeros(k_states, dtype=floatX)
+    d = np.zeros(k_endog, dtype=floatX)
+    data = rng.normal(size=(n, k_endog)).astype(floatX)
```
jessegrabowski: There should be a helper function to build all of this

tests/statespace/filters/test_kalman_filter.py
```diff
@@ -246,6 +246,36 @@ def test_output_with_multiple_observed(filter_name, rng):
...
+    y_hat, F, loglike = pytensor.function([], [outputs[2], outputs[5], outputs[-1]])()
+
+    residual = data - y_hat
+    solved_residual = np.linalg.solve(F, residual[..., None])[..., 0]
+    sign, logdet = np.linalg.slogdet(F)
+    expected_loglike = -0.5 * (
+        k_endog * np.log(2 * np.pi) + logdet + np.einsum("ni,ni->n", residual, solved_residual)
+    )
+
+    assert np.all(sign > 0)
```
jessegrabowski: why?

tests/statespace/filters/test_kalman_filter.py
```diff
@@ -246,6 +246,36 @@ def test_output_with_multiple_observed(filter_name, rng):
...
+    args = [data, a0, np.linalg.cholesky(P0), c, d, T, Z, R, H, Q]
+    outputs = SquareRootFilter().build_graph(*[pt.as_tensor(x) for x in args])
+    y_hat, F, loglike = pytensor.function([], [outputs[2], outputs[5], outputs[-1]])()
+
+    residual = data - y_hat
+    solved_residual = np.linalg.solve(F, residual[..., None])[..., 0]
+    sign, logdet = np.linalg.slogdet(F)
+    expected_loglike = -0.5 * (
+        k_endog * np.log(2 * np.pi) + logdet + np.einsum("ni,ni->n", residual, solved_residual)
+    )
```
jessegrabowski: not in love with this. Can we just assert that the logp of the square root filter is the same as the logp from the standard filter?

jessegrabowski (changes requested):

tests/statespace/filters/test_kalman_filter.py
4nmus: Okay, i will  switch the setup to the existing `make_test_inputs` helper.

tests/statespace/filters/test_kalman_filter.py
4nmus: I think `StandardFilter` currently has the separate multivariate normalization bug tracked in #745 , so a corrected SquareRootFilter won’t match it for k_endog > 1. I can probably  change it to standard but it wouldn't work i suppose

tests/statespace/filters/test_kalman_filter.py
4nmus: This was guarding that the covariance determinant is positive

4nmus:
Thank you for the review! I hope I answered your questions about my implementation. I can simplify the regression test to use the existing test helpers and remove Gaussian calculation. The only issue is with StandardFilter  that it can’t currently serve as the multivariate reference because of #745.

tests/statespace/filters/test_kalman_filter.py
jessegrabowski: its just testing the test. This value comes from handrolled numpy code, not from the filter itself.

4nmus: Yep, that makes sense. I removed it. The regression test now compares the SquareRootFilter directly against another filter implementation (hope this is the expected behavior).

tests/statespace/filters/test_kalman_filter.py
jessegrabowski: I merged https://github.com/pymc-devs/pymc-extras/pull/759 that closes #745. It includes a test with an xfail case for SquareRootFilter, can you rebase your PR and just remove that xfail on that one case? That will be sufficient for testing.

4nmus: Done. Also, there was an issue with the implementation but now it passes for updated tests.

jessegrabowski (approved):
Looks great. Sorry this took so long

PR #759, opened 2026-09-04
Fix the normalizing constant in the Kalman filter log-likelihood

jessegrabowski:
Scale standard scaler logp by the number of observed states to match the actual formula. 

No change to sampler behavior because it was just a normalizing constant. But it will make all the filters comparable and gives us the same answer as statsmodels. Also adds a test that we match statsmodels' implementation.

### DIFF
--- a/pymc_extras/statespace/filters/kalman_filter.py
+++ b/pymc_extras/statespace/filters/kalman_filter.py
@@ -279,7 +279,7 @@ class BaseFilter(ABC):
 
     def handle_missing_values(
         self, y, Z, H, d
-    ) -> tuple[TensorVariable, TensorVariable, TensorVariable, TensorVariable, float]:
+    ) -> tuple[TensorVariable, TensorVariable, TensorVariable, TensorVariable, TensorVariable]:
         """
         Handle missing values in the observation data ``y``.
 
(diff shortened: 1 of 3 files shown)


ricardoV94 (approved):

pymc_extras/statespace/filters/kalman_filter.py
```diff
@@ -806,11 +803,7 @@ def kalman_step(self, *args):
         P_filtered = stabilize(0.5 * (P_filtered + P_filtered.mT), self.cov_jitter)
         a_hat, P_hat = self.predict(a=a_filtered, P=P_filtered, c=c, T=T, R=R, Q=Q)
 
-        ll = pt.switch(
-            all_nan_flag,
-            0.0,
-            -0.5 * ((pt.neq(ll_inner, 0).sum()) * MVN_CONST + ll_inner.sum()),
-        )
+        ll = -0.5 * (pt.bitwise_not(nan_mask).sum() * MVN_CONST + ll_inner.sum())
```
Dekermanjian, line 806: @jessegrabowski, is `ll_inner.sum()` guaranteed to be 0 if `nan_mask` are all 1s?

Dekermanjian (approved):

pymc_extras/statespace/filters/kalman_filter.py
jessegrabowski: Yes, i went back and double checked.

Dekermanjian: Thank you, Jesse! I figured that was the case, I just couldn't track it in the code myself.

PR #762, opened 2026-09-06
Fix #736: Broadcast time-invariant matrices in _sample_conditional

### LINKED ISSUE #736: Conditional sampling breaks when a statespace matrix is assigned whole instead of by index
`SequenceMvNormal` builds its inner `multivariate_normal` against dummy `mus.type()` variables, so when `inline_ofg_expansion` expands the `KalmanFilterRV`, the inner RV gets a `size` taken from a core dim instead of the time axis — `ShapeOpt` happens to repair it and `canonicalize` undoes the repair, so whether you see it depends on how the surrounding matrices were built.

```python
import numpy as np
import pandas as pd
import pymc as pm
import pytensor.tensor as pt

from pymc_extras.statespace.core.statespace import PyMCStateSpace

K = 3


class MinimalVAR(PyMCStateSpace):
    def __init__(self):
        super().__init__(k_endog=K, k_states=K, k_posdef=K)

    def make_symbolic_graph(self) -> None:
        self.ssm["design"] = np.eye(K, K)
        # workaround: self.ssm[("design", *np.diag_indices(K))] = 1

        self.ssm["selection"] = np.eye(K, K)
        self.ssm["initial_state", :] = self.make_and_register_variable("x0", shape=(K,))
        self.ssm["initial_state_cov", :, :] = self.make_and_register_variable("P0", shape=(K, K))
        self.ssm["transition"] = self.make_and_register_variable("ar_params", shape=(K, K))
        self.ssm["state_cov", :, :] = self.make_and_register_variable("state_cov", shape=(K, K))
        self.ssm["obs_cov"] = pt.diag(self.make_and_register_variable("sigma_obs", shape=(K,)))

    @property
    def param_names(self):
        return ["x0", "P0", "ar_params", "state_cov", "sigma_obs"]

    @property
    def state_names(self):
        return [f"state_{i}" for i in range(K)]

    @property
    def observed_states(self):
        return [f"state_{i}" for i in range(K)]

    @property
    def shock_names(self):
        return [f"shock_{i}" for i in range(K)]


data = pd.DataFrame(rng.normal(size=(50, K)), columns=[f"y{i}" for i in range(K)])
mod = MinimalVAR()

with pm.Model(coords=mod.coords):
    pm.Normal("x0", dims=["state"])
    pm.Deterministic("P0", pt.eye(K))
    pm.Normal("ar_params", shape=(K, K))
    pm.Deterministic("state_cov", pt.eye(K))
    pm.Exponential("sigma_obs", 1, shape=(K,))
    mod.build_statespace_graph(data)
    idata = pm.sample_prior_predictive(draws=5, random_seed=0)

mod.sample_conditional_prior(idata, random_seed=0)  # ValueError: operands could not be broadcast
```

dhairya-motta:
If d (obs_intercept) and/or H (obs_cov) are not time-varying, PyTensor's local_add_of_sparse_write transformation erroneously collapses the time dimension while constructing the graph, triggering an AssertionError from SpecifyShape.

Solution: Ensure that d and H have dimensions (steps, ... ) by explicit broadcasting prior to their usage in obs_mu and obs_cov.

Fixes pymc-devs/pymc-extras#736

### DIFF
--- a/pymc_extras/statespace/core/forward_sampling.py
+++ b/pymc_extras/statespace/core/forward_sampling.py
@@ -119,6 +119,15 @@ def _sample_conditional(
             else (None, None)
         )
 
+        # Explicitly broadcast time-invariant observation matrices to avoid a PyTensor
+        # shape contraction triggered by the local_add_of_sparse_write graph rewrite.
+        # See: https://github.com/pymc-devs/pymc-extras/issues/736
+        steps = len(fit_coords[TIME_DIM]) if TIME_DIM in fit_coords else 1
+        if d.ndim == 1:
+            d = pt.broadcast_to(d, (steps, *tuple(d.type.shape)))
+        if H.ndim == 2:
+            H = pt.broadcast_to(H, (steps, *tuple(H.type.shape)))
+
         for name, (mu, cov) in zip(FILTER_OUTPUT_TYPES, grouped_outputs, strict=True):
             if name == "smoothed":
                 # The simulation smoother draws the whole latent path jointly, so the
--- a/tests/statespace/core/test_statespace.py
+++ b/tests/statespace/core/test_statespace.py
@@ -501,3 +501,58 @@ def test_post_estimation_leaves_template_matrices_alone(
 
     assert all(x is y for x, y in zip(before, after, strict=True))
     assert not set(pymc_mod.basic_RVs).intersection(ancestors(after))
+
+
+@pytest.mark.filterwarnings("ignore:No time index found on the supplied data.")
+def test_sample_conditional_with_whole_matrix_assignment(rng):
+    """Regression test for #736.
+
+    When time-invariant `H` (obs_cov) or `d` (obs_intercept) are assigned as
+    whole matrices (ndim 2 and 1 respectively), PyTensor's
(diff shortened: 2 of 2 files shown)


ricardoV94:
You claimed a PyTensor rewrite bug, so the fix should be in pytensor, not working around the bug here

dhairya-motta:
thanks for the review, i agree the actual fix belongs to PyTensor but since it takes time for PyTensor to release cycle update, itll affect pymc users, so would you mind leaving the current patch in pymc-extra in the meantime i can go open another issue and PR in PyTensor aswell, let me know what you think?

ricardoV94:
It takes as little time to release as we need since PyTensor is our package

dhairya-motta:
ill close this pr here and open one directly in PyTensor, thanks for the guidance.

dhairya-motta:
@ricardoV94 i looked into PyTensor to open the pr and i think this issue is already resolved in https://github.com/pymc-devs/pytensor/pull/2350, i verified that the minimal reproducer passes cleanly with that fix in place... correct me if ive misunderstood anything here.

ricardoV94:
that's good news.

PR #767, opened 2026-09-10
Smooth with a Durbin-Koopman disturbance smoother by default

jessegrabowski:
Added `DisturbanceSmoother`, a Durbin-Koopman disturbance smoother, and made it the default through a new `smoother_type` argument on `PyMCStateSpace` and the model builders. It computes the same smoothed moments as the RTS form but inverts only the innovation covariance, which is `k_endog` square. The RTS recursion inverts the `k_states` state covariance at every step and needs a pseudo-inverse whenever a state carries no process noise.

Smoother pass alone, over the same filter output. `k` is `k_states`, `p` is `k_endog`, `T` the series length. float64, warm compile cache, evaluation time only.

```
-------------------------------------------- benchmark: 10 tests ------------------------------------------------
Name (time in ms)                                          Min            Median          StdDev  Rounds  Speedup
-----------------------------------------------------------------------------------------------------------------
test_smoother[k=5-T=240-disturbance]             0.5986 (1.00)     0.6048 (1.00)   0.0137 (1.15)       5  1.9x
test_smoother[k=5-T=240-rts]                     1.1181 (1.87)     1.1331 (1.87)   0.0223 (1.87)       5
test_smoother[k=13-T=240-disturbance]            0.8598 (1.44)     0.8730 (1.44)   0.0119 (1.00)       5  6.1x
test_smoother[k=13-T=240-rts]                    5.0948 (8.51)     5.3608 (8.86)  0.1839 (15.43)       5
test_smoother[k=26-T=240-disturbance]            2.0806 (3.48)     2.1376 (3.53)  0.2888 (24.23)       5  6.1x
test_smoother[k=26-T=240-rts]                  13.0039 (21.72)   13.0171 (21.52)  0.1221 (10.25)       5
test_smoother[k=13-T=1000-disturbance]           3.5998 (6.01)     3.7178 (6.15)   0.0530 (4.45)       5  5.7x
test_smoother[k=13-T=1000-rts]                 20.5360 (34.31)   21.0227 (34.76)  0.4982 (41.80)       5
test_smoother[k=13-p=3-T=240-disturbance]        1.1837 (1.98)     1.2027 (1.99)   0.0278 (2.34)       5  3.9x
test_smoother[k=13-p=3-T=240-rts]                4.6090 (7.70)     4.7011 (7.77)   0.0673 (5.65)      49
-----------------------------------------------------------------------------------------------------------------
```

The parenthesised ratios are pytest-benchmark's own, relative to the fastest row. The `Speedup` column is `rts median / disturbance median` within each pair. The gap narrows as `k_endog` grows because the disturbance form's own inverse scales with it.

`KalmanSmoother` is renamed `RTSSmoother` and stays available as `smoother_type="rts"`. Its `build_graph` now takes `(data, matrices, filter_outputs)` rather than `(T, R, Q, filtered_states, filtered_covariances)`, so the two smoothers are interchangeable. Anyone constructing a smoother directly has to update that call.

`RTSSmoother` also honors `cov_jitter` now. It applied `JITTER_DEFAULT` on a second `stabilize` call regardless of the setting, so `cov_jitter=0.0` never disabled jitter in the smoother.

The two smoothers are checked against each other with `P` rank-deficient and with missing observations, at zero jitter, which is the case the pseudo-inverse existed for.


pymc_extras/statespace/filters/kalman_smoother.py
```diff
@@ -144,3 +171,131 @@ def predict(self, a, P, T, R, Q):
...
+        self.cov_jitter = JITTER_DEFAULT if cov_jitter is None else cov_jitter
+        self.missing_fill_value = MISSING_FILL if missing_fill_value is None else missing_fill_value
+
+        time_varying_short = {LONG_NAME_TO_SHORT[name] for name in time_varying_names}
+        self.seq_names = [name for name in DISTURBANCE_PARAM_NAMES if name in time_varying_short]
+        self.non_seq_names = [
+            name for name in DISTURBANCE_PARAM_NAMES if name not in time_varying_short
+        ]
+
+    def unpack_args(self, args):
```
Dekermanjian: It is a bit ambiguous how args needs to be passed in here.

pymc_extras/statespace/filters/kalman_smoother.py
```diff
@@ -144,3 +171,131 @@ def predict(self, a, P, T, R, Q):
...
+            for name in DISTURBANCE_PARAM_NAMES:
+                if name in self.seq_names:
+                    matrices.append(seqs[self.seq_names.index(name)])
+                else:
+                    matrices.append(non_seqs[self.non_seq_names.index(name)])
+
+        d, T, Z, H = matrices
+        return y, a, P, r, N, d, T, Z, H
+
+    def smoother_step(self, *args):
```
Dekermanjian: Same thing here. You really need to know the right order for the function to work properly.

Dekermanjian (approved):
Looks good Jesse! I just had a couple of nitpicks; feel free to ignore them.

pymc_extras/statespace/filters/kalman_smoother.py
jessegrabowski: good point!

jessegrabowski: I made some changes, let me know what you think. There's always going to be some ambiguity here, because we don't know the order the arguments are going to come to us ahead of time. It depends on what is time varying and not.

Dekermanjian: Yeah, that makes sense and also you were able to cut down a whole method out of both smoothers with the changes you made. I think it looks great the way you have it.

jessegrabowski: stop it i'll blush

PR #769, opened 2026-09-11
Fold the simulation smoother's two smoother passes into one

jessegrabowski:
`SimulationSmoother` now filters and smooths the difference `y - y_plus` once, with the state and observation intercepts zeroed, and returns `alpha_plus + smooth(y - y_plus)`. The filter and smoother are affine in the data, so that equals `alpha_plus - smooth(y_plus) + smooth(y)`, which took two passes. The op takes `data` in place of `a_smooth`, since the real-data smoothed mean no longer appears in the identity, and `_sample_conditional` passes the model's observed data.

Smoothed-state draw compiled on its own, `k` states over `T` observations. float64, warm compile cache, evaluation time only.

```
------------------------------------------- benchmark: 8 tests --------------------------------------------
Name (time in ms)                                    Min            Median          StdDev  Rounds  Speedup
-----------------------------------------------------------------------------------------------------------
test_smoothed_draw[k=4-T=240-one-pass]     1.2889 (1.00)     1.3168 (1.00)   0.0373 (1.00)      20  1.58x
test_smoothed_draw[k=4-T=240-main]         2.0535 (1.59)     2.0829 (1.58)   0.0388 (1.04)      20
test_smoothed_draw[k=13-T=240-one-pass]     5.4760 (4.25)     5.6522 (4.29)   0.1335 (3.58)      20  1.89x
test_smoothed_draw[k=13-T=240-main]       10.3865 (8.06)    10.6903 (8.12)   0.2363 (6.33)      20
test_smoothed_draw[k=26-T=240-one-pass]   16.0702 (12.47)   16.4842 (12.52)   0.2019 (5.41)      20  1.95x
test_smoothed_draw[k=26-T=240-main]      31.5207 (24.46)   32.1064 (24.38)   0.3347 (8.97)      20
test_smoothed_draw[k=13-T=1000-one-pass]   23.1487 (17.96)   23.5227 (17.86)   0.2626 (7.04)      20  1.90x
test_smoothed_draw[k=13-T=1000-main]     43.5898 (33.82)   44.7507 (33.98)  0.6152 (16.48)      20
-----------------------------------------------------------------------------------------------------------
```

The parenthesised ratios are pytest-benchmark's own, relative to the fastest row. The `Speedup` column is `main median / one-pass median` within each pair. On its own the smoothed draw drops the outer filter as well as the outer smoother, which is why it approaches 2x. Through `sample_conditional_posterior`, where the filtered and predicted draws still need the outer filter, the same `k=13, T=240` model goes from 18.14 to 13.84 ms per draw.

Added `joint_smoothed_draws` to `PyMCStateSpace` and the model builders, default `True`. With `False`, the conditional sampling methods draw each smoothed timestep from its own marginal through `SequenceMvNormal`, which reproduces the per-step moments but not the cross-time covariance. It is slower, not faster, because it factors a `k_states` covariance at every timestep. It exists to compare against.

The one-pass identity was checked against the two-pass form with nonzero `x0`, `c` and `d` and agrees to machine precision, and the existing simulation smoother tests, including the comparison against statsmodels, pass with only the constructor argument changed.


ricardoV94:
> With False, the conditional sampling methods draw each smoothed timestep from its own marginal through SequenceMvNormal, which reproduces the per-step moments but not the cross-time covariance. It is slower, not faster, because it factors a k_states covariance at every timestep. It exists to compare against.

Do I read that we have a flag for a slower, leas precise method just for testing purposes? Sounds dumb?

ricardoV94 (approved):

jessegrabowski:
> Do I read that we have a flag for a slower, leas precise method just for testing purposes? Sounds dumb?

It's a famous algorithm. We could maybe get rid of it I agree, but I figured it's there so w/e

PR #770, opened 2026-09-11
Move BayesianVARMAX exogenous regressors into the state equation

jessegrabowski:
`BayesianVARMAX` now puts exogenous regressors in the VAR equation, `y_t = A(L) y_{t-1} + B x_t + e_t`, instead of the observation intercept. That is Lutkepohl's VARX and what statsmodels' `VARMAX` does. `beta_exog` becomes an impact effect that the autoregressive dynamics propagate, so existing fits with regressors will change.

`exog_in_observation=True` on `BayesianVARMAX` and on `StationaryVAR` restores the previous regression-with-VAR-errors model. `BayesianSARIMAX` keeps that placement unconditionally, matching statsmodels' `SARIMAX`.

Both classes are checked against `statsmodels` log-likelihoods with regressors at fixed parameters, which pins the one-step timing of the state intercept.


pymc_extras/statespace/filters/distributions.py
```diff
@@ -712,39 +712,76 @@ def _lag_matrix(series, *, lags, order, n_rows):
...
+
+    if exog_in_observation:
+        return exog[:order] @ exog_coefficients.T
+
+    by_lag = coefficients.reshape((k_endog, order, k_endog))
+    intercepts = exog[1 : order + 1] @ exog_coefficients.T
+    initial = pt.linalg.solve(pt.eye(k_endog) - by_lag.sum(axis=1), intercepts[0], b_ndim=1)
+
+    history = [initial] * order
+    for t in range(1, order):
```
Dekermanjian, line 768: How big do you expect order to get in a typical model? Is the loop okay here?

Dekermanjian (approved):
Looks good. I will defer to you about the loop vs scan on the `order` parameter.

pymc_extras/statespace/filters/distributions.py
jessegrabowski: Good question and good catch. I want to improve support for huge VARs so this should be as robust as possible. I'll do some experiments.

jessegrabowski: I timed it. The answer is that on really big models, we get long compile times due to the loop. I switched to scan as a result. Here's a table!

```
k_endog    order     before      after 
10            12      3.3 s      0.3 s 
20            24     31.6 s      0.2 s
```

I want you to know I typed that myself. It's a 100% artisinal human table.

Dekermanjian: That is a really big difference! But wait a second how is the bigger model faster than the smaller one?

jessegrabowski: it's probably the same, just cpu noise.

jessegrabowski: Or the first run cached some functions the 2nd one used. I didn't do a careful benchmark, don't take it too seriously.

PR #771, opened 2026-09-11
Add a deterministic trend to BayesianVARMAX and BayesianSARIMAX

jessegrabowski:
`BayesianVARMAX` and `BayesianSARIMAX` take a `trend` argument, `None | "n" | "c" | "ct" | "ctt"` or a list of presence flags per power, that adds a deterministic polynomial to the VAR or ARMA equation with coefficients registered as `trend_params`. The trend is built from the symbolic timestep count, so forecasts continue it without scenario data. In `BayesianSARIMAX` it enters the equation of the differenced series, so a constant with `d=1` is a drift in levels, which the observation-intercept route through `exogenous_data` cannot express.

A constant keeps the closed-form `StationaryVAR` likelihood by entering as a level shift. Any time-varying trend falls back to the Kalman filter.

Observation `i` receives the trend evaluated at `i` with `trend_offset=1` in both classes. statsmodels' `VARMAX` uses the same convention. statsmodels' `SARIMAX` is one period behind its `VARMAX`, so the tests pass it `trend_offset=2` to compare the same model.


Dekermanjian (approved):
@jessegrabowski , this looks good to me. One thing I noticed in the notebook was that one of the impulse response figures looks very different than it did before, for the same code as before, is this due to making the disturbance smoother the default?

jessegrabowski:
> impulse response figures looks very different than it did before

That's because of #746. The orthogonalized IRF was simply wrong before. Probably that flag shouldn't be set here at all, it only makes sense for a VAR model.

PR #772, opened 2026-09-11
Bump pymc to 6.3 and pytensor to 3.3

jessegrabowski:
pymc is now pinned to 6.3.x, pytensor to 3.3.x, and preliz allows 0.28. The conda test environment carries the same pins.

`add_data_to_inference_data` now calls `compute_deterministics` with `extend_dataset` instead of the deprecated `merge_dataset`, and the statsmodels comparison test uses the `rng` keyword, with statsmodels floored at 0.15.

The jitter-tuned `test_unstable_lbfgs_update_mask` is replaced by tests that stub L-BFGS to produce `LOW_UPDATE_PCT` and all-paths-failed directly. It relied on softmax underflow breaking the Multinomial logp, which pymc 6.3's `logpow` fix (pymc-devs/pymc#8390) no longer allows.

The conda test env pins `scipy<1.18.1`. conda-forge's 1.18.1 build was compiled with Cython 3.3, whose fused-function export names no longer match the `__pyx_capi__` keys pytensor's numba bridge looks up.


jessegrabowski:
Pathfinder's progress display no longer runs a Rich refresh thread or redirects stdout/stderr; the parent redraws once per polling cycle. Forked workers inherited those locks and hung on Linux, the same bug as pymc-devs/pymc#8421, which is why the Ubuntu shard had excluded these tests. Each path's clock now starts at its first message.

Also: the jitter-tuned `test_unstable_lbfgs_update_mask` is replaced by tests that stub L-BFGS directly, and the conda env pins `scipy<1.18.1` because conda-forge's Cython 3.3 build breaks pytensor's numba symbol lookup.

jessegrabowski:
The previous comment is stale. This PR grew way out of scope and was broken up. That work ended up in #775.

tests/inference/pathfinder/test_multipath.py
```diff
@@ -99,6 +100,36 @@ def _small_serial_fit(model, **overrides):
         return pmx.fit(**kwargs)
 
 
+def test_low_update_pct_paths_are_counted_but_still_succeed(monkeypatch):
```
ricardoV94: unrelated to this PR?

ricardoV94: I see multinomial thing. why not an actual invalid model (neg sigma) iff we actually need to test this?

tests/inference/pathfinder/test_multipath.py
```diff
@@ -99,6 +100,36 @@ def _small_serial_fit(model, **overrides):
...
+
+    assert idata.lbfgs.status_counts.sel(status="LOW_UPDATE_PCT").item() == 3
+    assert idata.pathfinder.path_status_counts.sel(status="SUCCESS").item() == 3
+
+
+def test_all_paths_failing_raises(monkeypatch):
+    """When L-BFGS fails on every path there is nothing to sample from, and fit raises rather than
+    returning an empty result."""
+
+    def fail(self, callback, x0):
```
ricardoV94: I'm not a fan on monkeypatch tests like this.

jessegrabowski: Why, the test is of failure behavior. Trying to engineer a failure is fragile.

ricardoV94: negative sigma? Otherwise I wouldn't even add a test if we can't trigger it. Monkeypatch gives a false sense of coverage imo

jessegrabowski: ok

tests/inference/pathfinder/test_multipath.py
jessegrabowski: done

tests/inference/pathfinder/test_multipath.py
```diff
@@ -99,6 +101,27 @@ def _small_serial_fit(model, **overrides):
...
+
+    assert idata.lbfgs.status_counts.sel(status="LOW_UPDATE_PCT").item() > 0
+    assert idata.pathfinder.path_status_counts.sel(status="SUCCESS").item() > 0
+
+
+def test_all_paths_failing_raises():
+    """A negative prior sigma gives every path a non-finite logp, so L-BFGS cannot take a step
+    and fit raises rather than returning an empty result."""
+    with pm.Model() as model:
+        pm.Normal("x", mu=0.0, sigma=-1.0)
```
Dekermanjian, line 119: This is interesting. I didn't know `pm.Normal` even let you set a negative sigma.

ricardoV94 (approved):

pyproject.toml
```diff
@@ -34,12 +34,12 @@ keywords = [
 license = {file = "LICENSE"}
 dynamic = ["version"]  # specify the version in the __init__.py file
 dependencies = [
-  "pymc>=6.2.0,<6.3",
-  "pytensor>=3.2.3,<3.3",
+  "pymc>=6.3.0,<6.4",
+  "pytensor>=3.3.0,<3.4",
   "arviz>=1.2,<2.0",
   "better-optimize>=0.4.2,<1.0",
```
Dekermanjian, line 40: Does this conflict with the scipy pin set in `environment-test.yml`? The decencies listed in `better-optimize`'s pyproject.toml pin scipy to `'scipy>=1.15'`

Dekermanjian (approved):
@jessegrabowski, I think this looks good. I am just not sure if `better-optimize` will bump `scipy` above the pin set in the environment file.

pyproject.toml
jessegrabowski: good question, let me check

tests/inference/pathfinder/test_multipath.py
jessegrabowski: there's a `check_params` helper in pymc that inserts `Assert` Ops. It will let you write down the graph but it will immediately fail at constant folding time I think

pyproject.toml
jessegrabowski: Yes this is fine, optimize wants >=1.15, and this repo wants <0.18.1. So anything in the range `[1.15, 1.18.1)` will solve

PR #775, opened 2026-09-12
Fix pathfinder's forked workers hanging on MKL and crashing on Accelerate

jessegrabowski:
`multipath_pathfinder` now caps BLAS to one thread in the parent while forked workers run. MKL keeps an OpenMP team alive after any large product, and a forked worker inherits its locks with none of its threads, so its first large product blocks forever. The cap is set in the parent because touching the OpenMP runtime inside a forked process is what crashed pymc's workers in pymc-devs/pymc#7354.

On macOS the default `mp_ctx` is now `"spawn"`. Accelerate runs large products on libdispatch, which aborts in any process forked from a parent that already used it, forkserver included. Under fork every path died at the final draw and `fit` reported "All paths failed".

The progress display no longer runs Rich's refresh thread or redirects stdout and stderr. The parent redraws once per polling cycle, and each path's clock starts at its first message. Same hazard as pymc-devs/pymc#8421. The Ubuntu shard runs the pathfinder tests again, excluded since #597 for this hang.

Both failures reproduce in a bare `multiprocessing` child with no pathfinder involved.


tests/inference/pathfinder/test_multipath.py
```diff
@@ -99,6 +105,42 @@ def _small_serial_fit(model, **overrides):
         return pmx.fit(**kwargs)
 
 
+@pytest.mark.parametrize(
+    "system, expected", [("Darwin", "spawn"), ("Linux", None)], ids=["macos", "linux"]
+)
+def test_default_context_spawns_on_macos(monkeypatch, system, expected):
+    """macOS defaults to spawn; elsewhere the choice is left to pymc's resolver."""
```
ricardoV94, line 112: Why would this be different than pymc? One point Darwin is not a homogeneous thing, I believe people have flip flopped on this across versions over the years so you may need to check versions as well to have a consistent method.

tests/inference/pathfinder/test_multipath.py
```diff
@@ -99,6 +105,42 @@ def _small_serial_fit(model, **overrides):
...
+
+    _resolve_mp_context(None, mode=None)
+
+    assert seen == [expected]
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="fork start method is unavailable on Windows")
+@pytest.mark.parametrize("parallel", [True, False], ids=["parallel", "serial"])
+def test_fork_caps_blas_to_one_thread_only_while_workers_run(monkeypatch, parallel):
+    """A parallel fit under fork caps BLAS to one thread in the parent so forked workers inherit
```
ricardoV94, line 129: what does this mean? If I have 8 cores but do 4 fits only I don't get 2 blas cores per fit?

tests/inference/pathfinder/test_multipath.py
```diff
@@ -193,14 +235,41 @@ def patched(*args, **kwargs):
         np.testing.assert_array_equal(forward[var].values, backward[var].values)
 
 
+def test_live_progress_display_starts_no_thread_and_no_stream_proxies():
```
ricardoV94, line 238: Not a test I'd like to have in the CI, because it's too specific on today's implementation and because I don't trust the assessment of the problem

tests/inference/pathfinder/test_multipath.py
```diff
@@ -193,14 +235,41 @@ def patched(*args, **kwargs):
...
+        elbo="",
+        speed=0.0,
+        speed_unit="it/s",
+        total=1000,
+        completed=0,
     )
     return progress, task_id
 
 
+def test_progress_callback_starts_the_clock_on_first_message():
```
ricardoV94, line 262: The logical conclusion of having this kind of test for every sampling method we implement is absurd. We don't need a test for every line the bot changed

ricardoV94 (commented):
Smells a lot like local optimization code wise and possibly over-specialization to your current OS and version.

I may be wrong but do at least one step of pushing back against Claudio to see what happens

tests/inference/pathfinder/test_multipath.py
jessegrabowski: its not, i got pymc to fail in the same way. We don't see it because of the nutpie default.

jessegrabowski:
> Smells a lot like local optimization code wise and possibly over-specialization to your current OS and version.

Well the windows CI failure is what got me working on this. We also had the tests skipped on ubuntu, so this is legit broken.

The core issue is the ordering of starting new processes, rich bars, and pytensor compiles. pymc seems to be less fragile because we set up everything once up front. The way the pathfinder code is currently written, we start up new processes as the old ones finish.

tests/inference/pathfinder/test_multipath.py
ricardoV94: So shouldnt we fix the pymc resolver and this will pick it automatically?

PR #776, opened 2026-09-12
Share compiled graphs across statespace tests and trim the Windows matrix

### LINKED ISSUE #773: Statespace test suite is way too slow
It takes over an hour on the github action runners. Like bruh.

jessegrabowski:
The statespace tests that compiled the same Kalman filter graph once per case now compile it once per test class, and the Windows CI matrix runs one end-to-end workflow per statespace model instead of the whole statespace suite.

Closes #773

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,8 @@ jobs:
           - tests/model
           - tests/statespace/core/test_statespace.py
           - tests/statespace/filters/test_kalman_filter.py
-          - tests/statespace --ignore tests/statespace/core/test_statespace.py --ignore tests/statespace/filters/test_kalman_filter.py
+          - tests/statespace/models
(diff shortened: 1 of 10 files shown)


ricardoV94 (commented):
Have you considered using FAST_COMPILE or CVM backend mode instead of changing the STS tests (unless you prefer this form regardless)

jessegrabowski:
This PR takes us from 80 minute CI to 15 minutes so I want it as-is, happy if anyone wants to do more optimization.

ricardoV94:
It's just optimizing tests in this way makes them more coupled. And I'm also sure AI will immediately ignore this approach next iteration and start drifting towards the 80m again. I also suspect overzealous testing going on.

But I'm not blocking it at all, specially if you're happy with this setup. Just wanted to confirm you are

ricardoV94 (approved):

PR #777, opened 2026-09-14
Fit BayesianETS through its innovations recursion instead of the Kalman filter

jessegrabowski:
`BayesianETS` now evaluates its likelihood through the ETS error recursion when `stationary_initialization=True`, there is no measurement error, and the data has no gaps. Anything else is still filtered. ETS has a single source of error, so its states are exactly determined by the data and the one-step-ahead errors come from a recursion with no covariance in it. It returns the same density as the Kalman filter. `InnovationsStateSpace` in `filters/distributions.py` is the distribution, and `BayesianETS.make_likelihood` registers it, following the same pattern as #766.

Two breaking changes. `stationary_initialization` now defaults to True and fixes `P0` at `R Q R'`, so a model built with the defaults has no `P0` parameter and an existing `P0` prior becomes an unused variable. The `initialization_dampening` argument is gone. The dampened Lyapunov solve approximated a steady state the single-source-of-error filter reaches exactly at `R Q R'`, and that is also the textbook ETS initialization. `R Q R'` is singular, so `sample_conditional_posterior` and friends need `mvn_method="svd"` or `"eigh"` rather than `"cholesky"`. The example notebook accepts the default everywhere except the damped-trend model, which shows the `P0` prior.

Evaluation time, median of a pytest-benchmark run, both graphs over the same matrices in float64. The gradient is where the win is, and it grows with the seasonal period because the filter's per-step cost is quadratic in the state size.

```
model              measure   filter      recursion   speedup
ANN, T=200         logp        251.0us     49.1us      5.1x
ANN, T=200         grad        907.1us     46.5us     19.5x
ANN, T=1000        logp       1227.5us    207.8us      5.9x
ANN, T=1000        grad       4253.2us    203.4us     20.9x
ANN, T=5000        logp       6120.5us    974.2us      6.3x
ANN, T=5000        grad      20702.7us    949.1us     21.8x
AAN, T=1000        logp       1358.6us    204.3us      6.7x
AAN, T=1000        grad       4427.0us    202.0us     21.9x
AAA, m=12, T=1000  logp       3980.6us    242.4us     16.4x
AAA, m=12, T=1000  grad      13385.3us    245.8us     54.4x
```


Dekermanjian (commented):
@jessegrabowski I looked over the changes and they look good to me. One of the tests is failing due to a negative value being returned by `np.linalg.eigh`. 

```
w = array([-5.15983893e-09,  3.62526210e-09,  3.14261356e+00])
```

jessegrabowski:
Good catch. I switched the test to use `mvn_method='svd'`, which is more robust to these small negative eigenvalues.

PR #778, opened 2026-09-14
Refuse to forecast from a DatetimeIndex with no frequency

jessegrabowski:
`_build_forecast_index` now raises one `ValueError` when the fitted DatetimeIndex has no frequency. Before, `pd.date_range` defaulted to daily and the `periods`, `end`, and array-scenario paths failed at the frequency-mismatch guard with "Got <Day>, expected None", and a scenario DataFrame with `use_scenario_index=True` hit a `TypeError` in the step-back. A scenario index with no frequency gets its own message.

`NO_FREQ_INFO_WARNING` now reads "specified" instead of "specific" and says a frequency will be inferred and that forecasting needs one. The eight `filterwarnings` marks that matched the old text are updated.


pymc_extras/statespace/core/forecast.py
```diff
@@ -37,6 +37,15 @@
 
 _log = logging.getLogger("pymc.experimental.statespace")
 
+NO_FIT_FREQ_ERROR = (
+    "The DatetimeIndex the model was fit on has no frequency, so it cannot be extended into a "
```
Dekermanjian, line 41: @jessegrabowski, does this mean someone could fit a model and then after fitting find out they need to fit again because they didn't set a `DatetimeIndex`?

Dekermanjian (commented):
@jessegrabowski, the changes look good. I am a little concerned, does this mean that users can no longer work exclusively in `numpy` or `pytensor` and that they must use `pandas`?

jessegrabowski:
> I am a little concerned, does this mean that users can no longer work exclusively in numpy or pytensor and that they must use pandas?

No, this PR is behavior preserving. If you give a DatetimeIndex, we force you to be consistent at forecast time. If you don't, then it's all good. 

This PR is actually a first step to the narwhals cutover. I want to get away from pandas-only in general.

pymc_extras/statespace/core/forecast.py
jessegrabowski: Yes, in one obnoxious corner case. If the data shown at fit time was a DatetimeIndex with 1) no set frequency and 2) whose frequency cannot be inferred by pandas, every forecast will raise.

An example would be something like business days with a custom holiday calendar. We currently return the wrong forecast index in the case. I don't know a better solution than just raising. These types of custom calendars are very hard to support correctly.

jessegrabowski: To be clear though, numpy and integer indexes *never* hit any of these checks, so a user always has that escape hatch.

Dekermanjian (approved):

pymc_extras/statespace/core/forecast.py
Dekermanjian: Okay, that makes a lot of sense. Thank you for the clarification, Jesse!

PR #780, opened 2026-09-16
Keep missing observations masked in the simulation smoother

### LINKED ISSUE #779: SimulationSmoother treats missing observations as data: `data - y_plus` breaks the fill-value mask
`sample_conditional_prior` and `sample_conditional_posterior` return smoothed states on the order of the missing fill value whenever the data has missing entries and `joint_smoothed_draws` is on. `SimulationSmoother.rv_op` smooths `delta = data - y_plus` in one pass, and at a missing position that is `-9999 - y_plus`, which is neither NaN nor equal to `missing_fill_value`, so `mask_missing_values` no longer masks it and the smoother fits those periods to a value near -9999. The filter and the non-joint smoother path mask correctly, so the likelihood and `make_filter_outputs` are unaffected.

```python
import numpy as np
import pymc as pm
from pymc_extras.statespace.models import structural as st

y = np.cumsum(rng.normal(size=100))
y[40:70] = np.nan  # a missing block

ss_mod = (st.LevelTrend(order=1, innovations_order=1) + st.MeasurementError()).build(verbose=False)

with pm.Model(coords=ss_mod.coords) as model:
    pm.Normal("initial_level_trend", dims=["state_level_trend"])
    pm.HalfNormal("P0", 1, dims=["state", "state_aux"])
    pm.Exponential("sigma_level_trend", 1, dims=["shock_level_trend"])
    pm.Exponential("sigma_MeasurementError", 1)
    ss_mod.build_statespace_graph(y[:, None])
    prior = pm.sample_prior_predictive(draws=10, random_seed=1)

smoothed = ss_mod.sample_conditional_prior(prior, random_seed=1).smoothed_prior
print(float(np.abs(smoothed).max()))  # 10001.3

ss_mod.joint_smoothed_draws = False  # workaround
smoothed = ss_mod.sample_conditional_prior(prior, random_seed=1).smoothed_prior
print(float(np.abs(smoothed).max()))  # 14.0
```

Potential fix (requires testing):

```python
missing = pt.or_(pt.isnan(data_), pt.eq(data_, kalman_filter.missing_fill_value))
delta = pt.where(missing, kalman_filter.missing_fill_value, data_ - y_plus)
```

jessegrabowski:
`SimulationSmoother` now writes the missing fill sentinel back into `data - y_plus` at missing positions, so the one-pass smoother keeps those periods masked. The sentinel minus `y_plus` is neither NaN nor the sentinel, so the filter fit those periods to a value near -9999 whenever `joint_smoothed_draws` was on.

The NaN-or-sentinel predicate moved into a `missing_mask` helper in `filters/utilities.py`, which the filter, the convergent filter's data check, and the simulation smoother now share.

The filter log-likelihood test now scores the sentinel encoding as well as NaN. The statespace core hands the filter the sentinel, and no filter test exercised it before.

Closes #779


Dekermanjian (approved):
Looks good!

